PR #95, opened 2019-05-21
Add curl --fail to notify when redeployment fails

eddieantonio:
The `--fail` flag will give an error exit code when the server returns a 4xx, or 5xx error.

This will let us know if something is amiss during deployment. At least, theoretically!

e.g.,

```ssh
$ curl --fail sapir.artsrn.ualberta.ca/redeploy/cree-dictionary
curl: (22) The requested URL returned error: 405 Method Not Allowed
$ echo $?
22
$ curl sapir.artsrn.ualberta.ca/redeploy/cree-dictionary

Status: 405 Method Not Allowed
Invalid invocation. You must make a POST request with the secret.

    curl -XPOST -dsecret=XXXXXX sapir.artsrn.ualberta.ca/redeploy/cree-dictionary
$ echo $?
0
```


Madoshakalaka:
nice

PR #173, opened 2019-11-22
new fst releases available

Madoshakalaka:
this pull request is automated and triggered by new fst releases


eddieantonio:
This is not ready yet, but it's close.

PR #183, opened 2019-12-09
Show at least two lemmas for lexicalized dimunitive forms 

Madoshakalaka:
the importing process is fixed

### DIFF
--- a/CreeDictionary/DatabaseManager/xml_entry_lemma_finder.py
+++ b/CreeDictionary/DatabaseManager/xml_entry_lemma_finder.py
@@ -1,6 +1,7 @@
 """
 EXPAND lemma with inflections from xml according to an fst and paradigm/layout files
 """
+from collections import defaultdict
 from itertools import chain
 from typing import List, Dict, Tuple, Set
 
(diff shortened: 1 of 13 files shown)


CreeDictionary/DatabaseManager/xml_entry_lemma_finder.py
```diff
@@ -110,21 +111,40 @@ def extract_fst_lemmas(
 
                 ambiguities: List[Analysis] = []
 
+                # put priority in looking for an unique and identical wordform
```
eddieantonio:
```suggestion
                # put priority in looking for a unique and identical wordform
```

package.json
```diff
@@ -4,10 +4,10 @@
   "description": "Cree Intelligent Dictionary",
   "scripts": {
     "start": "rollup --config --watch & pipenv run dev",
-    "bg-watch": "rollup --watch &",
+    "bg-watch": "rollup -c && rollup --watch &",
```
eddieantonio: Interesting! Why this change?

cypress/integration/regressions.spec.js
```diff
@@ -97,10 +97,46 @@ context('Regressions', () => {
...
+    })
+  })
+  
+  // https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/176
+  describe('should show at least two lemmas for lexicalized diminutive forms', ()=>{
+    it('should show atâhk and acâhkos for acâhkos', () => {
+      cy.visit('/search/acâhkos')
+      cy.get('[data-cy=search-results]')
+        .should('contain', 'atâhk')
+        .and('contain', 'acâhkos')
```
eddieantonio: Nice!

Madoshakalaka:
@eddieantonio Need you to checkout this branch and do a `manage-db import /res/dictionaries/ --multi-process 2` and `push-db` before this is merged because I don't have access to Sapir.

Merging right now won't have any impact in the production environment though, just meaning we need to manually update database later

package.json
Madoshakalaka: I think `rollup --watch` does not build anything when you start the command. I may be wrong but adding rollup -c has no impact

cypress/integration/regressions.spec.js
Madoshakalaka: yeet!

PR #186, opened 2019-12-11
Add responsive CSS to reposition search based on desktop/mobile

eddieantonio:
This adds a few responsive changes to the CSS. The most notable one is that the search bar collapses as part of the top bar when the width is sufficiently wide.

The second change is that the size of the content never goes beyond a certain width. That means that on super wide screens, the content remains in a tight, readable column.


Madoshakalaka:
very very nice good job!

Issue #113, opened 2019-10-02
Refine ranking for fuzzy matching


eddieantonio:
In the goal of _progressive enhancement_, let's disable the Levenshtein distance fuzzy search. Fuzzy matching the way itwêwina does is fine for now.

@aarppe, could you provide your example of when this is a problem?

Later on, we'll need to adjust the edit distance calculation, but that's a future TODO!

Madoshakalaka:
It's easy to turn off fuzzy search. What is the bug though?

eddieantonio:
@aarppe says that some of the search results are DANGEROUSLY misleading.
Plus there is the possibility of having too many results. We don't yet have
a nice way of ranking the too many results, so it's better to just turn off
fuzzy search for now. The bug is that fuzzy search is on!

Madoshakalaka:
Whoa. Sure! I'll turn it off.

aarppe:
The main problem from the perspective of a naive user, the fuzzy search produced a lot of results, some of which were semantically funny, some repetitions (due to MD content, maybe), and then the lemma one was after might be quite a ways further down.

That does not induce confidence in newbie users (like Arok).

I think that this can be resolved by careful tuning of the fuzzy distance calculator (including edit distance with different costs for different edits, lemma frequency, and other stuff), but we need to come up with the weights (I'd suggest using an adaptation of those from the spell-checker) and start making use of the lemma-frequencies (we've got them), implement those in a trial version, and then test how that works.

But now, e.g. searching with ninipan, I get both nipâw 'sleep' and nipiw 'die'.

Of course, ed(â,a) and ed(â,i) is =1 is the same, but here, and in some other trials, the results, if not ordered according to some though-through metrix, are not good.

Eddie has written some tentative guidelines about what the weighting could be, pretty much similar to the spell-checker weighting.

ls ~/giella/langs/crk/tools/spellcheckers/fstbased/desktop/hfst/*.default.txt
/Users/arppe/giella/langs/crk/tools/spellcheckers/fstbased/desktop/hfst/editdist.default.txt
/Users/arppe/giella/langs/crk/tools/spellcheckers/fstbased/desktop/hfst/final_strings.default.txt
/Users/arppe/giella/langs/crk/tools/spellcheckers/fstbased/desktop/hfst/initial_letters.default.txt
/Users/arppe/giella/langs/crk/tools/spellcheckers/fstbased/desktop/hfst/strings.default.txt
[...]

Issue #96, opened 2019-06-10
Sort search results based on modified Levenshtein edit distance


eddieantonio:
The search results should be ranked in order of [edit distance](https://en.wikipedia.org/wiki/Levenshtein_distance) from the query.

These distances should be calculated *after* the query is converted to SRO/circumflex.

 * An **exact match** will have a distance of 0, and should be ranked first.
 * An **addition**  should have a distance of 1: e.g., minôs vs. minôhs
 * A **deletion** should have a distance of 1: e.g., acâhkos vs. acâhko.
 * A **substitution** should have a distance of 1: e.g., ekwa vs. ikwa

Additionally, there should be "half" substitutions:
 * vowel lengths should incur a distance of ½
 * an 'h' in a rime should incur a distance of ½

Vowel length on "ê" vs "e" should count as a distance of 0.

aarppe:
Some comments:

1. Would an exact match concern not just a lemma, but an inflected (i.e. such a form is written exactly according to the orthographical norm)? Even then one might want to order the results based on reverse corpus frequency for the lemma, unless we'd use a fully weighted analyzer FST.

2. Concerning the weights (for various orthographical edits), the error-correction model in the Giella infrastructure already has various specifications, cf. the following files:

ls tools/spellcheckers/fstbased/desktop/hfst/*.default.txt

3. Indeed, one could use the spell-checker version of the (H)FST, with 'hfst-ospell -S' to weight search results, when they don't exactly match with a search string. Interestingly, our collaborators at Carleton use this in their East Cree dictionary lookup. For instance:

hfst-ospell -S tools/spellcheckers/fstbased/desktop/hfst/crk.zhfst
acimew
"acimew" is NOT in the lexicon:
Corrections for "acimew":
âcimêw    5.079442
atimêw    13.079442
ahimêw    13.079442
akimêw    13.079442
âcimâw    15.079442
âcimow    15.079442
akimâw    23.079441
acimos    23.079441
aciyaw    23.079441
atimâw    23.079441

That allows even for the inclusion of word-form frequencies as part of the weighting of misspelling corrections.

Unfortunately, currently the Giella webpages describing error-correction weighting in full are only partially accessible due to a long-waited reorganization of their website. Nevertheless, one can take inspiration from the explicitly specified weights (and corrections).

aarppe:
Within the Giella infrastructure, the following HFST command and specification appears to do the job in the case on single-character edit-distance:

[...]

PR #195, opened 2019-12-15
Rank search results by levenshtein distance

Madoshakalaka:


Madoshakalaka:
@eddieantonio what do you suggest we do to #113

CreeDictionary/utils/cree_lev_dist.py
```diff
@@ -1,21 +1,56 @@
...
+
+
+def remove_cree_diacritics(input_str) -> str:
+    """    
+    >>> remove_cree_diacritics('ā')
+    'a'
+    >>> remove_cree_diacritics('â')
+    'a'
     """
+    nfkd_form = unicodedata.normalize("NFKD", input_str)
```
eddieantonio: Could you justify why you're using NFKD over NFD? For the expected inputs, they would both have the same effect.

eddieantonio:
We can't really close #113 until we incorporate wordform weights for each entry, but at least ranking the results using edit distance gets us better weights

CreeDictionary/utils/cree_lev_dist.py
Madoshakalaka: I googled "strip diacritics python" and this function is copy pasted. You can change it if neccesary.

CreeDictionary/utils/cree_lev_dist.py
```diff
@@ -1,21 +1,56 @@
...
+    only_ascii = nfkd_form.encode("ASCII", "ignore").decode("utf-8")
```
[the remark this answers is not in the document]
Madoshakalaka: ikr

CreeDictionary/utils/cree_lev_dist.py
eddieantonio: This is not a fool-proof code snippet, but I guess it works for now ¯\_(ツ)_/¯

eddieantonio:
Does this close #96?

Madoshakalaka:
Don't close it right now. I'm updating test examples from made-up ones to "acahkos", and just found the function returns not expected distance. I'm in the middle of debugging it.

aarppe:
@Madoshakalaka @eddieantonio From the linguist's perspective on what they might want to tweak, regardless of the computational implemention, where are the weights explicitly specified? Cf. the proposed modified weights in the comment: https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/96#issuecomment-549648171

aarppe:
@Madoshakalaka for a comprehensive overview with more granularity as to how the edits should be weighted, see https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/96#issuecomment-549648171

PR #198, opened 2019-12-20
Use accessible HTML & CSS for the search bar

eddieantonio:
This follows the best practice to make the search bar usable on screen readers.


eddieantonio:
@Madoshakalaka or @kobexamoh, you can review this; not sure why it was set as "draft"

Issue #188, opened 2019-12-12
Server error (HTTP 500) when accessing about page


eddieantonio:
But the integration tests that access the about page pass? 🤔

https://sapir.artsrn.ualberta.ca/cree-dictionary/about

eddieantonio:
@Madoshakalaka could you please have a look at this? This is what I've tried:

 - checking `/var/log/apache2/error.log` and `/var/log/apache2/access.log` -- nothing from this app shows up there
 - pinging the backend server, however it gives me `400 Bad Request` whenever I try to curl it, e.g., `curl http://localhost:8091/cree-dictionary/`

eddieantonio:
This is caused by the Django static middleware suddenly caring about 404s when `DEBUG=False`. There was a typo when requesting `nrc-cnrc-logo.svg` — it was actually requesting `nrc-cnrc-logo.png`, which does not exist!

PR #202, opened 2020-01-08
mega log

Madoshakalaka:
Logging is configured in django style, in order to investigate #188 and future issues alike.

Now in development, things don't change.

On production though, django will log everything under `CreeDictionary/django_logs`, including everything django shows when `DEBUG` is set to true in development.


Madoshakalaka:
@eddieantonio I merged it. It's a small logging change that won't break anything (probablly), but I need to use it to really investigate the 500 issue.

Issue #201, opened 2019-12-31
Show definition for inflected forms


Madoshakalaka:
As discussed [here](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/181#issue-533839591), it's desirable to show definition for inflected forms (not only lemmas) when they exist. 

Currently we only show users the matched lemmas. This issue depends on the following UI where we'd also show matched inflections.

<img width="477" alt="Capture" src="https://user-images.githubusercontent.com/44753941/71635777-afc9d780-2be4-11ea-912a-24a122bac58b.PNG">

aarppe:
A few important notes:

1. the string _ê-nêhiyawi-kiskinwahamâkosiyân_ could have two analyses, in which one has the preverb _nêhiyawi-_ lexicalized as the lemma _nêhiyawi-kiskinwahamâkosiw_, and the other with the preverb as a productive mofifier of the lemma _kiskinwahamâkosiw_.

Thus, the **two** correct representations for the search string _ê-nêhiyawi-kiskinwahamâkosiyân_ are the following:

![image](https://user-images.githubusercontent.com/15003407/71940173-94c20f00-3172-11ea-8cf0-426960af870b.png)

![image](https://user-images.githubusercontent.com/15003407/71940335-2df12580-3173-11ea-8405-42d8e3b8fa62.png)

Madoshakalaka:
@aarppe I'll try to produce them

Madoshakalaka:
I just noted we do not have "I am learning Cree" in our database but we do have definition for the lemma.
![Screenshot from 2020-01-14 15-27-22](https://user-images.githubusercontent.com/44753941/72388244-b0816400-36e2-11ea-81f4-1ea50442620b.png)


Since not all inflections have definition, and we'll possibly resort to English fst in the future, in the case of matched inflection without definition, should we add a piece of reminder in our user interface that says 

**showing definition for lemma below**

@aarppe @eddieantonio What do you think?

Madoshakalaka:
@aarppe @eddieantonio I just re-examed #153 , so we'll have infomation about the lemma embedded in a popup with "(?)" icon right next to "form of xxxxx".

If I get the idea correctly, we will not show any definition in the case metioned above, and the purpose of "Form of xxxxx (?)" will be to redirect the user to our lemma page. Is this more preferred?

eddieantonio:
[...]

PR #206, opened 2020-01-14
show normatized form and link that refers to lemma page

Madoshakalaka:
## Visual Change
As described in title

Screenshot from 2020-01-14 14-16-39

## Reminder

Despite we show possibly inflected normatized form now, the definition we show still comes from the lemma. We need to implement #201 and display definition for inflections  

## Technical Change/Reminder

`cree_results` in `CreeAndEnglish` is now of type 
`SortedSet[Tuple[Analysis, NormatizedCree, Lemma]]` 

where
`NormatizedCree = NewType(..., 'str')`

Sorting of the search results is enabled by providing a Callable as `key` var to `SortedSet` constructor. 

to fullfill #201 we'd want
`NormatizedCree = NewType(..., 'Wordform')` in the future
so that it carries information (specifically: definition) for the inflection


CreeDictionary/API/models.py
```diff
@@ -1,19 +1,19 @@
...
-from constants import SimpleLC, SimpleLexicalCategory, POS
+from constants import SimpleLC, SimpleLexicalCategory, POS, Analysis
 from fuzzy_search import CreeFuzzySearcher
-from shared import descriptive_analyzer_foma, normative_generator
-from utils import fst_analysis_parser
+from shared import descriptive_analyzer_foma, normative_generator_foma
+from utils import fst_analysis_parser, get_modified_distance
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("django")
```
eddieantonio: `logging.getLogger(__name__)` is a common Python idiom with the logger library, and the [recommended way to do it](https://docs.python.org/3.8/library/logging.html#logger-objects) That way, log messages from just this module will be prefixed with "CreeDictionary.api.models.py".

I would change this back to the way it was originally.

```suggestion
logger = logging.getLogger(__name__)
```

CreeDictionary/API/models.py
```diff
@@ -354,7 +374,7 @@ def search(user_query: str) -> Tuple[str, List[SearchResult]]:
         matched_wordforms: Set[EntryKey] = set()
 
         # First, let's add all matched analyses:
-        for analysis in cree_results.keys():
+        for analysis in [result[0] for result in cree_results]:
```
eddieantonio: I'm not sure what `result[0]` is. I think this is the `Tuple[Analysis, NormatizedCree, Lemma]`. If that's the case, could you make the latter a named tuple with properties `analysis: Analysis; cree: NormatizedCree; lemma: Lemma`?

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
```diff
@@ -9,31 +9,77 @@ <h2 class="definition-title definition-title--search-result">
...
 
       {# Show the matched lemma (when this is NOT a lemma. #}
-      {% if not word.is_lemma %}
-      <p class="definition-title__link-to-lemma">
+      {% if analysis != lemma_wordform.analysis %}
+      <p class="definition-title__reference-to-lemma" data-cy="reference-to-lemma">
         Form of
         {# TODO: this link should go to the lemma search result, further down the page #}
-        <a href="#">{{ word.lemma.text }}</a>
+        <a class="definition-title__link" href="javascript:void(0)" data-lemma-id="{{ lemma_wordform.id }}">{{ lemma_wordform.lemma.text }}</a>
```
eddieantonio: Could the `href` here link to the `id` of the lemma search result, instead of doing `javascript:void(0)`? That way, the browser will automatically scroll to the search result of the lemma (which has the actual definition).

cypress/integration/search.spec.js
```diff
@@ -82,22 +82,26 @@ context('Searching', () => {
     })
   })
 
-  // todo: this
   describe('I want to see the normatize form of my search', () => {
-    it.skip('should search the normatized form of the matched search string', () => {
+    it('should search the normatized form of the matched search string', () => {
```
eddieantonio: niiiiiiice!

CreeDictionary/API/models.py
Madoshakalaka: I know it's an idiom. I Just wanted to use the existing "django" handler and the logging handlers configured in `settings.py` so that the logger outputs both in console and the file in production. But I think that's achivable with proper configuration regarding logging propagation. I'll investigate later (or never, until I need investigate with logging again)

CreeDictionary/API/models.py
Madoshakalaka: Sure I'll name it CreeMatch what do you think

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
Madoshakalaka: TIL lol, will change

CreeDictionary/API/models.py
Madoshakalaka: or CreeResult, since the variable name is "cree_results"

eddieantonio: `CreeResult` 👍

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
eddieantonio: So the lemma will be like `<a id="defined:wâpamêw">wâpamêw</a>` and the matched wordform will be a link: `ê-wâpamât form of <a href="#defined:wâpamêw">wâpamêw</a>`. That way, redirection is automatically handled by the browser! No JavaScript required!

CreeDictionary/tests/conftest.py
```diff
@@ -1,17 +1,24 @@
...
 from API.models import Wordform
-from constants import SimpleLexicalCategory
-from DatabaseManager.xml_importer import import_xmls
-from utils import fst_analysis_parser, shared_res_dir
+from utils import shared_res_dir
+
+
+from hypothesis import settings
+
+settings.register_profile("default", deadline=timedelta(milliseconds=500))
```
eddieantonio: 👍

src/css/variables.css
```diff
@@ -47,6 +47,7 @@
   --branding-font-size:             1.50em; /* the title, e.g., "itwêwina" */
   --branding-subtitle-font-size:    0.75em; /* the subtitle, e.g., "Plains Cree Dictionary" */
   --definition-title-size:          2.00em;
+  --reference-to-lemma-size:             1.5em;
```
eddieantonio: Nitpick: this single line is spaced differently than the rest. Could you align it with all the others? I won't block merge if you don't lol

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
Madoshakalaka: OOooOOoh I think I misunderstood what you expected search to do here. So you expect 

> if user searches 'nipa' " we should show
> 1. nipâ - form of **nipâw**
> 2. **nipâw**
>
> and if user clicks on the first **nipâw**, it jumps to the second, but on the same page

Is that your expectation when you designed everything?

In the current search code, the second lemma is not displayed, and the first **nipâw** link actually jumps to the lemma detail page of nipâw.

Madoshakalaka: can we merge this for now, to achieve what's said requires modification to the search code. I think it will be easy to add on, based on the current detailedly typed search code.

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
```diff
@@ -1,46 +1,2 @@
...
-          <cite class="cite-dict">{{ source }}</cite>
-        {% endfor %}
-      </li>
-    {% endfor %}
-    </ol>
-  </article>
-</li>
-{% endfor %}
+{% include 'CreeDictionary/cree-entries.html' %}
+{% include 'CreeDictionary/english-entries.html' %}
```
eddieantonio: So if I understand this correctly, first it shows Cree results, then it shows English results? That is kinda weird, but I won't block merge on it.

eddieantonio:
Please merge master into this, then use _your_ test_db.sqlite3 and commit it here.

Issue #100, opened 2019-06-20
Presentation of orthographically standardized form of search string


aarppe:
This means presenting "somewhere" (potentially apparent by default but as a pop-up or button) the strictly normative form corresponding to the spell-relaxed linguistic analysis of the search string, e.g.

- [x] nikiwapamaw -> nikî-wâpamâw
- [x] nipa -> nipâ (the preverb _nipâ-_ is mispelled in MD, lacking vowel length and the hyphen)
- [x] nipaw -> nipâw
- [x] ninipan -> ninipân
- [x] nikimamiyonipan -> 'nikî-ma-miyo-nipân
- [x] enipat -> ê-nipât

The standardization should follow the selection of content representation, i.e. either flavor of SRO or Cree syllabics.

Note that there may be two or more possible linguistic analyses for the same search string, e.g.

- [x] e-nipayan
-> ê-nipâyân <- `PV/e+nipâw+V+AI+Cnj+Prs+1Sg`
-> ê-nipâyan <- `PV/e+nipâw+V+AI+Cnj+Prs+2Sg`

Or:

(`echo nipihk | hfst-lookup -q ~/giella/langs/crk/src/analyser-gt-desc.hfst | cut -f2`)

```
nipihk
-> nipiy+N+I+Loc
-> nîpiy+N+I+Loc
-> nipiw+V+AI+Cnj+Prs+X
-> nîpin+V+II+Cnj+Prs+3Sg
-> nipiw+V+AI+Imp+Imm+2Pl
```
(`echo nipihk | hfst-lookup -q ~/giella/langs/crk/src/analyser-gt-desc.hfst | cut -f2 | hfst-lookup -q ~/giella/langs/crk/src/generator-gt-norm.hfst`)

nipihk ->
- [x] nipîhk (`nipiy+N+I+Loc: 0.000000`)
- [x] nîpîhk (`nîpiy+N+I+Loc: 0.000000`)

- [x] nipihk (`nipiw+V+AI+Cnj+Prs+X: 0.000000`)
- [x] nipiniwahk (`nipiw+V+AI+Cnj+Prs+X: 0.000000`)
- [x] nipinâniwahk (`nipiw+V+AI+Cnj+Prs+X: 0.000000`)

- [x] nîpihk (`nîpin+V+II+Cnj+Prs+3Sg: 0.000000`)

- [x] nipik (`nipiw+V+AI+Imp+Imm+2Pl: 0.000000`)

[...]

PR #214, opened 2020-01-16
Err/Orth stripping in normalization

Madoshakalaka:


Madoshakalaka:
see #100. I'll merge it since it's too small a change

PR #216, opened 2020-01-16
Changed occurence of "Normatized" to "standardized"

Madoshakalaka:
What **are** English words:

- Normal
- Normalize
- Normative

What are **not**:
- Normatize 
- Normatized

Screenshot from 2020-01-16 16-26-51

We do have a "normative_generator", I guess in that sense, we can say "normatize" means to "make some word normative". But I hate the fact that it's not in the dictionary. And even #100 uses the word "standardized" instead of "normatized" in the title. So here's me spending 10 minutes change all occurence of "normatize/normatized" to "standardize/standardized"

### DIFF
--- a/CreeDictionary/API/models.py
+++ b/CreeDictionary/API/models.py
@@ -78,7 +78,7 @@ class EntryKey(NamedTuple):
     lc: SimpleLexicalCategory
 
 
-NormatizedCree = NewType("NormatizedCree", str)
+StandardizedCree = NewType("StandardizedCree", str)
 MatchedEnglish = NewType("MatchedEnglish", str)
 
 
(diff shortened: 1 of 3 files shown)


Madoshakalaka:
btw wrong commit message "extra words in test-db for regression", but the changed files are correct

eddieantonio:
Lol, is this a personal attack on @aarppe?

eddieantonio:
@aarppe: does it make sense to change all instances of "normatize" with "standardized"? This is in the context of orthography.

Madoshakalaka:
I guess normatized/standardized can be used interchangeably

Madoshakalaka:
@aarppe what is your opinion though

aarppe:
I didn't have a definitive answer to this question, as the two terms seem to be quite interchangeable (normative FST vs. SRO), though there appear to be some secondary differences. One speaks about the orthographical _norm_ or _standard_, where maybe norm is what has implicitly emerged and standard is what has been explicitly decided (i.e. Standard Roman Orthography).

eddieantonio:
We're going to close this for now, as our lab uses the term "normatize" in many different contexts. I don't want to conflate it with "standard" as in "standard Roman orthography", as that is a thing already.

PR #217, opened 2020-01-17
Migration workflow

Madoshakalaka:
We have decided to switch to a migration oriented workflow rather than building database locally from scratch and upload to sapir everytime we change the database.

`build-test-db` automation wasn't ready for that. Now it is. Specifically it now does the following steps.
- unlink test db
- apply existing migrations, API/0001 through API/0005
- build test_crkeng.xml test_engcrk.xml from res/test_db_words.txt
- import words to database from the built xml files.
- Apply all the rest migrations (including ones in the future)

The way it's used has not changed at all. i.e. `pipenv run build-test-db [--multi_processing N]`

This pull request also includes a light cleaning of old code and some extra documentation

#### Minor things included in this pr
3 more test words in the test database for regression testing
- maskos
- nimaskom
- niskak

a fix in regular expression that prevents lemma+N+A+Der/Dim to be recognized as NAD word (and the unit test for it)


CreeDictionary/DatabaseManager/__main__.py
```diff
@@ -48,18 +49,32 @@ def add_multi_processing_argument(ap: ArgumentParser):
...
+def import_and_migrate(xml_dir_name: Union[Path, str], process_count: int):
+    """
+    Unlink existing .sqlite3 file, apply migrations from API/0001 to API/0005
+    Import from specified engcrk.xml and crkeng.xml. Migrate the rest from all apps
+    """
+    try:
+        Path(settings.DATABASES["default"]["NAME"]).unlink()
+    except FileNotFoundError:
+        pass
+    call_command("migrate", "API", "0005")
```
eddieantonio: Why specifically are you hardcoding migrating 005?

CreeDictionary/res/test_db_words.txt
```diff
@@ -78,9 +85,8 @@ amiskwaciy-wâskahikan
...
 acâhkosa kâ-otakohpit
 
-# particles
-ispîhk
-ohci
-anohc
-pêyak
-nisto
+# https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/213
+# Cree word forms not getting some or all analyses
```
eddieantonio: 👍

CreeDictionary/utils/fst_analysis_parser.py
```diff
@@ -8,7 +8,7 @@
 
 
 analysis_pattern = re.compile(
-    r"(?P<category>\+N\+A(\+D)?|\+N\+I(\+D)?|\+V\+AI|\+V\+T[AI]|\+V\+II|(\+Num)?\+Ipc|\+Pron).*?$"
+    r"(?P<category>\+N\+A(\+D(?=\+))?|\+N\+I(\+D(?=\+))?|\+V\+AI|\+V\+T[AI]|\+V\+II|(\+Num)?\+Ipc|\+Pron).*?$"
```
eddieantonio: UHHHHHHHHH

`fst-lookup` returns the parts for you automatically, without you having to parse the string or need ad hoc, brittle regexps. Could you parsed the analysis _without_ using a regular expression?

CreeDictionary/DatabaseManager/__main__.py
Madoshakalaka: 0005 is the migration to date. This makes sense since we switched to the workflow mid way. Say if we started with this workflow, we'd still need to migrate 0001_initial.py to create a well formed database at first, then import all the words, and then all the rest migrations.

CreeDictionary/utils/fst_analysis_parser.py
Madoshakalaka: this is needed because we almost had to use hfstol to import xml files for the efficiency. We can of course always use `fst-lookup` everywhere else for a more robust code.

Madoshakalaka: but hey, I have a parametrized unit test for this!

eddieantonio: I'd like to see this ad hoc regular expression go away in future iterations, because the information that it's trying to extract is already available if we use the FST in a different way. Additionally, the FST can change its labels, and this will cause us nightmares when we have to deal with languages  that are _not_ Cree, e.g., Tsuut'ina, and Haida.

I will not block merge on this, but just FYI.

Madoshakalaka: @eddieantonio You need to think of this as an optimization here.

eddieantonio: Protip: do NOT tell your supervisors what to think.

Regardless if this is an "optimization" or not, it is a liability:

 - the regular expression is coupled with the extrinsic behaviour of the FST analysis string, which has and _can_ **change** as we patch things in the FST
   - this code must be in sync with changes to the FST, which adds a maintainability burden -- we have to remember to change this line if the FST gains new tags or changes its existing tags
 - the regular expression will not support FSTs for other languages, which @aarppe DOES want to see happening in the future
 - the regular expression is difficult to modify, as it is a lot of difficult to read code written all on one string. Have you considered using `re.VERBOSE`?
 - once again, the FST already has this information preparsed! We absolutely do not have to manually maintain a regular expression from scratch when the FST data structure already contains this information.

I give this regex a rating of:
not stonks

Madoshakalaka: whoa OK. Yes.

PR #240, opened 2020-02-04
user-firendly tags are here but ugly and goes out of screen sometimes

Madoshakalaka:
Screenshot from 2020-02-04 15-42-56

### DIFF
--- a/CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
+++ b/CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
@@ -25,7 +25,7 @@
                             <img class="definition-title__question-mark"
                                  src="{% static 'CreeDictionary/images/fa/question-circle-solid.svg' %}"
                                  alt="linguistic breakdown" data-cy="question-mark">
-                            <span class="definition-title__tooltip-text" data-cy="linguistic-breakdown">{% if result.linguistic_breakdown_head %}{{ result.linguistic_breakdown_head|join:" + " }} + {% endif %}<strong>{{ result.lemma_wordform.text }}-</strong> + {{ result.linguistic_breakdown_tail|join:" + " }}</span>
+                            <span class="definition-title__tooltip-text" data-cy="linguistic-breakdown">{% if result.linguistic_breakdown_head %}{{ result.linguistic_breakdown_head|join:" + " }} + {% endif %}<strong>{{ result.lemma_wordform.text }}</strong> + {{ result.linguistic_breakdown_tail|join:" + " }}</span>
                         </div>
                         {% endif %}
                     </dfn>
--- a/CreeDictionary/res/README.md
+++ b/CreeDictionary/res/README.md
@@ -2,7 +2,7 @@
 
 > Convenient variable `utils.shared_res_dir` is a `pathlib.Path` object that refers to this directory
 
-# `test_db_words.txt`
+## test_db_words.txt
 
 newline separated cree words that goes into `test_db.sqlite3`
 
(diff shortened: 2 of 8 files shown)


CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
```diff
@@ -25,7 +25,7 @@ <h2 class="definition-title definition-title--search-result">
                             <img class="definition-title__question-mark"
                                  src="{% static 'CreeDictionary/images/fa/question-circle-solid.svg' %}"
                                  alt="linguistic breakdown" data-cy="question-mark">
-                            <span class="definition-title__tooltip-text" data-cy="linguistic-breakdown">{% if result.linguistic_breakdown_head %}{{ result.linguistic_breakdown_head|join:" + " }} + {% endif %}<strong>{{ result.lemma_wordform.text }}-</strong> + {{ result.linguistic_breakdown_tail|join:" + " }}</span>
+                            <span class="definition-title__tooltip-text" data-cy="linguistic-breakdown">{% if result.linguistic_breakdown_head %}{{ result.linguistic_breakdown_head|join:" + " }} + {% endif %}<strong>{{ result.lemma_wordform.text }}</strong> + {{ result.linguistic_breakdown_tail|join:" + " }}</span>
```
eddieantonio: Good! We're showing _just_ the lemma now, instead of pretending it's the stem. @aarppe will be happy.

CreeDictionary/res/README.md
```diff
@@ -2,12 +2,16 @@
...
 
 Comments should be added using preceding hash sign.
 
 Empty lines are allowed and will be ignored
 
-After editing the file `$ USE_TEST_DB=true manage-db build-test-db [-m|--multi-processing N]` to rebuild the test database.
\ No newline at end of file
+After editing the file `$ manage-db build-test-db [-m|--multi-processing N]` to rebuild the test database.
+
+## crk.altlabel.tsv
```
eddieantonio: 👍

CreeDictionary/utils/fst_analysis_parser.py
```diff
@@ -1,17 +1,61 @@
...
 
 
+class LabelFriendliness(Enum):
+    LINGUISTIC_SHORT = 0
+    LINGUISTIC_LONG = 1
+    ENGLISH = 2
+    NEHIYAWEWIN = 3
+
+
+def read_labels() -> Dict[FSTTag, Dict[LabelFriendliness, Optional[Label]]]:
```
eddieantonio: I would return a bunch of `namedtuples` or frozen `attr` instances but I guess this will do. ¯\_(ツ)_/¯

aarppe:
Great, thanks! Being able to review all the new functionality now, it does indeed make sense that the relabeling of the linguistic breakdown (or ling. analysis) uses the same terminology as is shown in the paradigms. So using now plain English relabeling for the FST tags matches with the plain English column/row labels in the paradigms.

In practice, we'll probably want to use something like 'naming word' and 'action word' for nouns and words, so that we don't end up in the current reduplication with V='like: X' and TA='like: X, Y' - and it makes much intuitive sense to get 'Action word + like: wâpamêw'. There likely might be even better plain terms, which we are happy to use, when such are pointed to us.

Putting each of the labels on its own row and stacking them as a column seems like the immediate solution for the relabeling running of the screen. Though some dynamic solution will be optimal on the longer term. Indeed, Arok's feedback on the old itwêwina was that he found the random line change among the relabelings confusing and hard to parse, and that is true, cf.

image

Issue #209, opened 2020-01-15
Presentation of orthographically standardized form and linguistic breakdown


aarppe:
A. Search string _enipat_ -> single analysis, English definition only for lemma, no English definition for inflected form:

![image](https://user-images.githubusercontent.com/15003407/73315132-89e82080-41ec-11ea-976b-364de384eac8.png)

When clicking the `(?)` buttons, the following information becomes apparent (pops up):

![image](https://user-images.githubusercontent.com/15003407/73315155-9a989680-41ec-11ea-963a-553249482bc5.png)

B. Search string: ewapamat -> two analyses, same lemma, English definition only for lemma; no English definition for the inflected forms:

B1. **_First analysis:_**
![image](https://user-images.githubusercontent.com/15003407/73315226-c7e54480-41ec-11ea-9250-f36394d8c33f.png)

When clicking the `(?)` buttons, the following information becomes apparent (pops up):

![image](https://user-images.githubusercontent.com/15003407/73315283-eb0ff400-41ec-11ea-85b4-2046620855a5.png)

B2. _**Second analysis:**_
![image](https://user-images.githubusercontent.com/15003407/73315319-00851e00-41ed-11ea-84a2-f1b3115765a7.png)

When clicking the `(?)` buttons, the following information becomes apparent (pops up):

![image](https://user-images.githubusercontent.com/15003407/73315339-0da20d00-41ed-11ea-8b47-fba05f72e92c.png)

![image](https://user-images.githubusercontent.com/15003407/73315366-1c88bf80-41ed-11ea-827f-214cfa3d80dc.png)

C. Search string: _nicihciy_ -> one analysis, English definition for both inflected form + lemma

![image](https://user-images.githubusercontent.com/15003407/73315550-9ae56180-41ed-11ea-870b-5e88a54694fc.png)

When clicking the `(?)` button , the following information becomes apparent (pops up):

![image](https://user-images.githubusercontent.com/15003407/73315603-c2d4c500-41ed-11ea-9c41-545630deac40.png)

aarppe:
@eddieantonio Did my best understanding of the UI design, with the two cases above. I suppose each of these could be split with the (inflected) wordform in one box and the lemma in another, but I have to say that that starts getting somewhat confusing/repetitive, though I may be wrong.

Madoshakalaka:
[...]

PR #241, opened 2020-02-05
#209 actually asks for LINGUISTIC_SHORT rather than ENGLISH

Madoshakalaka:
Instead of this:

Screenshot from 2020-02-04 15-42-56

#209 actually asks for this:

Screenshot from 2020-02-04 19-22-13

also fixed a minor bug that causes some tags to not show


src/css/variables.css
```diff
@@ -41,6 +41,10 @@
   --link-active-color:      #0E6790; /* Currently active link. */
   --link-action-color:      #A568DB; /* A link that implies an immediate action. */
   --placeholder-text-color: #949494; /* The placeholder should be faint, but visible. */
+  /* special way of changing svg fill color when svg is embedded in <img> */
+  /* to change the color. Generate a "filter" attribute with the desired color code*/
+  /* https://codepen.io/sosuke/pen/Pjoqqp */
+  --blue-filter: invert(20%) sepia(20%) saturate(7344%) hue-rotate(220deg) brightness(90%) contrast(93%);
```
eddieantonio: Yeah, this works

PR #244, opened 2020-02-07
Preverb breakdown NOW

Madoshakalaka:
Screenshot from 2020-02-07 15-50-10

# So many Questions?

- Ugly

    I'll fix it later

- No Question Mark described in #209 

    I'll fix it later

- Why does the preverb _ê-_ NOT FREAKING have a link

    Screenshot taken with test database. _ê-_ is not in test database.

- Tests

    Will come up with tests

# Technical Changes

- new migration creates index on "full_lc", "pos" field so that we can finally exhaustivelly match preverbs **fast**

- Now preverbs are exhaustively matched and more perfect. Back then preverbs with diacritics couldn't get matched


aarppe:
@Madoshakalaka Only preverbs that are described in CW have the specific word class of `IPV` - preverbs in MD do not have that code, and can thus be ignored (for now). On the longer term, we'd want to present MD translations for preverbs, but we haven't done the comparison yet.

CreeDictionary/utils/fst_analysis_parser.py
```diff
@@ -56,7 +56,7 @@ def read_labels() -> Dict[FSTTag, Dict[LabelFriendliness, Optional[Label]]]:
 FST_TAG_LABELS = read_labels()
 
 
-def partition_analysis(analysis: str) -> Tuple[List[str], FSTLemma, List[str]]:
+def partition_analysis(analysis: str) -> Tuple[List[FSTTag], FSTLemma, List[FSTTag]]:
```
eddieantonio: 👍

Madoshakalaka:
@aarppe Yes it is on the way.

Issue #237, opened 2020-01-30
Linguistic breakdown popup (❓) should be keyboard-accessible


eddieantonio:
The linguistic breakdown popup (the thing selectable by hovering over the (❓) button) should be accessible via the keyboard, namely by pressing the <kbd>Tab</kbd> key.

Here is the current (incorrect) behaviour:

![tabbing 2020-01-30 10_03_58](https://user-images.githubusercontent.com/2294397/73472185-dc8a1f80-4347-11ea-8a3a-2bbad7fb240d.gif)

**EXPECTED**: when I press <kbd>Tab</kbd>  when the search button 🔍 is selected, then the (❓) should be selected. Pressing <kbd>Shift</kbd> + <kbd>Tab</kbd> when "form of nipâw" is selected should move focus to (❓). I'm not sure whether it's a good idea to automatically popup the linguistic breakdown as soon as it's selected, or to let it act as a button, which can be activated by pressing <kbd>Space</kbd>.

---

@kobexamoh: this can probably be done in the same pull request that fixes #209.

Madoshakalaka:
@eddieantonio very keyboard accessible now

PR #251, opened 2020-02-08
Preverb breakdown popups, popup responsiveness

Madoshakalaka:
# Splendid

- preverbs get tooltips now

Screenshot from 2020-02-08 04-41-44

- popups are intelligent now

    They wrap. They adapt, they don't do crap.

Screenshot from 2020-02-08 04-41-37

@kobexamoh also I'm not trying to steal yo job. I worked on preverb breakdown and found `poppers.js` is easily adapted to all other tooltips. You can really work futher based on the current CSS!

### DIFF
--- a/CreeDictionary/API/models.py
+++ b/CreeDictionary/API/models.py
@@ -105,6 +105,16 @@ def fetch_preverbs(user_query: str) -> Set["CreeResult"]:
     return result
 
 
+class Preverb(NamedTuple):
+    """
+    Contains information about a preverb, should be used inside templates
+    """
+
+    id: Optional[int]  # might be not in our database
+    text: str
+    definitions: Tuple[str, ...]  # might be not in our database, so might be empty
+
+
 @attrs(auto_attribs=True, frozen=True)  # frozen makes it hashable
 class SearchResult:
     """
@@ -133,7 +143,7 @@ class SearchResult:
 
     # Sequence of all preverb tags, in order
     # Optional: we might not have some preverbs in our database
-    preverbs: Tuple[Tuple[str, Optional[WordformID]], ...]
+    preverbs: Tuple[Preverb, ...]
 
     # TODO: there are things to be figured out for this :/
     # Sequence of all reduplication tags present, in order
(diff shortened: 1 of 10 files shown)


Madoshakalaka:
The coverage doesn't look good. I'll come up with tests under this PR later then.

CreeDictionary/API/models.py
```diff
@@ -133,7 +143,7 @@ class SearchResult:
 
     # Sequence of all preverb tags, in order
     # Optional: we might not have some preverbs in our database
-    preverbs: Tuple[Tuple[str, Optional[WordformID]], ...]
+    preverbs: Tuple[Preverb, ...]
```
Madoshakalaka: @eddieantonio could haved passed in Wordform. But I remembered the convention that dumb objects should be passed.

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
```diff
@@ -21,15 +21,22 @@ <h2 class="definition-title definition-title--search-result">
...
                         {% if result.linguistic_breakdown_head or result.linguistic_breakdown_tail %}
-                            <div class="definition-title__tooltip">
-                                <img class="definition-title__question-mark"
-                                     src="{% static 'CreeDictionary/images/fa/question-circle-solid.svg' %}"
-                                     alt="linguistic breakdown" data-cy="question-mark">
-                                <span class="definition-title__tooltip-text"
-                                      data-cy="linguistic-breakdown">{% if result.linguistic_breakdown_head %}
+
+
+                            <div tabindex="0" class="definition-title__tooltip-icon" aria-describedby="tooltip">
```
Madoshakalaka: fixes #237

src/css/_tooltip.css
```diff
@@ -0,0 +1,55 @@
+/* adapted from https://popper.js.org/docs/v2/tutorial/ */
+
+/* how to use: Create `icon` element and `popup` element in html
+
+`popup` needs to be the direct next sibling of `icon` and has class tooltip
+
+.tooltip class in this file decides how the popup looks like
```
Madoshakalaka: @kobexamoh I think you should work on this file to improve the looks of the popup.

CreeDictionary/API/models.py
eddieantonio: Dumb objects _only when rendering a view_! Passing in a wordform here is okay!

Madoshakalaka: We do use SearchResult instances directly while rendering entries

aarppe:
@Madoshakalaka Yes, now the linguistic breakdown/analysis and preverb pop-ups are actually genuinely usable. @kobexamoh So the next step is working on making the solution more general and fully working on different screen sizes/types, as well as polishing the output so that the linguistic breakdown is visually easy to perceive. E.g. AEW has commented that having one feature relabeling per line would be better than having some relabeling split on two lines. I don't have any definitive views on this, other than perhaps making the lemma/stem easily discernible among all the feature relabelings.

Issue #274, opened 2020-02-17
Cannot find âyiman in new dictionary contents


eddieantonio:
When using a newly build test DB, I can't find "âyiman" or "ayiman"

<img width="612" alt="search for âyiman" src="https://user-images.githubusercontent.com/2294397/74686016-6ed65400-518d-11ea-8225-00b92ed34acc.png">

But it's in the test database!

```
SQLite version 3.28.0 2019-04-15 14:49:49
Enter ".help" for usage hints.
sqlite> SELECT * FROM API_wordform WHERE text='âyiman';
7|âyiman|NI-1|âyiman+N+I|1|1|N|7
8|âyiman|VII-n|âyiman+V+II|1|1|V|8
```

The response is... empty?

This is causing errors in #272.

eddieantonio:
I wonder if this is due to a mismatch between the FST and the dictionary contents 🤔🤔🤔

eddieantonio:
I tried both the new and the old FST and they have the same lemma/POS

```python
# within CreeDictionary/res/fst:
>>> fst = FST.from_file('crk-descriptive-analyzer.fomabin')
[('âyiman', '+N', '+I', '+Sg'), ('âyiman', '+V', '+II', '+Ind', '+Prs', '+3Sg')]
>>> list(fst.analyze('ayiman'))
[('âyiman', '+N', '+I', '+Sg'), ('âyiman', '+V', '+II', '+Ind', '+Prs', '+3Sg')]
```

The generated dictionary content looks like this:

```xml
<?xml version="1.0"?>
<r>
	<e>
		<lg>
			<l pos="V">âyiman</l>
			<lc>VII-n</lc>
			<stem>âyiman-</stem>
		</lg>
	</e>
	<e>
		<lg>
			<l pos="N">âyiman</l>
			<lc>NI-1</lc>
			<stem>âyiman-</stem>
		</lg>
	</e>
</r>
```

Although I edited it for brevity, there's actually no `<mg>` or `<tg>` tags??!???!??!?! Is this an issue in the test data generator?

eddieantonio:
**UPDATE**: the reason that there is no definition for _âyiman_ in the dictionary is because the dictionary does not contain a definition for _âyiman_!

Did something go wrong in the creation of `crkeng.xml`?

---

SVN r187348:

```sh
$ sha384sum crkeng_cw_md_200208.xml
[...]

PR #275, opened 2020-02-17
Update test dictionary contents

eddieantonio:
This uses the latest `crkeng.xml` to generate the test database. It appears that some things have changed in the CW/MD comparisons, so at least one test had to change.

Please review the changed test to see if it still makes sense --- because it didn't really make sense to me!


eddieantonio:
Note: this PR is affected by #274. The solution here is to skip the test that searches for "âyiman" :/

eddieantonio:
I'm closing this until #274 is resolved.

PR #283, opened 2020-02-19
preverb search

Madoshakalaka:
# faster preverb match (supposedly)

The preverb search is faster by hashing all the preverbs prior to the start of the app.
I tested this new method on the real database. Actually It's not significantly faster (I wouldn't say even 10%). As  @eddieantonio discovers, one of the time consuming part is `fst-lookup` and one being the use of wrong indexes.

### DIFF
--- a/CreeDictionary/API/apps.py
+++ b/CreeDictionary/API/apps.py
@@ -1,6 +1,8 @@
 from django.apps import AppConfig
 from django.db import connection
 
+from utils.cree_lev_dist import remove_cree_diacritics
+
 
 class APIConfig(AppConfig):
     name = "API"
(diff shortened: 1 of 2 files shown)


CreeDictionary/API/apps.py
```diff
@@ -1,20 +1,46 @@
...
+from utils.cree_lev_dist import remove_cree_diacritics
+
 
 class APIConfig(AppConfig):
     name = "API"
 
     def ready(self):
         """
-        initialize fuzzy search (build the data structure)
+        This function is called prior to app start.
```
Madoshakalaka: This function should  be the place where expensive initialization happens.

CreeDictionary/API/models.py
```diff
@@ -161,6 +163,12 @@ def is_inflection(self) -> bool:
 class Wordform(models.Model):
     _cree_fuzzy_searcher = None
 
+    # this is initialized upon app ready.
+    # this helps speed up preverb match
+    # will look like: {"pe": {...}, "e": {...}, "nitawi": {...}}
+    # pure MD content won't be included
+    PREVERB_ASCII_LOOKUP: Dict[str, Set["Wordform"]] = defaultdict(set)
```
Madoshakalaka: the whole idea of this pr

Issue #247, opened 2020-02-08
Words takes too long to appear


Madoshakalaka:
After #244, some words take apparently longer time to be searched

e.g. search string `wapamat` takes a full 1.8 seconds for server to process.

This must be caused by the exhaustive match of preverbs. Currently, matching for preverbs in the database will happen for whatever query. 

We can skip the matching for non-preverbs with some simple hashing.

PR #290, opened 2020-02-21
Blazing fast again

Madoshakalaka:
After investigation, #247 was majorly caused by the use of the wrong database indexex.

After the creation of `pos` and `lc` indexes, `Wordform.filter` calls uses `pos` index for everything, even if the `Wordform.text` field is provided  (wtf django?)

My investigation shows there is no easy way of specifying a specific index to use except for resorting to raw SQL, not even RawSQL, which is different (see [django docs about database optimization](https://docs.djangoproject.com/en/3.0/topics/db/optimization/#do-database-work-in-the-database-rather-than-in-python)).

Fortuantely, `pos` and `lc` indexes were introduced to speed up on-the-fly preverb matching, which we don't need any more, as we now uses hashing to match preverbs!

So thoes two indexes were deleted, and Django were forced to use the index on `text` field again, and it makes everything blazing fast again. 

How fast? **100x** faster on the real database :boom:

Before, SQL profiling shows `/search/wapamaw` takes **377** ms on my machine.
After the removal of the indexes, it takes  **3** ms.


Madoshakalaka:
fixes #247

Madoshakalaka:
> #247 was majorly caused by the use of the wrong database indexes.
As @eddieantonio discovers, maybe the use of `fst-lookup` python package is the other equally-impacting factor. As profiling also shows for `/search/wapamat`, the total server CPU time still reaches 241 ms with SQL queries less than 3 ms

Issue #292, opened 2020-02-21
Address issues discovered in February 2020 design QA


eddieantonio:
We had a design QA session with Diane <time date="2020-02-19">a few days ago</time>. Here's a laundry list of issues to address:

 - [x] ~~"search for Cree or English" — placeholder text too big~~
 - [x] Search icon seems to be missing on production site #279 
 - [x] ~~Search icon on mobile should be left-aligned: offers affordance of "I can use this to search"~~
 - [x] ~~Search icon on tablet/desktop should be right-aligned: offers affordance of "I can click this to begin my search!"~~
 - [x] ~~prefix search should work~~ (future feature!)
 - [x] ℹ️ icon overflows its box
    - Use a `&nbsp;` between it and the (general) word class?
    -  needs margin
    - make (general) word class smaller?
 - [x] `<cite>` should be dark blue: `#182457`
 - [x] ~~Paradigm: fix issue with headings vs. column labels; headings should be left-aligned (blocked by #167) (see also: #193)~~
 - [x] ~~Paradigm table is misaligned in several places (blocked by #167)~~
 - [x] ~~there should be icons to explain the paradigm labels~~ (obsoleted by #363)
 - [x] row label & wordform should be `vertical-align: top`
 - [x] "show more" doesn't work — disable until we have the full paradigms
 - [x] `.paradigm-label--col { overflow-x: hidden }`
 - [x] More margin between head and its definitions.
 - [x] ~~There should be a circle around the ℹ️ icon when it's hovered/activated~~
 - [x] There should be a shadow under the box for the linguistic breakdown
 - [x] "Enable linguist mode" button should use `<button>` markup, but look like the "about" link
 - [x] update the copyright year — make it dynamic
 - [x] About page: margin between headings should be tighter to create a stronger visual hierarchy
 - [x] About page: links should have `font-weight: 700`
 - [x] logos should be laid out more carefully, using a 2x3 grid (#300)
 - [x] Search does not work on About page (crashing JavaScript) #280 
 - [x] ~~Search: preverbs should all display under a common heading~~
 - [x] decrease margin when there is no explicit definition
 - [x] change the colour of the 🔊 icon: `#286995` — shift in hue from the ℹ️ icon
 - [x] Tablet & desktop: the footer should use more horizontal space than vertical space
 - [x] There needs to be a clearer indication that a page is loading
 - [x] less margin between head and "form of {lemma}" text
 - [x] more margin between head and definition
[...]

PR #293, opened 2020-02-21
Fix a few design issues

eddieantonio:
Resolves a few bullet points in #292.

### DIFF
--- a/src/css/variables.css
+++ b/src/css/variables.css
@@ -54,7 +54,7 @@
   --branding-font-size:             1.50em; /* the title, e.g., "itwêwina" */
   --branding-subtitle-font-size:    0.75em; /* the subtitle, e.g., "Plains Cree Dictionary" */
   --definition-title-size:          2.00em;
-  --tooltip-icon-size:              0.75em; /* size of the ℹ️  button or 🔊 button */
+  --tooltip-icon-size:              1.75rem; /* size of the ℹ️  button or 🔊 button */
   --reference-to-lemma-size:        1.50em;
   --preverb-breakdown-size:         1.50em;
   --prominent-font-size:            1.50em; /* For text that should be prominent! */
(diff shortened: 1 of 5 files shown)


src/css/variables.css
```diff
@@ -53,7 +54,7 @@
   --branding-font-size:             1.50em; /* the title, e.g., "itwêwina" */
   --branding-subtitle-font-size:    0.75em; /* the subtitle, e.g., "Plains Cree Dictionary" */
   --definition-title-size:          2.00em;
-  --tooltip-icon-size:              2rem; /* size of the question mark or the exclamation mark. Kept the same across the app */
+  --tooltip-icon-size:              0.75em; /* size of the ℹ️  button or 🔊 button. Kept the same across the app */
```
Madoshakalaka:
```suggestion
  --tooltip-icon-size:              0.75em; /* size of the ℹ️  button or 🔊 button */
```

I used `em` at first too, then it turned out the icon beside the entry title and the preverb breakdown has different sizes. It looked kinda bad cuz they are not uniform. So I changed it to `rem` and added _Kept the same across the app_

Madoshakalaka: but it's ok if em produces the same icon size, or it's simply different aesthetics

Madoshakalaka: Screenshot from 2020-02-21 14-57-05

so yeah, one is bigger than the other. I guess it's OK? The mockup on Zeplin asked for the same size though.

eddieantonio: You're right:

Screen Shot 2020-02-21 at 2 58 25 PM

That _does_ look kinda dumb. `rem`s it is!

src/css/variables.css
```diff
@@ -54,7 +54,7 @@
   --branding-font-size:             1.50em; /* the title, e.g., "itwêwina" */
   --branding-subtitle-font-size:    0.75em; /* the subtitle, e.g., "Plains Cree Dictionary" */
   --definition-title-size:          2.00em;
-  --tooltip-icon-size:              0.75em; /* size of the ℹ️  button or 🔊 button. Kept the same across the app */
+  --tooltip-icon-size:              0.75em; /* size of the ℹ️  button or 🔊 button */
```
Madoshakalaka:
```suggestion
  --tooltip-icon-size:              0.75rem; /* size of the ℹ️  button or 🔊 button. Kept the same across the app*/
```

dErP

Madoshakalaka: wait, 0.75rem is actually too small. Please bumpt it up a bit

Issue #143, opened 2019-10-24
Use more permanent url for paradigm page


Madoshakalaka:
currently paradigm pages have url as: `/lemma/<lemma_id>` such as `/lemma/12311`

Lemma id is the id for the lemma stored in database. While they gaurantee one-to-one correspondence. They are highly subjective to change if we rebuild the database.

We should use  `/lemma/<wordform>` such as `/lemma/nipâw`

Madoshakalaka:
Since both `pos` and `lc` fields could be unspecified in the database,

I propose the following scheme for middle part of the url:
```
if <lc> specified:
    use <lc> (such as vii, vai, nid)
else:
    if <pos> specified
        use <pos> (such as v, n)
    else
        use "etc"
```

eddieantonio:
> If a url could match multiple lemmas, `pos` or `lc` or `analysis` or `id` should only be appended as url parameter only when necessary, each one of them gaurantees a increasingly stricter match with `id` that gaurantees a one-to-one match.
> 
> So the following are all legal urls:
> `/lemma/nipâw?pos=V`
> `/lemma/nipâw?lc=VAI`
> `/lemma/nipâw?analysis=nipâw+V+AI+Ind+Prs+3Sg`
> `/lemma/nipâw?id=302152`
> 
> All urls should fall back to search page if there are multiple matches or no lemma is matched.

I really like this suggestion! We'll add it to tasks, but prioritize it after the things we have to deal with now :/

Madoshakalaka:
fixed by #295

PR #295, opened 2020-02-22
More Semantic URL for Lemma Detail Page

Madoshakalaka:
This implements #143 

# User Friendly Lemma Page URL
URLs for lemma detail page (or the paradigm page) is more user friendly now, the linguistic term "lemma" is changed to "word", the url component used for matching is now "lemma_text" instead of "lemma_id"
```diff
- /lemma/123214/ 
+ /word/niskak/
```

`/word/<lemma_text>/` is tolerant of mismatches or ambiguities, it would redirect the user to the search page using `<lemma_text>` as search query if it didn't match anything in the database, or if it matched multiple lemmas

`/word/<lemma_text>/` allows additional constraints, e.g. 
- `/word/<lemma_text>/?pos=V` 
- `/word/<lemma_text>/?full_lc=NI`
- `/word/<lemma_text>/?id=123234`

Lemmas on the search page is aware of **the least strict unambiguous** URL, for most words, the links on the search page will only specify `<lemma_text>` alone, e.g.
`/word/nipâw/`

But for possibly ambiguous words, e.g. _pipon_, which is both a verb and a noun, the search page will distinguish them with relevant constraints, e.g.
- `/word/pipon/?pos=V`
- `/word/pipon/?pos=N`


CreeDictionary/API/models.py
```diff
@@ -127,8 +132,15 @@ class SearchResult:
...
 
     # The matched lemma
+    # todo: pass in dumb object instead to help serialization and avoid unnecessary database access
+    #   spread the fields of lemma_wordform here, and refactor word-entries.html
     lemma_wordform: "Wordform"
 
+    # looks like "/words/nipaw" "/words/nipâw?pos=xx" "/words/nipâw?full_lc=xx" "/words/nipâw?analysis=xx" "/words/nipâw?id=xx"
+    # it's the least strict url that guarantees unique match in the database
+    # see https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/143
+    lemma_url: str
```
Madoshakalaka: I could have avoided adding this fields but use {% url lemma_wordform %} in the template (Protip: this will actually call the `get_absolute_url` method of the instance). But I'm thinking of the "dumb object" convention.

CreeDictionary/API/models.py
```diff
@@ -169,6 +181,30 @@ class Wordform(models.Model):
     # pure MD content won't be included
     PREVERB_ASCII_LOOKUP: Dict[str, Set["Wordform"]] = defaultdict(set)
 
+    def get_absolute_url(self) -> str:
```
Madoshakalaka: [Django convention](https://docs.djangoproject.com/en/3.0/ref/models/instances/#get-absolute-url) to have this function. But hey! their example in the docs passes smart object in the template!!!

Screenshot from 2020-02-21 19-50-58

CreeDictionary/API/models.py
```diff
@@ -285,7 +321,7 @@ def save(self, *args, **kwargs):
         super(Wordform, self).save(*args, **kwargs)
 
     @staticmethod
-    def fetch_lemma_by_user_query(user_query: str) -> "CreeAndEnglish":
+    def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
```
Madoshakalaka: allow additional constraints to be specified

CreeDictionary/CreeDictionary/templates/CreeDictionary/index.html
```diff
@@ -10,7 +10,7 @@
             <h1 class="prose__heading"><span lang="cr">tânisi!</span></h1>
             <p><span lang="cr">itwêwina</span> is a Plains Cree Dictionary.</p>
             {% with long_word='ê-kî-nitawi-kâh-kîmôci-kotiskâwêyâhk' %}
-            {% url 'cree-dictionary-index-with-word' long_word as long_word_url %}
+            {% url 'cree-dictionary-index-with-query' long_word as long_word_url %}
```
Madoshakalaka: Rename

CreeDictionary/tests/DatabaseManager_tests/xml_importer_test.py
```diff
@@ -85,12 +86,10 @@ def test_import_xml_crkeng_small_common_xml_lemma_different_lc(shared_datadir):
     )
 
 
-def test_load_engcrk(one_hundredth_xml_dir):
-    res = load_engcrk_xml(one_hundredth_xml_dir / "engcrk.xml")
```
Madoshakalaka: not using `one_hundreth_xml` anymore

CreeDictionary/tests/conftest.py
```diff
@@ -56,33 +54,3 @@ def random_lemmas(draw) -> Wordform:
     """
     lemmas = Wordform.objects.filter(is_lemma=True, as_is=False)
     return draw(sampled_from(list(lemmas)))
-
-
-@pytest.fixture(scope="session")
-def one_hundredth_xml_dir(topmost_datadir) -> Path:
```
Madoshakalaka: We have a test database now, not using this anymore

Pipfile
```diff
@@ -11,10 +11,7 @@ pytest-cov = "~=2.7"
 pytest-datadir = "~=1.3"
 pytest-mypy = "~=0.4"
 hypothesis = {extras = ["django"],version = "~=4.34"}
-xml-subsetter = "~=0.0"
 codecov = "*"
-snakeviz = "~=2.0"
```
Madoshakalaka: snakeviz was used in the custom profiling code

Pipfile
```diff
@@ -11,10 +11,7 @@ pytest-cov = "~=2.7"
 pytest-datadir = "~=1.3"
 pytest-mypy = "~=0.4"
 hypothesis = {extras = ["django"],version = "~=4.34"}
-xml-subsetter = "~=0.0"
```
Madoshakalaka: no longer needed

Pipfile
```diff
@@ -11,10 +11,7 @@ pytest-cov = "~=2.7"
...
 codecov = "*"
-snakeviz = "~=2.0"
-paramiko = "~=2.6"
```
Madoshakalaka: ssh client utils, I installed it but never used it

Pipfile
```diff
@@ -39,7 +36,7 @@ sortedcontainers = {git = "https://github.com/Madoshakalaka/python-sortedcontain
 # look stuff up
 analyze = "hfst-optimized-lookup CreeDictionary/res/fst/crk-descriptive-analyzer.hfstol"
 generate = "hfst-optimized-lookup CreeDictionary/res/fst/crk-normative-generator.hfstol"
-# tests
+# unit tests
```
Madoshakalaka: wording

Pipfile
```diff
@@ -49,7 +46,6 @@ format = "black CreeDictionary"
 make-migrations = "python ./CreeDictionary/manage.py makemigrations"
 migrate = "python ./CreeDictionary/manage.py migrate"
 collect-static = "python ./CreeDictionary/manage.py collectstatic --clear --no-input"
-profile = "python -m profiling"
```
Madoshakalaka: old custom profiling script

cypress/integration/recordings.spec.js
```diff
@@ -12,7 +12,7 @@ context('Recordings', function () {
       cy.contains('a', 'wâpamêw')
         .click()
       cy.url()
-        .should('contain', '/lemma/')
+        .should('contain', '/word/')
```
Madoshakalaka: one day I'll figure out how to reverse urls in Cypress

src/index.js
```diff
@@ -184,6 +184,10 @@ function setupAudioOnPageLoad() {
         recording.play()
       })
     })
+    .catch(e=>{
+      // fixme (eddie): now what?
+      console.log(e)
+    })
```
Madoshakalaka: I added the catch because the console was showing annoying errors for many words that do not have recordings

CreeDictionary/API/models.py
eddieantonio: This adds more [accidental state](https://en.wikipedia.org/wiki/No_Silver_Bullet#Summary) to the databased, and I'm pretty wary of that. See my comment about dumb and smart objects.

CreeDictionary/API/models.py
eddieantonio: I think I should have been more clear what I meant by smart and dumb objects. The models can be plenty smart and _should_ have methods like `get_absolute_url()`. That said, **immediately before** the model gets passed to a template, it should be _serialized_.

So, there should be some code that converts the model into a dumb object that _includes_ the absolute URL.

Here's some pseudo-code of what I mean:

```python
def my_view(request):
    search_result = Wordform.search(...)
    assert not hasattr(search_result, 'lemma_url'), "this should not be explicitly stored!"
    assert search_result.get_absolute_url(), "this should work!"
    # serialize the result:
    dumb_object = search_result.to_dict()
    assert "lemma_url" in dumb_object
    assert dumb_object["lemma_url"] == search_result.lemma.get_absolute_url()
    return render("my-template.html", search_result=dumb_object)
```

CreeDictionary/API/models.py
```diff
@@ -169,6 +181,30 @@ class Wordform(models.Model):
...
+
+        return iri_to_uri(lemma_url)
+
+    @cached_property
+    def homograph_disambiguator(self) -> Optional[str]:
+        """
+        :return: the least strict field name that guarantees unique match together with the text field.
+            could be pos, full_lc, analysis, id or None when the text is enough to disambiguate
+        """
+        homographs = Wordform.objects.filter(text=self.text)
```
eddieantonio: This is pretty neat, and I understand the name—while requiring some good English vocabulary skills—makes perfect sense!

CreeDictionary/API/models.py
```diff
@@ -560,6 +611,7 @@ def get_preverbs_from_head_breakdown(
                         replace_user_friendly_tags(linguistic_breakdown_tail)
                     ),
                     lemma_wordform=cree_result.lemma,
+                    lemma_url=cree_result.lemma.get_absolute_url(),
```
eddieantonio: `SearchResult` should be able to figure out its own absolute URL!

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
```diff
@@ -43,7 +43,7 @@ <h2 class="definition-title definition-title--search-result">
             {% if result.is_inflection %}
                 <p class="reference-to-lemma" data-cy="reference-to-lemma">
                     Form of
-                    <a class="reference-to-lemma__lemma" href="{% url 'cree-dictionary-index-with-lemma' result.lemma_wordform.id %}"
+                    <a class="reference-to-lemma__lemma" href="{{ result.lemma_url }}"
```
eddieantonio: Even with my proposed changes, using `result.lemma_url` should remain the same. 👍

CreeDictionary/CreeDictionary/urls.py
```diff
@@ -31,8 +31,13 @@
 _urlpatterns = [
     # user interface
     ("", views.index, "cree-dictionary-index"),
-    ("search/<str:query_string>/", views.index, "cree-dictionary-index-with-word"),
-    ("lemma/<int:lemma_id>/", views.index, "cree-dictionary-index-with-lemma"),
+    ("search/<str:query_string>/", views.index, "cree-dictionary-index-with-query"),
+    # word is a user-friendly alternative for the linguistic term "lemma"
```
eddieantonio: 👍 👍 👍 👍

CreeDictionary/CreeDictionary/views.py
```diff
@@ -12,26 +12,44 @@
...
-    homepage with optional initial query or initial lemma. query_string and lemma_id can not both be given
+    lemma detail page. Fall back to search page if no lemma is found or multiple lemmas are found
+
+    :param request: accepts query params `pos` `full_lc` `analysis` `id` to further specify query_string
+    :param lemma_text: the exact form of the lemma (no spell relaxation)
+    """
+    extra_constraints = {
+        k: v for k, v in request.GET.items() if k in {"pos", "lc", "analysis", "id"}
+    }
+    lemma = Wordform.objects.filter(text=lemma_text, is_lemma=True, **extra_constraints)
```
eddieantonio: This is actually pretty neat!

CreeDictionary/CreeDictionary/views.py
```diff
@@ -12,26 +12,44 @@
...
+        lemma = lemma.get()
+        context = {
+            "lemma_id": lemma.id,
+            "lemma": lemma,
+            "paradigm_size": ParadigmSize.BASIC.display_form,
+            "paradigm_tables": lemma.paradigm if lemma else None,
+        }
+        return HttpResponse(render(request, "CreeDictionary/index.html", context))
+    else:
+        return redirect("cree-dictionary-index-with-query", query_string=lemma_text)
```
eddieantonio: I really like this redirect! 🤩🤩🤩

Could the logic actually be inverted with a [guard clause](https://wiki.c2.com/?GuardClause)?

e.g., 

```python
if lemma.count() != 1:
    # ambiguous results: search for the lemma instead:
    return redirect("cree-dictionary-index-with-query", query_string=lemma_text)

# this is the normal code:
lemma = lemma.get()
context = {...}
return HttpResponse(render(request, "CreeDictionary/index.html", context))
```

CreeDictionary/res/test_db_words.txt
```diff
@@ -99,3 +99,7 @@ niskak
 
 # https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/150
 kotiskâwêw
+
+# ambiguous examples (multiple pos)
```
eddieantonio: 👍

CreeDictionary/CreeDictionary/views.py
```diff
@@ -46,7 +64,7 @@ def search_results(request, query_string: str):  # pragma: no cover
     )
 
 
-def lemma_details(request, lemma_id: int):  # pragma: no cover
+def _lemma_details(request, lemma_id: int):  # pragma: no cover
```
eddieantonio: Could you please rename this to something like `lemma_details_internal` or something like that? Basically, a name that makes it clear that it's for internal use only and not a traditional view.

cypress/integration/recordings.spec.js
eddieantonio: lol, fair enough!

cypress/integration/routing.spec.js
```diff
@@ -0,0 +1,82 @@
...
+
+    // both results should be present
+    const x = cy.get('[data-cy=definition-title] a').first()
+      .should('have.attr', 'href')
+      .then(lemmaUrl=>lemmaUrls.add(x))
+
+
+    cy.get('[data-cy=definition-title] a').eq(1)
+      .should('have.attr', 'href')
+      .then(lemmaUrl=>lemmaUrls.add(lemmaUrl))
```
Madoshakalaka: `.eq(1)` is the way to specify _the second selected element_, compared to `.first()` that specifies the first element. But I should probably use `forEach` here.

cypress/integration/routing.spec.js
```diff
@@ -0,0 +1,82 @@
...
+  })
+
+  it('should redirect to search page if the lemma_text in /word/lemma_text matches multiple results', function () {
+
+    // pipon is a verb as well as a noun
+    cy.visit('/word/pipon/')
+    cy.get('[data-cy=paradigm]')
+      .should('not.exist')
+
+    // test if the redirection happens
```
eddieantonio: This is awesome!

src/index.js
eddieantonio: Fair enough!

CreeDictionary/tests/conftest.py
eddieantonio: I approve and appreciate this change, but I don't think it belongs in _this_ PR. Please strive to keep PRs super focused on one thing.

cypress/integration/routing.spec.js
eddieantonio: Good point. I didn't realize that's the logic here!

CreeDictionary/API/models.py
Madoshakalaka: our current SearchResult is a NamedTuple, isn't that already serialized?

eddieantonio:
Good point! I think that dumb objects are really supposed to be more like
dicts than tuples. Luckily, .to_dict() is already implemented for named
tuples!

CreeDictionary/CreeDictionary/views.py
Madoshakalaka: I will

CreeDictionary/API/models.py
```diff
@@ -147,13 +151,29 @@ class SearchResult:
 
     definitions: Tuple["Definition", ...]
 
-    @property
-    def is_inflection(self) -> bool:
+    def serialize(self) -> Dict[str, Any]:
```
Madoshakalaka: @eddieantonio This function introduces a lot of hidden attributes and structures in the return value. I have some ideas.

We should write an extra file `schema.py` with a [TypedDict](https://www.python.org/dev/peps/pep-0589/) that describes the serialized result. 
we do have access to TypedDict (try `from typing_extensions import TypedDict` :wink:)

for the time being, I prefer the first because it has higher readibility and IDE syntactical support.

The other proposal is to go full blown json, and write json schema object in with [python-jsonschema](https://github.com/Julian/jsonschema).

cypress/integration/routing.spec.js
```diff
@@ -0,0 +1,74 @@
...
+
+    // pipon is a verb as well as a noun
+    cy.visit('/search/pipon/')
+
+    let lemmaUrls = []
+
+    // both results should be present
+    cy.get('[data-cy=definition-title] a').each(($e)=>{
+      lemmaUrls.push($e.attr('href'))
+    }).then(()=>expect(lemmaUrls).to.have.members(['/word/pipon/?pos=N', '/word/pipon/?pos=V']))
```
Madoshakalaka: this is how it's done correctly.

CreeDictionary/API/models.py
eddieantonio: I'd prefer the first option for now ;)

CreeDictionary/API/models.py
```diff
@@ -147,13 +151,29 @@ class SearchResult:
...
         """
-        Is this an inflected form? That is, is this a wordform that is NOT the
-        lemma?
+        serialize the instance, can be used before passing into a template / in an API view, etc.
         """
-        return not self.is_lemma
+        # note: passing in serialized "dumb" object instead of smart ones to templates is a Django best practice
+        # it avoids unnecessary database access and makes APIs easier to create
```
eddieantonio: 👍

CreeDictionary/API/models.py
```diff
@@ -147,13 +151,29 @@ class SearchResult:
...
+        result["lemma_wordform"] = model_to_dict(self.lemma_wordform)
+        result["lemma_wordform"]["definitions"] = [
+            definition.serialize()
+            for definition in self.lemma_wordform.definitions.all()
+        ]
+
+        # looks like "/words/nipaw" "/words/nipâw?pos=xx" "/words/nipâw?full_lc=xx" "/words/nipâw?analysis=xx" "/words/nipâw?id=xx"
+        # it's the least strict url that guarantees unique match in the database
+        # see https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/143
+        result["lemma_url"] = self.lemma_wordform.get_absolute_url()
```
eddieantonio: This is excellent!

PR #299, opened 2020-02-25
Implement basic mixed paradigm

### LINKED ISSUE #277: Basic mixed paradigm
Should show 3 > 1,2 rather than 3', 3 would be much more common and thus the logical choice

eddieantonio:
I ran `update-layouts` to— update the layouts! — based on contributions from @aarppe. I also added a test for this.

Fixes #277

### DIFF
--- a/cypress/integration/regressions.spec.js
+++ b/cypress/integration/regressions.spec.js
@@ -171,4 +171,17 @@ context('Regressions', () => {
           .should('be.greaterThan', minimumWidth)
       })
   })
+
+  it.only('should show 3>1,2 rather than 3\', 3 in the VTA layout', function () {
+    // Go to a VTA word:
+    cy.visit('/search/wâpamêw')
+    cy.contains('a', 'wâpamêw')
+      .click()
+
+    // N.B.: This test will fail if the word changes. ¯\_(ツ)_/¯
+    cy.contains('th', 's/he → him/her/them (further)') // 3Sg -> 4 (lemma)
+    cy.contains('th', 's/he → me')                  // 3Sg -> 1Sg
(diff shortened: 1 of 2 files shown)


aarppe:
So, it seems that from a bird's eye perspective things are fine, as coverage does not change (no code should change) and the tests pass. Presumably I needn't look at the code, as that should not have changed. I can also find the new test specification created for the TA paradigm, but where do I find the video of the test, as that is perhaps the one thing I'd check as to the change working as it should (before approving)?

cypress/integration/regressions.spec.js
```diff
@@ -171,4 +171,17 @@ context('Regressions', () => {
...
       })
   })
+
+  it('should show 3>1,2 rather than 3\', 3 in the VTA layout', function () {
+    // Go to a VTA word:
+    cy.visit('/search/wâpamêw')
+    cy.contains('a', 'wâpamêw')
+      .click()
+
+    // N.B.: This test will fail if the word changes. ¯\_(ツ)_/¯
```
Madoshakalaka: ¯\_(ツ)_/¯

aarppe: I can see the test file here, but where's the video clip?

eddieantonio: That's a great question! Typically cypress comments and adds the video clip, but now you have to dig through the dashboard and find it. Here it is: https://dashboard.cypress.io/projects/8r2xra/runs/369/specs

eddieantonio: Honestly, the video clip is too fast to really notice anything :/

PR #300, opened 2020-02-25
Add SSHRC logo

eddieantonio:
Just adding the logo SVG from https://www.sshrc-crsh.gc.ca/about-au_sujet/sshrc-logo-crsh/index-eng.aspx.

I'm gonna tweak some CSS and then this will be ready to review.

@aarppe what should this logo link to?


aarppe:
The default answer would be to put that with the other logos. If we put it after the institutions, then we'll have two almost similar logos on top of each other. I suppose we could bookend the institutions with SSHRC first and NRC last. Somebody might complain about the Canadian flags, but it's Canadian money, and I don't know a better solution now. I just also reviewed the SSHRC guidelines (but didn't look into the linked law text), and we'd want to change 'funded' to 'supported' on our About page, but otherwise we should be fine.

eddieantonio:
> The default answer would be to put that with the other logos. If we put it after the institutions, then we'll have two almost similar logos on top of each other. I suppose we could bookend the institutions with SSHRC first and NRC last. Somebody might complain about the Canadian flags, but it's Canadian money, and I don't know a better solution now. I just also reviewed the SSHRC guidelines (but didn't look into the linked law text), and we'd want to change 'funded' to 'supported' on our About page, but otherwise we should be fine.

I meant, when you click the logo, where should it take you?

aarppe:
Here (somebody might complain about us choosing English, but at least that skips the annoying page about 'Terms and conditions of use'): https://www.sshrc-crsh.gc.ca/home-accueil-eng.aspx

kobexamoh:
LGTM. On mobile, the SSHRC logo does look a bit cramped – is that a problem or not something we need to worry about?

eddieantonio:
I wouldn't worry about it, but the federal government logos be like that :/

PR #302, opened 2020-02-25
Serialization refactoring

Madoshakalaka:
`Preverb`, as a part of `SearchResult` class, had a serialization function before #295, where serialization of the rest of the whole class was implemented.

This PR unifies the them and let them happend together immediately before passing into a template.


CreeDictionary/tests/API_tests/model_test.py
```diff
@@ -194,3 +195,44 @@ def deep_contain(container: Iterable, testee):
...
+    results = Wordform.search("nitawi-nipâw")
+    assert len(results) == 1
+    search_result = results.pop()
+
+    assert len(search_result.preverbs) == 1
+    assert search_result.preverbs[0].text == "nitawi-"
+
+
+@pytest.mark.django_db
+def test_search_text_with_ambiguous_word_classes():
```
Madoshakalaka: I initially used the name `test_search_text_with_ambiguous_pos()`, but then I remembered in the glossary, pos has been superceded by "word class".

This should be a good start of using the new glossary for me

CreeDictionary/API/models.py
```diff
@@ -548,7 +537,7 @@ def get_preverbs_from_head_breakdown(
             results = []
             for tag in head_breakdown:
 
-                result: Optional[Preverb] = None
+                preverb_result: Optional[Preverb] = None
                 if tag.startswith("PV/"):
                     # use altlabel.tsv to figure out the preverb
```
eddieantonio: I should have mentioned this way but this code is kind of weird with its usage of `altlabel.tsv` to figure out the normative form of the preverb 😬 

There isn't anything to address now, but I think it can and should be done better in the future.

CreeDictionary/API/models.py
```diff
@@ -74,43 +74,19 @@ def replace_user_friendly_tags(fst_tags: List[FSTTag]) -> List[Label]:
 WordformID = NewType("WordformID", int)  # the id of an wordform object in the database
 
 
-class Preverb(NamedTuple):
+def fetch_preverbs(user_query: str) -> Set["Wordform"]:
```
eddieantonio: I very much appreciate that `serialize` is no longer an argument to this! 👍

CreeDictionary/API/models.py
```diff
@@ -668,6 +652,10 @@ def get_preverbs_from_head_breakdown(
         return results
 
 
+# it's a str when the preverb does not exist in the database
+Preverb = Union[Wordform, str]
```
eddieantonio: Beauty! 😍

CreeDictionary/tests/API_tests/model_test.py
```diff
@@ -198,20 +198,20 @@ def deep_contain(container: Iterable, testee):
 
 
 @pytest.mark.django_db
-def test_search_serialization():
+@pytest.mark.parametrize("query", ["nipaw", "nitawi", "nitawi-nipaw", "e-nitawi-nipaw"])
```
eddieantonio: 👍

Issue #123, opened 2019-10-12
Dynamically switch orthographies


eddieantonio:
Switch orthographies from a drop-down menu.

Addresses the following user stories:

> * As a Cree language learner, I want to see all written Cree in Standard Roman Orthography with circumflexes (âêîô).
> * As a Cree language learner, I want to see all written Cree  in Standard Roman Orthography with macrons (āēīō).
> * As a Cree language learner, I want to see all written Cree in Western Cree syllabics.

aarppe:
For the sake of explicitness, this selection of orthography concerns Cree words anywhere in the application, not just in the paradigms or normatized search results, but in the interface as well - if I've understood correctly our discussions with Eddie.

eddieantonio:
> For the sake of explicitness, this selection of orthography concerns Cree words anywhere in the application, not just in the paradigms or normatized search results, but in the interface as well - if I've understood correctly our discussions with Eddie.

Correct. It should affect the display of Cree text EVERYWHERE!

eddieantonio:
I increased the priority of this, since I think it's pretty important to MESC. As well, I've been itching to implement it!

aarppe:
Yes, it is a very high MESC priority, and a tangible feature they will appreciate.

PR #309, opened 2020-03-03
Orthography selection menu

eddieantonio:
This implements the orthography switch menu.

GIF of using the orthography switch menu

For now, this **only affects the current page**. Reloading the page or navigating to a different page will reset the orthography to SRO/circumflex. Persisting the orthographical preference will occur in a different PR.

Addresses #123.

### DIFF
--- a/src/orthography.js
+++ b/src/orthography.js
@@ -9,14 +9,38 @@ const AVAILABLE_ORTHOGRAPHIES = new Set(['Cans', 'Latn', 'LatnXMacron'])
  * have the data-orth-switch.
  */
 export function registerEventListener() {
+  // Try to handle a click on ANYTHING.
+  // This way, if new elements appear on the page dynamically, we never
+  // need to register new event listeners: this one will work for all of them.
   document.body.addEventListener('click', function (evt) {
     let target = evt.target
+    // Determine that this is a orthography changing button.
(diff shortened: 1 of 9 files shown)


src/orthography.js
```diff
@@ -0,0 +1,91 @@
...
+ *    <ul>
+ *      <li class="menu-choice menu-choice--selected">
+ *        <button data-orth-switch value="ORTH">CURRENT ORTHOGRAPHY</button>
+ *      </li>
+ *      <li class="menu-choice">
+ *        <button data-orth-switch value="ORTH">DIFFERENT ORTHOGRAPHY</button>
+ *      </li>
+ *    </ul>
+ *  </details>
+ */
```
kobexamoh: Regarding the commented block from lines 53 - 28: are we keeping it for later usage or no? If no, perhaps we remove it?

eddieantonio: This commented code is not intended to be used! It's supposed to be documentation in order to follow what the JavaScript in `updateUI()` believes the HTML to look like. In general, I try to avoid committing commented out code.

Is there anyway I can make it clear that this code is documentation and, you know, actual real code?

kobexamoh: Ooh I see, my mistake! In that case, totally good – I'll blame it on me being rusty after a few days away 😅 s

PR #310, opened 2020-03-05
Set orthography in the request

### LINKED ISSUE #123: Dynamically switch orthographies

eddieantonio:
This is the last part of the three part epic of #123.

This adds a cookie called `orth` which affects the displayed orthography on all the pages. This cookie can be changed via a POST to `/change-orthography`. There are also two new tags to handle the change.

I also fought with CI a bit in this one so :/

Fixes #123.


CreeDictionary/CreeDictionary/settings.py
```diff
@@ -161,7 +161,17 @@
...
+
+
+if DEBUG:
+    # Use the default static storage backed for debug purposes.
+    STATICFILES_STORAGE = "django.contrib.staticfiles.storage.StaticFilesStorage"
+else:
+    # In production, use a manifest to encourage aggressive caching
+    # Note requires `python manage.py collectstatic`!
+    STATICFILES_STORAGE = (
+        "django.contrib.staticfiles.storage.ManifestStaticFilesStorage"
```
eddieantonio: This storage backend _requires_ `collectstatic` to be run before the URL reversing can work. Doing this avoids a crash when testing.

CreeDictionary/CreeDictionary/templatetags/creedictionary_extras.py
```diff
@@ -5,18 +5,52 @@
...
-from django.utils.html import conditional_escape
-from django.utils.safestring import mark_safe
+from django.utils.html import format_html
 
 CIRCUMFLEX_TO_MACRON = str.maketrans("êîôâ", "ēīōā")
 
 register = template.Library()
 
 
+@register.simple_tag(name="orth", takes_context=True)
```
eddieantonio: I made a new tag, as opposed to updating the existing filter, because filters cannot take in `context`. Since the `orth` cookie must be accessed through `context`, then this functionality _must_ be used from a tag.

cypress/integration/orthography.spec.js
```diff
@@ -41,5 +41,55 @@ describe('Orthography selection', function () {
...
+      cy.get('[data-cy=language-selector]')
+        .click()
+        .parent('details')
+        .as('menu')
+      cy.get('@menu')
+        .contains('Syllabics')
+        .click()
+
+      // The cookies should have changed.
+      cy.getCookie('orth')
```
eddieantonio: For some strange reason, the cookies change properly on my work laptop and this test continues and passes fine. BUT on Travis, the `orth` cookie is never set, and I have no idea why 🙃

CreeDictionary/CreeDictionary/templatetags/creedictionary_extras.py
```diff
@@ -29,19 +63,32 @@ def orth(sro_original: str):
...
+        inner_text = sro_circumflex
+    elif orthography == "Latn-x-macron":
+        inner_text = sro_macrons
+    elif orthography == "Cans":
+        inner_text = syllabics
+    else:
+        raise ValueError(f"invalid orthography: {orthography}")
 
-    return mark_safe(
+    return format_html(
```
eddieantonio: Django's documentation recommends this over using `conditional_escape()` and `mark_safe()`: https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.html.format_html

cypress/integration/orthography.spec.js
Madoshakalaka: is it caused by the **LatnXMacro** spelling issue?

eddieantonio: I don't think so, because that code path isn't touched at all! Instead, I think it's due to this: https://github.com/cypress-io/cypress/issues/4433.

Regardless, I changed all instances of `LatnXMacro` to `Latn-x-macron`. The reason it was in camelCase was because that's how the [data-\* API](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset) allows you to access values.

eddieantonio:
I've tried many things, but I can't seem to get that test to pass in Travis 😩 

@Madoshakalaka, please have another look. This one should be ready now!

cypress/integration/orthography.spec.js
```diff
@@ -41,5 +41,57 @@ describe('Orthography selection', function () {
...
+        .contains('Syllabics')
+    })
+
+    // XXX: This test fails in headless mode for Electron version < v6.0
+    // There was a bug with setting cookies via fetch():
+    //    https://github.com/cypress-io/cypress/issues/4433
+    // This should work in Cypress 3.5.0 and greater.
+    // As of 2020-03-05 this STILL doesn't work in CI :((((
+    it.skip('should persist my preference after a page load', function () {
+      cy.visit('/')
```
Madoshakalaka: How do you like
```javascript
if (Cypress.env('CI') === 'true'){
    this.skip()
}
```

so this beautiful test can still be run locally

eddieantonio: THAT'S BEAUTIFUL 🤩🤩🤩

Madoshakalaka:
I see I trapped you into another intrepid endeavor 😅
sorry I've never actually tried to use env vars on Cypress so I have no idea of the caveats

cypress/integration/orthography.spec.js
eddieantonio: > I see I trapped you into another intrepid endeavor 😅
> sorry I've never actually tried to use env vars on Cypress so I have no idea of the caveats

Sneaky sneaky, train!

Yeah, so `Cypress.env()` does not work in anyway that I expect. I've reverted those commits, and I'll add the smart skipping in a new PR, since this is getting out of hand 😓

Issue #120, opened 2019-10-12
KeyError in hfstol when searching for strings in quick succession 


eddieantonio:
```
Internal Server Error: /_search_results/ac/
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/travis/build/UAlbertaALTLab/cree-intelligent-dictionary/CreeDictionary/CreeDictionary/views.py", line 26, in search_results
    lemmas = Inflection.fetch_lemmas_by_user_query(query_string)
  File "/home/travis/build/UAlbertaALTLab/cree-intelligent-dictionary/CreeDictionary/API/models.py", line 158, in fetch_lemmas_by_user_query
    fst_analyses: Set[str] = descriptive_analyzer.feed_in_bulk_fast([user_query])[
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/hfstol/main.py", line 130, in feed_in_bulk_fast
    return self._call_hfstol(list(strings), multi_process)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/hfstol/main.py", line 214, in _call_hfstol
    results[original_input].add(res)
KeyError: 'a'
```

Madoshakalaka:
the heck. I'm on it.

eddieantonio:
Here's another example of it:

https://travis-ci.org/UAlbertaALTLab/cree-intelligent-dictionary/jobs/597054345#L386-L402

```
Internal Server Error: /_search_results/acâhko/
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
[...]

PR #323, opened 2020-03-11
Use the newest HFSTOL during Search

Madoshakalaka:
- The newest hfstol file is used
- test database is rebuilt with the newest hfstol and the newest crkeng.xml
- pipenv update

### DIFF
--- a/CreeDictionary/API/models.py
+++ b/CreeDictionary/API/models.py
@@ -32,7 +32,10 @@ from constants import POS, Analysis, FSTTag, Label, Language, ParadigmSize
 from fuzzy_search import CreeFuzzySearcher
 from paradigm import Layout
 from .schema import SerializedSearchResult, SerializedWordform, SerializedDefinition
-from shared import descriptive_analyzer_foma, normative_generator_foma
+from shared import (
+    descriptive_analyzer,
+    normative_generator,
+)
 from shared import paradigm_filler
 from utils import fst_analysis_parser, get_modified_distance
 from utils.cree_lev_dist import remove_cree_diacritics
@@ -357,12 +360,10 @@ class Wordform(models.Model):
         # 2. definition containment of the query word
         cree_results: Set[CreeResult] = set()
 
-        # utilize the spell relax in descriptive_analyzer
-        # TODO: use shared.descriptive_analyzer (HFSTOL) when this bug is fixed:
-        # https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/120
-        fst_analyses: Set[str] = set(
-            "".join(a) for a in descriptive_analyzer_foma.analyze(user_query)
-        )
+        fst_analyses: Set[Analysis] = {
+            Analysis(a)
+            for a in descriptive_analyzer.feed_in_bulk_fast([user_query])[user_query]
+        }
 
         all_standard_forms = []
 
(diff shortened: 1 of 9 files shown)


CreeDictionary/API/models.py
```diff
@@ -393,10 +394,12 @@ def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
...
                 # notice Err/Orth tags needs to be stripped because it makes our generator generate un-normatized forms
+                normatized_analysis = analysis.replace("+Err/Orth", "")
                 normatized_form_for_analysis = [
-                    *normative_generator_foma.generate(
-                        analysis.replace("+Err/Orth", "")
-                    )
+                    form
+                    for form in normative_generator.feed_in_bulk_fast(
+                        [normatized_analysis]
+                    )[normatized_analysis]
```
Madoshakalaka: what makes more sense is `normative_generator.feed(normatized_analysis)`, but sadly `.feed` from `hfstol` python package freezes with some wordforms. https://github.com/Madoshakalaka/hfstol/issues/1

CreeDictionary/API/models.py
```diff
@@ -357,12 +360,10 @@ def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
         # 2. definition containment of the query word
         cree_results: Set[CreeResult] = set()
 
-        # utilize the spell relax in descriptive_analyzer
-        # TODO: use shared.descriptive_analyzer (HFSTOL) when this bug is fixed:
-        # https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/120
```
Madoshakalaka: I released `hfstol-1.2.11` for #120, hopefully the bug is fixed
Screenshot from 2020-03-10 21-07-45

Madoshakalaka: good news: the ci passes

cypress/integration/regressions.spec.js
```diff
@@ -99,15 +99,6 @@ context('Regressions', () => {
       .should('contain', 'atâhk')
   })
 
-  // https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/176
-  describe('should show idiomatic lemmas', () => {
-    it('The Cree word ayiwinis should give you ayiwinisa as lemma', () => {
```
Madoshakalaka: the new hfstol gives "ayiwin" as the lemma, as the linguists decide "ayiwinisa" is not idiomatic enough.

cypress/integration/regressions.spec.js
```diff
@@ -129,13 +120,6 @@ context('Regressions', () => {
     )
   })
 
-  it('The Cree word ayiwinis should give you ayiwinisa as lemma', () => {
```
Madoshakalaka: these "idiomatic lemma" test doesn't make sense anymore, so we can just delete them. We had the issue cuz we used a lemma_tags.tsv as our lemma form guideline, (which say, decides the lemma of NI words should be N+I+Sg). The importing process has been changed and the lemma form depends solely on the fst we have.

CreeDictionary/API/models.py
eddieantonio: ¯\_(ツ)_/¯

PR #326, opened 2020-03-12
Dumb hfstol

eddieantonio:
Because the `hfstol` library has issues, I've started to implement a dumb version that simply calls `hfst-optimized-lookup` for every time it's called. This is inefficient but ¯\\\_(ツ)\_/¯ .

@Madoshakalaka, I've written the first part of this. Once this is merged, could you please use this in `models.py`?


CreeDictionary/CreeDictionary/hfstol.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
+# -*- coding: UTF-8 -*-
+
+"""
+Temporary HFSTOL.
```
eddieantonio: This code is uglyyyyy 🙈🙈🙈

Madoshakalaka: whatever works is beautiful

Madoshakalaka:
@eddieantonio Do you want to re run the tests after #327 is merged?

eddieantonio:
@Madoshakalaka: yup: rerunning now!

CreeDictionary/CreeDictionary/hfstol.py
```diff
@@ -0,0 +1,81 @@
...
+@contextmanager
+def write_file(text: str) -> Generator[IO[str], None, None]:
+    with TemporaryFile("w+", encoding="UTF-8") as tmp:
+        tmp.write(text)
+        tmp.flush()
+        tmp.seek(0)
+        yield tmp
+
+
+def analyze(wordform: str) -> Generator[Analysis, None, None]:
```
Madoshakalaka: aside from descriptive analyzer, normative generator is also used in the [search code] (https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/cf6057755f949501c5dc728fb2e1be55431f1c3e/CreeDictionary/API/models.py#L382-L403) to generate normative forms when the user gives analyzable words outside of the paradigm.

so we need a `generate` function. Do you think we need a separate file, a separate function in this same file or do you prefer other methods?

eddieantonio: I'll write a separate `generate()` function in this file!

Issue #268, opened 2020-02-14
Linguistic breakdown popup should not overlap icons ℹ️


eddieantonio:
And it shoudn't. This can probably be fixed with a `z-index`.

@kobexamoh might fix this in the process of fixing #266.

Madoshakalaka:
yep try z-index

kobexamoh:
Was this knocked off already at some point? I went to look into it a while back and made a note that (for me) it didn't seem like the breakdowns were overlapping icons.

Did a previous PR from someone take care of this issue or did I totally misread it while under the weather?

eddieantonio:
I can still replicate this in Firefox:

![Screen Shot 2020-03-13 at 9 33 01 AM](https://user-images.githubusercontent.com/2294397/76635833-ab3b6c80-650d-11ea-9cb6-ba60c65e30d3.png)

eddieantonio:
Fixed in #337.

Issue #266, opened 2020-02-13
Improving the formatting of the linguistic analysis pop-up


aarppe:
Following up on #209, we now have a practically usable presentation of the linguistic breakdown.

@kobexamoh The next step improving the formatting of the linguistic breakdown, e.g. presenting

1.  one feature per line on smaller screens, and/or

2. i) prefix features and ii) suffix features together and iii) the stem on a line of their own.

... and all this dynamically adapted based on screen size.

aarppe:
This is a meta-issue for a number of issues concerning the pop-up presentation of the relabeled analysis features:
1. Relabeling feature sequences: #348
2. Showing prefixal features on their own lines: #373
3. Showing the stem instead of the lemma in the pop-up: #235

PR #337, opened 2020-03-13
Visual formatting changes for linguistic breakdown popups, focus selectors.

kobexamoh:
Hola y'all!

Nothing excessive in here – just some updates to the formatting of the tooltips and focus styles on the icons. I think that together, these knock some of the issues in #266 (e.g. "one feature per line" and "the stem on a line on its own"). I think that one is a gradual work in progress, but if the aim is to get it in front of users, I'd like to believe this is a good first step in seeing what works (visually) and what doesn't. I think it's much more parsable this way, but would love feedback or critiques. 

I didn't write a test for the focus fix because it seems to me there's a test already written for it? If I've read that wrong then I can do that up.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/Users/kobexamoh/.local/share/virtualenvs/cree-intelligent-dictionary-8u1H7t5v/bin/python"
```
eddieantonio: Heads up! Please do not commit editor files in your PRs. I've gone ahead and placed `.vscode/` in the gitignore, so no need to worry about that in this PR!

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
```diff
@@ -31,7 +31,7 @@ <h2 class="definition-title definition-title--search-result">
                                         data-cy="linguistic-breakdown">{% if result.linguistic_breakdown_head %}
                                     {{ result.linguistic_breakdown_head|join:" + " }}
                                     + {% endif %}
-                                    <strong>{{ result.lemma_wordform.text }}</strong> + {{ result.linguistic_breakdown_tail|join:" + " }}</span>
+                                    <strong>{{ result.lemma_wordform.text|safe|add:":" }}</strong> <br />+ {{ result.linguistic_breakdown_tail|join:" <br />+ " }}</span>
```
eddieantonio: Why add `safe`? Is there better markup for making many lines than the `<br/>`? `<br>`s tend to be a last resort if there is absolutely no semantic meaning for a line break (or if the line break _is_ the semantic meaning!)

Perhaps instead of using `join` you could make an `<ol>` (the breakdown has an inherent order) and hide the numerals using CSS?

Madoshakalaka:
@kobexamoh you should fight for your human right to keep your .vscode folder (not really)

src/css/styles.css
```diff
@@ -449,6 +449,14 @@ html {
   vertical-align: middle;
 }
 
+.definition-title__tooltip-icon > img:focus {
+  outline-color: var(--search-focus-color);
+}
+
+.definition-title__tooltip-icon:focus {
+  outline-color: var(--search-focus-color);
+}
```
eddieantonio: Perhaps use [`:focus-within`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within) instead of having two rules?

```suggestion
.definition-title__tooltip-icon:focus, .definition-title__tooltip-icon:focus-within {
  outline-color: var(--search-focus-color);
}
```

eddieantonio: I also tested this in Firefox, and the outline doesn't actually appear without setting its width and colour? So here's the revised suggestion:

```suggestion
.definition-title__tooltip-icon:focus, .definition-title__tooltip-icon:focus-within {
  outline: 1px solid var(--search-focus-color);
}
```

.vscode/settings.json
kobexamoh: oof, thanks for catching that! I thought I had excluded it – my bad 😅

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
kobexamoh: `safe` was definitely a leftover from something I had tried previously – thank you for catching that! I removed it in the latest commit.
If I remember right, `join` was there previously: I didn't want to break the existing code structure too much.
I'll give the `<ol>` idea a shot – that seems sound.

src/css/styles.css
```diff
@@ -562,6 +570,15 @@ html {
   filter: var(--blue-filter); /* blue color */
 }
 
+/**
+ * TOOLTIP specific styling
+ *
+ * Should remove the numbering from the lists that come with the linguistic breakdowns
+ */
+.tooltip > span > ol {
```
eddieantonio: We're using BEM as a way to modularize our CSS. The reason to use BEM is that `span > ol` selectors are quite brittle (what happens if you remove the `<span>`?)

Here's a primer on how to create BEM classes: http://getbem.com/naming/

I think what would need to happen is that the `<ol>` would get a class like `.linguistic-breakdown` as its **BLOCK**, and these styles would apply to it.

src/css/styles.css
```diff
@@ -562,6 +570,15 @@ html {
...
+  list-style: none;
+  padding-left: 20px;
```
eddieantonio: Please change this to: a CSS custom property (CSS variable) defined in `variables.css`; then also please specify this in `rem` rather than `px` to be consistent with most other measures.

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
```diff
@@ -31,7 +31,14 @@ <h2 class="definition-title definition-title--search-result">
                                         data-cy="linguistic-breakdown">{% if result.linguistic_breakdown_head %}
                                     {{ result.linguistic_breakdown_head|join:" + " }}
                                     + {% endif %}
-                                    <strong>{{ result.lemma_wordform.text }}</strong> + {{ result.linguistic_breakdown_tail|join:" + " }}</span>
+                                    <strong>{{ result.lemma_wordform.text|add:":" }}</strong>  <ol>
```
eddieantonio: The `<ol>` should probably be on its own line, and be given a [BEM](http://getbem.com/naming/) class.

```suggestion
                                    <strong>{{ result.lemma_wordform.text|add:":" }}</strong>
                                    <ol class="linguistic-breakdown">                                    
```

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
```diff
@@ -31,7 +31,14 @@ <h2 class="definition-title definition-title--search-result">
                                         data-cy="linguistic-breakdown">{% if result.linguistic_breakdown_head %}
                                     {{ result.linguistic_breakdown_head|join:" + " }}
                                     + {% endif %}
-                                    <strong>{{ result.lemma_wordform.text }}</strong> + {{ result.linguistic_breakdown_tail|join:" + " }}</span>
+                                    <strong>{{ result.lemma_wordform.text|add:":" }}</strong>  <ol>                                    
+                                    {% for result in result.linguistic_breakdown_tail %}
+                                        <li>
+                                            {{ result }} +
```
eddieantonio: Remove the plus... that's just a leftover from the linguists :)

```suggestion
                                            {{ result }}
```

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
```diff
@@ -31,7 +31,15 @@ <h2 class="definition-title definition-title--search-result">
                                         data-cy="linguistic-breakdown">{% if result.linguistic_breakdown_head %}
                                     {{ result.linguistic_breakdown_head|join:" + " }}
                                     + {% endif %}
-                                    <strong>{{ result.lemma_wordform.text }}</strong> + {{ result.linguistic_breakdown_tail|join:" + " }}</span>
+                                    <strong>{{ result.lemma_wordform.text|add:":" }}</strong>
```
eddieantonio: Please add the colon to the template rather than using the `add` filter:

```suggestion
                                    <strong>{{ result.lemma_wordform.text }}:</strong>
```

The `add` filter [attempts to coerce both arguments to integers](https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#add), which is just weird...

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
```diff
@@ -31,7 +31,15 @@ <h2 class="definition-title definition-title--search-result">
...
+                                    <ol class="linguistic-breakdown">                                    
+                                        {% for result in result.linguistic_breakdown_tail %}
+                                            <li class="linguistic-breakdown__breakdown-tail" data-cy="linguistic-breakdown__breakdown-tail">
+                                                {{ result }}
+                                            </li>
+                                        {% endfor %}
```
eddieantonio: Please rename the `result` variable throughout this loop. I think a better name would be `linguistic_tag` or `explanation`; `result` is a bit too generic.

cypress/integration/search.spec.js
```diff
@@ -30,6 +30,40 @@ context('Searching', () => {
...
+      // begin from the homepage
+      cy.visit('/');
+
+      // lock onto the searchbar
+      cy.get('[data-cy=search]')
+      // get a word (nipaw)
+      .type('nipaw');
+      
+      // tab through the elements to force the tooltip to pop up
+      cy.get('[data-cy=information-mark]').first().click({force:true});
```
eddieantonio: What's the purpose of `{force: true}`? Does this work without `{force: true}`?

cypress/integration/search.spec.js
```diff
@@ -30,6 +30,40 @@ context('Searching', () => {
...
+
+      // see the linguistic breakdown as an ordered list
+      cy.get('[data-cy=linguistic-breakdown]').children().eq(1).should('have', 'li');
```
eddieantonio: This is fine, but I'm puzzled why this code gets 1-th element (second element in human speech). I think I'd write this using the two argument form of the `contains()` assertion:

```suggestion
      cy.get('[data-cy=linguistic-breakdown]').contains('li', 'Action word');
```

src/css/_tooltip.css
```diff
@@ -52,4 +52,17 @@ tooltip.js attaches all the event handlers
...
+ *
+ * Should remove the numbering from the lists that come with the linguistic breakdowns
+ */
+ .linguistic-breakdown__breakdown-tail {
+    list-style: none;
+  }
+  
+  .linguistic-breakdown__breakdown-tail > li {
+    padding-left: var(--small-pad);
+  }
```
eddieantonio: 👍

src/css/styles.css
```diff
@@ -449,6 +449,11 @@ html {
   vertical-align: middle;
 }
 
+.definition-title__tooltip-icon:focus,
+.definition-title__tooltip-icon:focus-within {
+  outline: 3px solid var(--search-focus-color);
+}
+
```
eddieantonio: 👍

cypress/integration/search.spec.js
```diff
@@ -30,6 +30,40 @@ context('Searching', () => {
...
+      cy.get('[data-cy=search]')
+      // get a word (use nipaw)
+      .type('nipaw');
+
+      // tab through the page elements until arriving on the '?' icon
+      cy.get('[data-cy=information-mark]').first().click({force:true});
+      
+      // it should trigger the focus icon's outline's focused state
+      cy.get('[data-cy=information-mark]').focus().should('have.css', 'outline', 'rgb(28, 157, 254) solid 3px'); // apparently, Cypress likes for the colours to be in the style of rgb() so we can't default to the CSS variables we've been using. Oof.
+    })
```
eddieantonio: This is a weird and fragile test, and will break if we decide to change some colours later. Is there a way to check for the outline without checking its colour? If not, we can keep this test as is. It will just be annoying if we ever decide to tweak colours. ¯\\\_(ツ)\_/¯

kobexamoh: Good eye! I just tested it as `.should('have.css', 'outline')` and that actually works well without testing for the colour, so I'll commit that 💪

cypress/integration/search.spec.js
kobexamoh: It does, yes! Thanks for spotting this.

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
kobexamoh: I ended up going with `linguistic_tag` – thanks!

kobexamoh:
> Also, could you see if you can sneak in a change to the `z-index` of the popups to fix #268?
Right, I pushed a commit to hit #268 as well: can that be brought into this easily enough, or should I make a new PR for it?

eddieantonio:
Yes! Feel free to bring in that change into this PR, no worries.

Issue #269, opened 2020-02-14
Use HFST to generate paradigms, if FOMA flookup / python-fst-lookup don't work


aarppe:
@Madoshakalaka Upon waiting for the resolution of #256, if FOMA flookup / python-fst-lookup don't work, use HFST for generating the inflected wordforms in the paradigms. That way, at least the paradigms will be correct, even if the recognition might not exactly work yet.

aarppe:
I've consulted Miikka on this, and his first diagnosis is that there might be an error in the HFST-to-FOMA conversion, rather than the original FST. He's looking into this, having received the original FSTs from me.

eddieantonio:
@Madoshakalaka, could you take a look at this?

eddieantonio:
The latest HFSTOL is merged into master. @Madoshakalaka, please merge master and use this to update the new database :D

Madoshakalaka:
:D

Issue #256, opened 2020-02-09
Update crk FST to match new version of crkeng_md_cw*.xml


aarppe:
In conjunction with #255, we need to update our crk FST to the latest version matching with the new dictionary content.

aarppe:
Ideally, we would have an semi-automatic pipeline for taking the latest version of the crk FST. But what we need to figure out for ourselves is whether the FST used by itwêwina is updated automatically when the FST is updated, or whether this needs to be initiated by a linguist/programmer.

Nevertheless, we would want a simple, single updating command for this, likely with some diagnostics to ensure that the functionalities of itwêwina won't break (-> new issue?).

aarppe:
This update is really important, as the latest FST implements fixes to the affixation of nouns (only shorter _-is_ diminutives, unless otherwise indicated by corpus evidence, plural-only nouns) and verbs (correct suffixes for several inflectional subtypes of TA verbs with some frequent lemmas such as _itêw_ - e.g. `2Sg>3SgO` imperative should be _isi_ with an _-s-_). In addition, it adds several thousand new lemmas, so that their inflected forms can be recognized and generated in paradigms.

Thus, it should be undertaken at our earliest convenience.

aarppe:
Examples of less desirable behavior that the new FST fixes:

1. Productive generation of double diminutives: search with _asikanis_ (or _askikanisis_), and receive both _asikanis_ and _asikanisis_. Since we have no corpus example of the longer form, only the shorter form, _asikanis_, should be recognized and presented.

![image](https://user-images.githubusercontent.com/15003407/74401271-fadb2b00-4ddd-11ea-9527-f13a6e7ad706.png)

2. Wrong imperative forms for TA verbs: iti is recognized and generated as the 2Sg>3SgO imperative for itêw, where isi would be the correct (there are also some other fixes to the VTA imperatives):

![image](https://user-images.githubusercontent.com/15003407/74401483-9c627c80-4dde-11ea-90fc-367ae506eb77.png)

aarppe:
@eddieantonio As for getting multicharacter symbols out of HFST (or XFST), I recalled that there is an option that sort of does that, the `print-space` variable toggled with the `-X` option, cf.

```
echo "me-miyo-miyo-na-nah-nipacik" | hfst-lookup -q -X print-space src/analyser-gt-desc.hfst
[...]

Issue #205, opened 2020-01-14
Minimum features required for first milestone of new itwêwina


aarppe:
@eddieantonio @Madoshakalaka In order for us to release the new itwêwina, we need to implement the following minimum set of features:

- [ ] Step 0 - use latest dictionary content (XML) and latest improved FST (#255, #256, and #269).
- [x] Step 1 - AEW compliance
   - [x] 1. Presentation of orthographically standardized forms of search strings: exemplified in #209 and #224 (and earlier detailed in #100 and #140). 
   - [x] 2. Presentation of linguistic analysis (breakdown) as user-friendly tags: exemplified in #209 and #224 (and previously detailed in #153 and #99)
      - [x] 2.1. Presentation of linguistic analysis features as user-friendly tags.
      - [x] 2.2. Presentation of preverbs in linguistic breakdown/analyses and their opening out.

- [x] Step 2 - MESC compliance
   - [x] 3. Presentation of spoken recording snippets linked to the dictionary entry: detailed in #128 and #149.

Everything else than the tasks 1-3 above is priority-wise secondary (we do want to support syllabics and we do want proper prefix search on the Cree side, but those can wait until the above are implemented). If they are resolved quickly, then we can progress to task 4 below.

- [x] Step 3 - old itwêwina will not be needed anymore
   - [x] 4. Syllabics and both flavors of SRO are supported in presenting Cree content: detailed in #123

aarppe:
A general note for interpreting the more detailed issues above. Since we've been working on the new itwêwina since the summer and our understanding of the functionalities has evolved over time, the earliest issues are best in addressing the overall general gist of the functionalities, and our views may and will have changed concerning whatever detail is presented. In this respect, the best reliable guidance on visual details can always be found in the very latest issue concerning a functionality.

We might well need to aggregate previous issues into new ones representing our current best understanding as well as laying out the details we've decided to focus on (alongside other nice things that might have turned obsolute or ones to be pursued later on with better time).

aarppe:
Moved into its separate issue #296: _0. Visual organization of standardized forms, lemmas, definitions, linguistic breakdowns, etc., as detailed in #224 and #232._

PR #350, opened 2020-03-17
Use new .hfstol files

Madoshakalaka:
The new `.hfstol` files is finally used in the search code.

Meanwhile test database is rebuilt with the newest `.hfstol` files. (It was done once and reverted.)

I expected a performance drop because of the new python implementation that repeatedly reads in `.hfstol` files, but It's is decently snapy when I test on my computer. So I have no complain whatsoever.

### DIFF
--- a/CreeDictionary/API/models.py
+++ b/CreeDictionary/API/models.py
@@ -32,7 +32,7 @@ from constants import POS, ConcatAnalysis, FSTTag, Label, Language, ParadigmSize
 from fuzzy_search import CreeFuzzySearcher
 from paradigm import Layout
 from .schema import SerializedSearchResult, SerializedWordform, SerializedDefinition
-from shared import descriptive_analyzer_foma, normative_generator_foma
+import CreeDictionary.hfstol as temp_hfstol
 from shared import paradigm_filler
 from utils import fst_analysis_parser, get_modified_distance
 from utils.cree_lev_dist import remove_cree_diacritics
@@ -363,8 +363,8 @@ class Wordform(models.Model):
         # utilize the spell relax in descriptive_analyzer
         # TODO: use shared.descriptive_analyzer (HFSTOL) when this bug is fixed:
         # https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/120
-        fst_analyses: Set[str] = set(
-            "".join(a) for a in descriptive_analyzer_foma.analyze(user_query)
+        fst_analyses: Set[ConcatAnalysis] = set(
+            a.concatenate() for a in temp_hfstol.analyze(user_query)
         )
 
         all_standard_forms = []
@@ -397,9 +397,7 @@ class Wordform(models.Model):
                 # now we generate the standardized form of the user query for display purpose
                 # notice Err/Orth tags needs to be stripped because it makes our generator generate un-normatized forms
                 normatized_form_for_analysis = [
-                    *normative_generator_foma.generate(
-                        analysis.replace("+Err/Orth", "")
-                    )
+                    *temp_hfstol.generate(analysis.replace("+Err/Orth", ""))
                 ]
                 all_standard_forms.extend(normatized_form_for_analysis)
                 if len(all_standard_forms) == 0:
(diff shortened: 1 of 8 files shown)


CreeDictionary/API/models.py
```diff
@@ -363,8 +363,8 @@ def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
         # utilize the spell relax in descriptive_analyzer
         # TODO: use shared.descriptive_analyzer (HFSTOL) when this bug is fixed:
         # https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/120
-        fst_analyses: Set[str] = set(
-            "".join(a) for a in descriptive_analyzer_foma.analyze(user_query)
+        fst_analyses: Set[ConcatAnalysis] = set(
+            a.concatenate() for a in temp_hfstol.analyze(user_query)
```
Madoshakalaka: I could have used `Set[Analysis]` instead of `Set[ConcatAnalysis]` here but that's going to be a big refactor (and hard actually), because in the database, analysis are stored in the format of concatenated analysis. Some lines below, we need to use both the analysis from this set with analysis from the database to construct search results. So for the time being I added a `.concatenate` method to Analysis class I made the code consistently use `ConcatAnalysis`.

CreeDictionary/constants/__init__.py
```diff
@@ -137,3 +137,10 @@ class Analysis(NamedTuple):
     raw_prefixes: str
     lemma: str
     raw_suffixes: str
+
```
Madoshakalaka: Does the following code in `utils/fst_analysis_parser.py` belong here?

https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/1e6479cc855f36706923d0c713a0bed0a6a912b3/CreeDictionary/utils/fst_analysis_parser.py#L59-L68

CreeDictionary/tests/conftest.py
```diff
@@ -14,7 +14,7 @@
 
 from DatabaseManager.xml_importer import import_xmls
 
-settings.register_profile("default", deadline=timedelta(milliseconds=1000))
+settings.register_profile("default", deadline=timedelta(milliseconds=2000))
```
Madoshakalaka: `test_search_for_exact_lemma` took 1500 ms in a run and `hypothesis` threw `DeadlineExceededError`.

Note it's not necessarily our searching is too slow. As commented before, the `random_lemma` custom strategy draws lemma in an inefficient way, causing the function to run for longer time.

cypress/integration/regressions.spec.js
```diff
@@ -99,14 +99,6 @@ context('Regressions', () => {
...
   })
 
-  // https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/176
-  describe('should show idiomatic lemmas', () => {
-    it('The Cree word ayiwinis should give you ayiwinisa as lemma', () => {
-      cy.visitSearch('ayiwinis')
-      cy.get('[data-cy=search-results]')
-        .should('contain', 'ayiwinisa')
-    })
-  })
```
Madoshakalaka: commented earlier these tests are not needed anymore

https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/pull/323#discussion_r390724428
https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/pull/323#discussion_r390724841

CreeDictionary/API/models.py
eddieantonio: Fair enough. And I like the name! :)

CreeDictionary/constants/__init__.py
eddieantonio: Like an `unconcatenate()`? Yeah, that would be good!

CreeDictionary/tests/conftest.py
eddieantonio: :( sounds like time for future improvement.

eddieantonio:
This will fix #269 and #256, I believe. As these are the only things remaining in #205, I think this will end up fixing #205 as well!

Madoshakalaka:
@eddieantonio fix em all

Madoshakalaka:
@eddieantonio yep I closed them all

Issue #351, opened 2020-03-17
Update from crkeng_cw_md_200218.xml to engcrk_cw_md_200226.xml


Madoshakalaka:
Antti has made a few additions to the file
see:
https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/255#issuecomment-600247687

The file is located at `/data/texts/crk/dicts/itwewina` on sapir

Madoshakalaka:
Outdated as there's a newer xml

Issue #325, opened 2020-03-12
Create straight-forward processes for updating dictionary FSTs and DB sources


aarppe:
Going forward, we would need to have in place straight-forward processes by which the analysis and generation FSTs and the dictionary database source(s) can be updated and the results diagnosed and validated (starting with `crk`). The diagnostics are needed to ensure that the new FSTs and new dictionary source do not break any instance of our intelligent dictionaries.

- [ ] update process in place for dictionary FSTs, for descriptive analysis and normative generation (also made use of in the generation of full paradigms for all entries in the dictionary source). Specified in #257.
- [ ] diagnostics in place for checking that updated FSTs work with enumerated test cases. Specified in #257.
- [ ] update process in place for (crk) dictionary source. Specified in #258.
- [ ] diagnostics in place for checking that dictionary source works with enumerated test cases. Specified in #258.

aarppe:
@andrewdotn I believe this has now in practice been implemented with or in conjunction with incremental import? Whatever remains might be commented here, but otherwise this and the associated issues can be considered closed?

andrewdotn:
See https://github.com/UAlbertaALTLab/morphodict/issues/258#issuecomment-932613386

Issue #258, opened 2020-02-09
Create a simple process for updating itwêwina dictionary content


aarppe:
Creating a more permanent solution for #255, since CW is being continuously updated, and we will be expanding MD based on the Cree recordings (and this involves a gradually increasing number of these spoken snippets matching dictionary entries), we would want a simple, straight-forward process for updating our crk lexical database providing content for itwêwina.

In this, we'd want some diagnostics as well.

aarppe:
@Madoshakalaka Based on today's discussion, this meta-issue should concern the following more specific tasks (these can be modified to be coding-wise more accurate):

- [ ] monitor the ALTLab repository for changes to `crkeng.xml`.
- [ ] automatically attempt importing `crkeng.xml` if that has been changed.
- [ ] check that tests are passed successfully before replacing old crk dictionary content within itwêwina, and deploying this new content.
- [ ] if tests are unsuccessful, keep to the previously imported crk dictionary content within itwêwina.

eddieantonio:
@Madoshakalaka please prepare:

 - [ ] PR for just-in-time database generation
 - [ ] PR for a management command, that when invoked, attempts to update `crkeng.xml`

aarppe:
@andrewdotn I believe this has now in practice been implemented with incremental import? Whatever remains might be commented here, but otherwise this and the associated issues can be considered closed?

andrewdotn:
The process for getting changes into the dictionary into production is documented [here](https://morphodict.readthedocs.io/en/latest/developers-guide.html?highlight=hour#importing-dictionaries) and [here](https://github.com/UAlbertaALTLab/crk-db/tree/main#steps-to-incrementally-update-the-production-database). I will add a link from the first to the second.

It is definitely feasible to very quickly make small changes to the dictionary. When the generation or phrase translation FSTs are updated, or when the set of paradigm cell templates is changed, a bigger reimport should be done, but it’s the same command with a different flag that just takes longer: about an hour. But you can start it and then check back later to see if it finished ok, you don’t have to bring the site down or do any extra manual steps at the end.

The process could definitely be streamlined and automated in the future.

Issue #257, opened 2020-02-09
Create a simple process for updating crk FST into itwêwina


aarppe:
Since the crk FST will be updated based on CW updates and modifications to the affixation, we would need a stream-lined process which will update the itwêwina FSTs (descriptive analyzer and normative generator) and the content that is generated with the FSTs (paradigm content).

As part of that process, we would need some diagnostics for checking that the change won't wreck the functionality of itwêwina. Likely, the paradigms in `giella/langs/crk/test/src/gt-norm-yamls`.

aarppe:
@andrewdotn @nienna73  I believe this has now in practice been implemented with or in conjunction with incremental import? Whatever remains might be commented here, but otherwise this and the associated issues can be considered closed?

aarppe:
The following XFSCRIPT code should generate the normative generator (with morpheme boundaries) and the descriptive analyzer for crk, from the elements existing in giellalt/lang-crk/ - What is needed is the latest full lexicon, in `lexicon.tmp.lexc` or `lexicon.hfst` / `lexicon.fomabin`, the phonological rules, in `phonology.xfscript`, and the composable version of the spell-relax rules, in `spellrelax.compose.hfst`.

```
read lexc src/fst/lexicon.tmp.lexc
# load src/fst/lexicon.hfst
define Morphology

source src/fst/phonology.xfscript
define Phonology

regex ~[ $[ "+Err/Frag" ]];
define removeFragments

regex ~[ $[ "+Err/Orth" ]];
define removeNonStandardForms

regex [ 0 <- "+Err/Orth" ];
define deleteErrOrthTag

regex ~[ $[ [ "+N" | "+V" ] ?* "+Err/Orth" ]];
define removeNonStandardNounVerbForms

regex $[ "+N" | "+V" | "+Ipc" | "+Pron" ];
define selectDictPOS

set flag-is-epsilon ON
regex [ selectDictPOS .o. removeNonStandardForms .o. removeFragments .o. Morphology .o. Phonology ];
save stack generator-gt-dict-norm.hfst
define NormativeGenerator

regex [ [ "<" | ">" | "/" ] -> 0 ];
define removeBoundaries

load src/orthography/spellrelax.compose.hfst
define SpellRelax

regex [ deleteErrOrthTag .o. selectDictPOS .o. removeFragments .o. Morphology .o. Phonology .o. removeBoundaries .o. SpellRelax ];
[...]

PR #352, opened 2020-03-17
Proposal: what's on prod

Madoshakalaka:
Before closing #256, I noticed the desired search results antti mentions was not showing up. I went on to investigate. It took me a long time to found out the production database is built with the  newer crkeng.xml but not the newer fst files.

It has come to me that it's very hard to determine what's being used in the database on Sapir, one of the complication is that the database depends on both the xml version and the fst version used to build it, which again can be accidentally differerent from vcs-tracked fst being used in the search code.

An emerging issue is Antti also needs the xml to be updated, see #351, 

```
# and we got this many xmls already
root@arrl-web003:/data/texts/crk/dicts/itwewina# ls
crkeng_cw_md_200208.xml  crkeng_cw_md_200218.xml  engcrk_cw_md_200226.xml  W_aggr_corp_morph_log_freq.txt

```

which demands the production database to be replaced manually. I understand we will switch to a solve-it-once-for-all new database. But I really want to write an API with minimal effort.

We should have an API `/status/`. It only accepts get request, and should output the following:

```json
{"Database": {"CrkengVersion": "200218", "FSTUsed": "2020-03-01"}, 
"Code": {"FSTUsed": "2020-03-01"} }
```

The value of "Database" key is obtained and uploaded to Sapir while newly built database is uploaed to Sapir. While `.Code.FSTUsed`  can be obtained by accessing `lasted modified` status of the fst file at runtime

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,6 @@ CreeDictionary/CreeDictionary/static/CreeDictionary/js/index.js
 CreeDictionary/CreeDictionary/static/CreeDictionary/css/styles.css
 *.css.map
 *.js.map
+
+# documents which version of xml file the production database is built with
+prod_crkeng_version.txt
--- a/Pipfile
+++ b/Pipfile
@@ -62,7 +62,7 @@ update-layouts = 'update-layouts'
 # warning: do not use these.
 # eventually we will scrap them and use a migration based work flow to preserve database changes AFTER a new
 # , well structured database is ready.
-push-db = 'rsync --progress CreeDictionary/db.sqlite3 $SAPIR_USER@sapir.artsrn.ualberta.ca:/opt/cree-intelligent-dictionary/CreeDictionary/db.sqlite3'
+push-db = 'sh ./libexec/push_db.sh'
 pull-db = 'rsync --progress $SAPIR_USER@sapir.artsrn.ualberta.ca:/opt/cree-intelligent-dictionary/CreeDictionary/db.sqlite3 CreeDictionary/db.sqlite3'
 
 [requires]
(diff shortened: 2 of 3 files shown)


Madoshakalaka:
The changed files is not complete.

aarppe:
@Madoshakalaka By design, we update neither FST nor dictionary XML automatically, because it is possible that there is a compilation error or because we explicitly do not want to use the latest FST for whatever reason.

The latest XML sources to be used have been explicitly indicated in the GitHub issues. We would want to have just single file names `crkeng.xml` and `engcrk.xml` that would be version controlled - since we don't have version control yet running on sapir, the files are designated by `YYMMDD` coding to indicate date.

I suppose we can keep special non-dated XML files which can be overwritten by the date-coded ones, and when that happens an automatic update will take place. However, we should ensure that if anything goes wrong in that update, one reverts back to the previous version instead of itwêwina becoming non-functional.

aarppe:
To be specific, the latest files are:

```
crkeng_cw_md_200314.xml
engcrk_cw_md_200317.xml
```

aarppe:
I'd love to use only one single file each for containing the crkeng.xml and engcrk.xml source, but currently we don't have version control on sapir (where I'd want to move the dictionary content, instead of Giella), so that is the reason I've been using dated XML file names.

Just to understand you correctly, wouldn't we yet need some way of importing the XML source into the dictionary database (in its current implementation), even if the database has information about the names of such XML files? Then, one could update the filenames of the XML sources to be imported in the Database, but the importation should still be implemented in some way.

We do have dedicated issues for this, namely #325, #257, and #258.

Madoshakalaka:
I'll close this for the time being, Eddie's suggestion makes sense.

PR #356, opened 2020-03-23
Remove Foma (for real this time!)

eddieantonio:
Since we literally are not using Foma at all for anything anymore, we can FINALLY delete it! And now that the latest HFST works, this speeds up boot times! :D


Madoshakalaka:
LGTM

Issue #133, opened 2019-10-16
Implement prefix search


aarppe:
Alongside potentially finding a lexical entry of which the entire search string (whatever characters have been typed up to any point) is an inflected form, one should also show lexical entries, the beginning of which match the search string - if these haven't already been matches through the linguistic analysis.

Since many lexical entries have lexical preverbs (over 200 but known), the prefix search should also try to match the beginnings of stems occurring after the preverbs.

In all cases, prefix search should ignore vowel length differences, so [a|â] should match [a|â], likewise for [e|i|o].

Example:

1 . Search string:
- nipaw

2a. Matched lexical entry (as a result of successful spell-relaxed analysis):
- nipâw

2b. Partially matched lexical entries (without skipping lexical prefixes): 
- nipaw ayiwinisa (Noun)
- nipawihaw (Verb)
- nipawâkan (Noun, NDA-1)
- nipâwin (Noun, NI-1)
- nîpawi (Verb)
- nîpawihêw (Verb, VTA)
- nîpawipayihow (Verb, VAI)
- nîpawiw (Verb, VAI)
- nîpawiwinihk (Noun, INM)
- nîpawîstam (Verb, VTI)
- nîpawîstamawêw (Verb, VTA)
- nîpawîstawêw (Verb, VTA)
- nîpawîstâtowak (Verb, VAI) 

2c. Partial matches (skipping lexical prefixes):
- matwê-nipâw (Verb, VAI)
- mâci-nipâw (Verb, VAI)
- mâmawi-nipâwak (Verb, VAI)
- mêkwâ-nipâw (Verb) 
- nôhtê-nipâw (Verb, VAI)
- pê-nipâw (Verb, VAI)
- pêyako-nîpawiw (Verb, VAI)
- têpi-nipâw (Verb, VAI)

The following below for later consideration, once the above has been satisfactorily implemented:

I'm not sure if we'd also want to skip reduplicative prefixes (Ca-, Câh, and Ca-Câh-), i.e.

Search: nanipaw or na-nipâw results in nipâw.

I'm also not sure if we'd want to consider skipping grammatical prefixes (ni-, ki-, kî-, wî, ê-, ka-, kâ, etc.)

aarppe:
The (new) user story for this is:

[...]

PR #359, opened 2020-03-25
On the fly affix search

Madoshakalaka:
This PR is a failure and is meant to inform that this is a terrible idea

This PR tries to implement #133 with minimal effort: a new field `ascii_text` (text field with diacritics stripped) is added to the database and `ascii_text__startwith` `ascii_text__endswith` is  used to do prefix/suffix search on the fly. This PR also creates ascii_text index (some say an index helps startswith/endswith performance)

It turns out to be an exordinary performance hit: tests locally on the smaller test database shows query string `nipaw` takes ~ 10 seconds to show any result.


Madoshakalaka:
@eddieantonio Oops I meant to make this a draft PR.

Madoshakalaka:
This is a failure and should be closed

PR #360, opened 2020-03-25
FAST affix search

Madoshakalaka:
@aarppe Hi please forgive me diverting from the priorities. But I get really intrigued by implementing a usable affix search tonight. I only spent this afternoon on this, and the result is really decent!

So prefix/suffix search will be conducted for user queries bigger than length 4 now (the length can be adjusted easily in code), for shorter results will possibly have too many matches .

I limited the search space with early hashing: Wordforms are classified by their n-th letters beforehand. The result is a lot better than #359 as search results show up generally fast when I test locally.

Here is the screenshot for _nipaw_, it's imperfect compared to the requirements in #133. But I guess that's due to other representation problems in our faulty database.  

Screenshot from 2020-03-24 23-22-16

### DIFF
--- a/CreeDictionary/API/apps.py
+++ b/CreeDictionary/API/apps.py
@@ -1,7 +1,10 @@
+from collections import defaultdict
 from pathlib import Path
 
 from django.apps import AppConfig
 from django.db import connection, OperationalError
+from typing import List, Dict, Set
+import string
 
 from utils import shared_res_dir
 from utils.cree_lev_dist import remove_cree_diacritics
@@ -62,6 +65,58 @@ def read_morpheme_rankings():
             Wordform.MORPHEME_RANKINGS[morpheme] = float(freq)
 
 
+def initialize_affix_search():
+    """
+    build dictionaries to facilitate affix search
+    """
+    # how many letters from the start/end do we check, larger length means more memory used, longer initialization time, but faster search
+    length = 5
+
+    n_th_letter_to_lemma_wordform_ids: List[Dict[str, Set[int]]] = [
+        defaultdict(set) for _ in range(length)
+    ]
+    inverse_n_th_letter_to_lemma_wordform_ids: List[Dict[str, Set[int]]] = [
+        defaultdict(set) for _ in range(length)
+    ]
+
+    from .models import Wordform
+
+    cree_letter_to_ascii = {
+        ascii_letter: ascii_letter for ascii_letter in string.ascii_lowercase
+    }
+    cree_letter_to_ascii.update(
+        {"â": "a", "ā": "a", "ê": "e", "ē": "e", "ī": "i", "î": "i", "ô": "o", "ō": "o"}
+    )
+    for wf in Wordform.objects.filter(is_lemma=True):
+        for i in range(length):
+            try:
+                lowered_letter = wf.text[i].lower()
+            except IndexError:
+                break
+            n_th_letter_to_lemma_wordform_ids[i][
+                cree_letter_to_ascii.get(lowered_letter, lowered_letter)
+            ].add(wf.id)
+
+        for i in range(length):
+            try:
+                lowered_letter = wf.text[-i - 1].lower()
+            except IndexError:
+                break
+
+            inverse_n_th_letter_to_lemma_wordform_ids[i][
+                cree_letter_to_ascii.get(lowered_letter, lowered_letter)
+            ].add(wf.id)
+
+    Wordform.N_TH_LETTER_TO_LEMMA_IDS = n_th_letter_to_lemma_wordform_ids
+    Wordform.INVERSE_N_TH_LETTER_TO_LEMMA_IDS = (
+        inverse_n_th_letter_to_lemma_wordform_ids
+    )
+
+    Wordform.WORDFORM_LEMMA_IDS = set(
+        Wordform.objects.filter(is_lemma=True).values_list("id", flat=True)
+    )
+
+
 class APIConfig(AppConfig):
     name = "API"
 
@@ -73,4 +128,5 @@ class APIConfig(AppConfig):
         """
         initialize_fuzzy_search()
         initialize_preverb_search()
+        initialize_affix_search()
         read_morpheme_rankings()
--- a/CreeDictionary/API/apps.py
+++ b/CreeDictionary/API/apps.py
@@ -1,13 +1,17 @@
+import logging
 from collections import defaultdict
 from pathlib import Path
 
 from django.apps import AppConfig
+from django.conf import settings
 from django.db import connection, OperationalError
 from typing import List, Dict, Set
 import string
 
 from utils import shared_res_dir
 from utils.cree_lev_dist import remove_cree_diacritics
+
+logger = logging.getLogger(__name__)
 
 
 def initialize_fuzzy_search():
@@ -67,10 +71,20 @@ def read_morpheme_rankings():
 
 def initialize_affix_search():
     """
-    build dictionaries to facilitate affix search
+    build dictionaries as Wordform class attributes to facilitate affix search
     """
+    from .models import Wordform
+
+    try:
+        lemma_wordforms = Wordform.objects.filter(is_lemma=True)
+    except OperationalError:
+        # when past version of the database is yet to migrate all the changes,
+        # we might not have Wordform table, is_lemma field in the database
+        return
+
+    logger.info("Building affix search dictionaries")
     # how many letters from the start/end do we check, larger length means more memory used, longer initialization time, but faster search
-    length = 5
+    length = settings.AFFIX_SEARCH_LENGTH
 
     n_th_letter_to_lemma_wordform_ids: List[Dict[str, Set[int]]] = [
         defaultdict(set) for _ in range(length)
@@ -79,18 +93,19 @@ def initialize_affix_search():
         defaultdict(set) for _ in range(length)
     ]
 
-    from .models import Wordform
-
     cree_letter_to_ascii = {
         ascii_letter: ascii_letter for ascii_letter in string.ascii_lowercase
     }
     cree_letter_to_ascii.update(
         {"â": "a", "ā": "a", "ê": "e", "ē": "e", "ī": "i", "î": "i", "ô": "o", "ō": "o"}
     )
-    for wf in Wordform.objects.filter(is_lemma=True):
+
+    for wf in lemma_wordforms:
+        lowered_wf_text = wf.text.lower()
+
         for i in range(length):
             try:
-                lowered_letter = wf.text[i].lower()
+                lowered_letter = lowered_wf_text[i]
             except IndexError:
                 break
             n_th_letter_to_lemma_wordform_ids[i][
@@ -99,7 +114,7 @@ def initialize_affix_search():
 
         for i in range(length):
             try:
-                lowered_letter = wf.text[-i - 1].lower()
+                lowered_letter = lowered_wf_text[-i - 1]
             except IndexError:
                 break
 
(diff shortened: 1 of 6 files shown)


cypress/integration/routing.spec.js
```diff
@@ -63,6 +63,6 @@ describe('urls for lemma detail page should be handled correctly', ()=>{
     // both results should be present
     cy.get('[data-cy=definition-title] a').each(($e)=>{
       lemmaUrls.push($e.attr('href'))
-    }).then(()=>expect(lemmaUrls).to.have.members(['/word/pipon/?pos=N', '/word/pipon/?pos=V']))
+    }).then(()=>expect(lemmaUrls).to.include('/word/pipon/?pos=N').and.to.include('/word/pipon/?pos=V'))
```
Madoshakalaka: now query `pipon` show more than 2 entries as a result of prefix/suffix search

CreeDictionary/tests/API_tests/model_test.py
```diff
@@ -234,8 +234,10 @@ def test_search_text_with_ambiguous_word_classes():
     """
     # pipon can be viewed as a Verb as well as a Noun
     results = Wordform.search("pipon")
-    assert len(results) == 2
-    assert {r.lemma_wordform.pos for r in results} == {"N", "V"}
+    assert {r.lemma_wordform.pos for r in results if r.matched_cree == "pipon"} == {
+        "N",
+        "V",
+    }
```
Madoshakalaka: now query `pipon` gives more than 2 results

CreeDictionary/API/models.py
```diff
@@ -356,10 +361,57 @@ def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
...
-        # 1. spell relax in descriptive fst
+        # 1. affix search (return all results that ends/starts with the query string)
+        # 2. spell relax in descriptive fst
         # 2. definition containment of the query word
+
         cree_results: Set[CreeResult] = set()
 
+        if (
+            len(user_query) > 4
+        ):  # there will be too many matches for some shorter queries
```
eddieantonio: It's SUPER WEIRD that Black chose to format it this way... what happens if you get rid of the outer parens?

CreeDictionary/API/apps.py
```diff
@@ -62,6 +65,58 @@ def read_morpheme_rankings():
             Wordform.MORPHEME_RANKINGS[morpheme] = float(freq)
 
 
+def initialize_affix_search():
+    """
+    build dictionaries to facilitate affix search
+    """
+    # how many letters from the start/end do we check, larger length means more memory used, longer initialization time, but faster search
+    length = 5
```
eddieantonio: I'm confused.... the length here is 5... but later the length is 4?

---

In addition, could you please extract this into a constant in `settings.py`? Some name like `CREE_DICTIONARY_AFFIX_SEARCH_LENGTH`.

CreeDictionary/API/models.py
```diff
@@ -356,10 +361,57 @@ def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
...
         # build up result_lemmas in 2 ways
-        # 1. spell relax in descriptive fst
+        # 1. affix search (return all results that ends/starts with the query string)
+        # 2. spell relax in descriptive fst
         # 2. definition containment of the query word
+
         cree_results: Set[CreeResult] = set()
 
+        if (
+            len(user_query) > 4
```
eddieantonio: Use the configurable constant `CREE_DICTIONARY_AFFIX_SEARCH_LENGTH` or whatever you decide to name it.

cypress/integration/search.spec.js
```diff
@@ -153,6 +153,15 @@ context('Searching', () => {
       .and('not.contain', 'Kill')
   })
 
+  it('should do prefix search and suffix search', () => {
+
+    cy.visitSearch('nipaw')
+
+    cy.get('[data-cy=search-results]')
+      .should('contain', 'nipawâkan')
+      .and('contain', 'mâci-nipâw')
```
eddieantonio: 👍

CreeDictionary/API/apps.py
```diff
@@ -62,6 +65,58 @@ def read_morpheme_rankings():
...
+    cree_letter_to_ascii = {
+        ascii_letter: ascii_letter for ascii_letter in string.ascii_lowercase
+    }
+    cree_letter_to_ascii.update(
+        {"â": "a", "ā": "a", "ê": "e", "ē": "e", "ī": "i", "î": "i", "ô": "o", "ō": "o"}
+    )
+    for wf in Wordform.objects.filter(is_lemma=True):
+        for i in range(length):
+            try:
+                lowered_letter = wf.text[i].lower()
```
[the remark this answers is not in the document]
Madoshakalaka: mind blowing

CreeDictionary/API/apps.py
```diff
@@ -62,6 +65,58 @@ def read_morpheme_rankings():
...
+    length = 5
+
+    n_th_letter_to_lemma_wordform_ids: List[Dict[str, Set[int]]] = [
+        defaultdict(set) for _ in range(length)
+    ]
+    inverse_n_th_letter_to_lemma_wordform_ids: List[Dict[str, Set[int]]] = [
+        defaultdict(set) for _ in range(length)
+    ]
+
+    from .models import Wordform
```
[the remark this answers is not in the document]
Madoshakalaka: yep I'll add a guard

CreeDictionary/API/models.py
Madoshakalaka: black insists to format this way and adds the parenthesis back.

That's cuz the comment is on the same line and black doesn't like any line exceding the maximum length. I moved the comment up a line and it looks a lot better.

CreeDictionary/API/apps.py
Madoshakalaka: what's more, this length and the later length is not neccessarily the same. e.g. We can check the beginning 3 letters and the ending 3 letters but we might choose to do the affix search for words longer than, say, 10 letters (which won't be the case but just to explain my point).

eddieantonio: Oh geez. Even more reason to give them both constant names! ;)

CreeDictionary/API/models.py
eddieantonio: Ah! I did not know that about Black. I tend to put this kind of comment above the `if`, so maybe that's why I've never encountered it!

CreeDictionary/API/models.py
eddieantonio: So... I guess there are two constants:

 - min word length for affix search?
 - affix length?

Oh dear.

aarppe:
Is there a lay'ish explanation how this algorithm used in this fast prefix search works in practice? I see a note above about indexing based on the n'th character, but is that just the n'th character, of starting with that n'th one?

CreeDictionary/API/apps.py
```diff
@@ -62,6 +69,66 @@ def read_morpheme_rankings():
...
+    build dictionaries as Wordform class attributes to facilitate affix search
+    """
+    from .models import Wordform
+
+    try:
+        lemma_wordforms = Wordform.objects.filter(is_lemma=True)
+    except OperationalError:
+        # when past version of the database is yet to migrate all the changes,
+        # we might not have Wordform table, is_lemma field in the database
+        return
```
Madoshakalaka: since app.py will also get called during migration

CreeDictionary/API/models.py
```diff
@@ -356,10 +361,68 @@ def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
...
+            no_diacritics_query_string = remove_cree_diacritics(user_query)
+            possible_wf_ids = Wordform.N_TH_LETTER_TO_LEMMA_IDS[0][
+                no_diacritics_query_string[0]
+            ].copy()
+            for letter_index, letter_to_ids in enumerate(
+                Wordform.N_TH_LETTER_TO_LEMMA_IDS
+            ):
+                if letter_index == 0:
+                    # accounted for before the for loop
+                    continue
```
Madoshakalaka: In a past commit I used All_IDS & LETTER_1 & LETTER_2 & LETTER_3 .... The ALL_IDs part is redundant and a waste of performanc.

CreeDictionary/CreeDictionary/settings.py
```diff
@@ -151,6 +149,13 @@
 
 USE_TZ = True
 
+############################## API app settings ###############################
+
+# how many letters from the start/end do we check, larger length means more memory used, longer initialization time, but faster search
+AFFIX_SEARCH_LENGTH = 4
+# We only apply affix search for user queries longer than the threshold length
+AFFIX_SEARCH_THRESHOLD = 4
+
```
Madoshakalaka: yes

Madoshakalaka:
@aarppe here's an explanation for prefix search:

The program builds the following four tables before hand:

Table 1:
---
words that has letter a in their first letter: {apple, ash ...}
words that has letter b in their first letter: {back, bass, ...}
words that has letter c in their first letter: {cat, cease...}
words that has letter .... in their first letter: {......}

Table 2:
---
words that has letter a in their **second** letter: {kappa, radish, ...}
words that has letter b in their **second** letter: {.........}

.
.
.
.
Table 4:
---
words that has letter a in their **fourth** letter: {....}
... 

then for each user query, We use the first letter to look up possible words in Table 1, we use the second letter to look up in Table 2, the third letter in table 3, the fourth letter in table 4. In total, we get 4 sets of possible words. Finally the intersection of the sets is computed, which will only contain possibly a handful of possible words. Then we can check one by one in the intersection.

The algorithm is fast, in that these tables are hash tables and the action of looking up is very fast (O(1) time complexity), also the action of intersection is reasonably fast. We still need to check one by one at the end but the search space is greatly reduced.

CreeDictionary/CreeDictionary/settings.py
```diff
@@ -151,6 +149,13 @@
 
 USE_TZ = True
 
+############################## API app settings ###############################
+
+# how many letters from the start/end do we check, larger length means more memory used, longer initialization time, but faster search
+AFFIX_SEARCH_LENGTH = 4
```
Madoshakalaka: @aarppe The number of tables that will be built is configurable here.

aarppe:
@Madoshakalaka Ok, but can you twist out of iron wire - to use a Finnish metaphor - how this works for prefix search?

So, do we use the first character to look up words in the later tables, e.g. Table 6 for mâci-nipâw, when searching with n[ipâ] (since you need to match the 6th character -n- in that case)?

CreeDictionary/API/models.py
```diff
@@ -356,10 +361,68 @@ def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
...
 
         # build up result_lemmas in 2 ways
-        # 1. spell relax in descriptive fst
+        # 1. affix search (return all results that ends/starts with the query string)
+        # 2. spell relax in descriptive fst
         # 2. definition containment of the query word
+
         cree_results: Set[CreeResult] = set()
 
+        # there will be too many matches for some shorter queries
```
eddieantonio: 👍

CreeDictionary/CreeDictionary/settings.py
```diff
@@ -194,14 +199,30 @@
         },
         "write_info_to_console_dev": {
             "level": "INFO",
-            "filters": ["require_debug_true"],
+            # without require_run_main_true, loggers from API.apps will print twice
+            "filters": ["require_debug_true", "require_run_main_true"],
```
eddieantonio: Weird. But okay!

Madoshakalaka:
@aarppe we only build four tables to limit the results to a manageable set. An exhastive check will still be performed afterwards, e.g. if the user searches for __startle__, we will use the initial four letters: s, t, a, r and yield: {star, stare, starlet, starlettify, stark, stary, starte, startling, startled}. Then every word in the intersection is still checked for whether they really starts with __startle__, in this case, only "startled" makes to the final list

Madoshakalaka:
@aarppe Eddie proposed a better solution by building a trie for the Cree vocabulary. I think that's more intuitive and maybe more efficient in time complexity with a proper python library. I'll explore later if necessary but not now.

Issue #355, opened 2020-03-19
Preverb translations have mixed up


aarppe:
Using our example wordform _ê-kî-nitawi-kâh-kîmôci-kotiskâwêyâhk_, all the preverbs get weird and incorrect definitions/translations, cf.

![image](https://user-images.githubusercontent.com/15003407/77111626-48564380-69ed-11ea-9bab-ee00e0a7f792.png)

These all appear to have multiple definitions, when there should be only one. And among the multiple translations, one appears to be correct, but the others entirely incorrect. Those incorrect definitions do not come out of the XML source, I've just checked.

![image](https://user-images.githubusercontent.com/15003407/77111656-54da9c00-69ed-11ea-9bc0-742cfae52adb.png)

![image](https://user-images.githubusercontent.com/15003407/77111686-6328b800-69ed-11ea-9e0c-c88b8aa4bdbb.png)

![image](https://user-images.githubusercontent.com/15003407/77111762-7e93c300-69ed-11ea-96bc-496c0a855657.png)

![image](https://user-images.githubusercontent.com/15003407/77111732-73409780-69ed-11ea-9526-9aabfb831bdc.png)

aarppe:
@Madoshakalaka Drawing your attention to this weird issue that Arok noticed today.

Madoshakalaka:
@aarppe sure thing

eddieantonio:
@Madoshakalaka Please have a look at this soon. I want to record demos, and I can't demo the preverb feature because literally all of the preverbs are wrong! 😩

aarppe:
@Madoshakalaka This issue has a higher priority than the spoken paradigms (though that is important as well, but can wait the little while until this one is solved).

Madoshakalaka:
@aarppe Looking into it right now

Madoshakalaka:
@aarppe @eddieantonio I can't reproduce any of it both locally and in prod?? what is happening???

aarppe:
The error was there when I reported it, but seems to have gotten at least partially fixed. I don't know why or by whom, but we should know explicitly why that happened, and potentially got fixed by itself.

Nevertheless, _maci-_ still has one duplicate entry, where the readings _maci-_ 'evil, etc.' and _mâci-_ 'begin etc.' are both correct, cf.

![image](https://user-images.githubusercontent.com/15003407/77493850-d07f8300-6e09-11ea-9aca-7356ba6905e9.png)

[...]

PR #361, opened 2020-03-25
regression test for #355 preverb issue

Madoshakalaka:
Since a hallmark of #355 is an extra entry with un-related translation, this regression test notifies the developer when future alike issues happen.

### DIFF
--- a/cypress/integration/regressions.spec.js
+++ b/cypress/integration/regressions.spec.js
@@ -212,7 +212,7 @@ context('Regressions', () => {
 
     // there should be only one result
     cy.get('[data-cy=search-results]').should('have.length', 1)
-
+      .and('should.contain', 'go and')
   })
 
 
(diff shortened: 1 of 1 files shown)


cypress/integration/regressions.spec.js
```diff
@@ -200,4 +200,20 @@ context('Regressions', () => {
...
+   * Ensure preverbs don't get weird search results
+   *
+   * See: https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/355
+   */
+  it('should not present un-related translation for preverbs', function () {
+    cy.visitSearch('nitawi-')
+
+    // there should be only one result
+    cy.get('[data-cy=search-results]').should('have.length', 1)
+
```
eddieantonio: So the issue was that the _pop-ups_ were not getting the right definition. "nitawi-" should have a pretty constant definition containing the words "go and", so could you add an assertion for the content as well? Something like:

```suggestion
      .and('should.contain', 'go and')
```

cypress/integration/regressions.spec.js
```diff
@@ -200,4 +200,20 @@ context('Regressions', () => {
...
+   */
+  it('should not present un-related translation for preverbs', function () {
+    cy.visitSearch('nitawi-')
+
+    // there should be only one result
+    cy.get('[data-cy=search-results]').should('have.length', 1)
+      .and('contain', 'go and')
+  })
+
+
```
eddieantonio:
```suggestion
```

PR #368, opened 2020-03-27
Use Trie to Do Prefix/Suffix Search

Madoshakalaka:
[marisa-trie](https://github.com/pytries/marisa-trie) is a python wrapper to a C library. The library implements Trie data structure with convinient prefix search utilities.

I re-implemented prefix/suffix search with tries. Based on my profiling, the end result is not necessarily faster because the bottleneck happens at database access. (The trie look up finishes in e^-5 seconds while the database access takes from 100 to 500 ms depending on how many results we present)

However the code is definitely more readible.

### DIFF
--- a/CreeDictionary/API/apps.py
+++ b/CreeDictionary/API/apps.py
@@ -8,6 +8,7 @@ from django.db import connection, OperationalError
 from typing import List, Dict, Set
 import string
 
+from .affix_search import AffixSearcher
 from utils import shared_res_dir
 from utils.cree_lev_dist import remove_cree_diacritics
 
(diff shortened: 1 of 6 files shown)


CreeDictionary/API/apps.py
```diff
@@ -71,62 +72,31 @@ def read_morpheme_rankings():
...
     cree_letter_to_ascii = {
         ascii_letter: ascii_letter for ascii_letter in string.ascii_lowercase
     }
     cree_letter_to_ascii.update(
         {"â": "a", "ā": "a", "ê": "e", "ē": "e", "ī": "i", "î": "i", "ô": "o", "ō": "o"}
     )
+    try:
+        lowered_no_diacritics_text_with_id = [
+            ("".join([cree_letter_to_ascii.get(c, c) for c in text.lower()]), wf_id)
+            for text, wf_id in Wordform.objects.filter(is_lemma=True).values_list(
```
eddieantonio: :+1: I never knew about `.values_list()`!

CreeDictionary/API/models.py
```diff
@@ -369,59 +371,14 @@ def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
...
-                    no_diacritics_query_string[letter_index]
-                ]
+            # prefix and suffix search
+            ids_by_prefix = Wordform.affix_searcher.search_by_prefix(user_query)
+            ids_by_suffix = Wordform.affix_searcher.search_by_suffix(user_query)
 
-            for possible_wf in Wordform.objects.filter(
-                id__in=possible_wf_ids, is_lemma=True
+            for wf in Wordform.objects.filter(
+                id__in=set(chain(ids_by_prefix, ids_by_suffix))
```
eddieantonio: yo dawg, I heard you like `chain()`

PR #378, opened 2020-04-01
Experiment: Create a Docker image for reproducible deployments

eddieantonio:
I'm trying to build everything in a Docker container so that the production server can just pull this image, and then run it super quickly. This drastically reduces the amount of dependencies on the production server, since many of the dependencies—including node!—are install/build dependencies.

This is a rough draft of my experiments. Please comment.


Makefile
```diff
@@ -0,0 +1,11 @@
+dockerimage: docker/requirements.txt
+	cd docker/ && docker build .
+
+# Exclude the editable install of this package, because:
+#  - the editable install is only useful for running tests,
+#  - we want to install Python dependencies BEFORE copying over the code
+#  (our code changes more often than dependencies)
+docker/requirements.txt:
+	pipenv lock -r | grep -v -- '-e' > $@
```
Madoshakalaka, line 9: This code also works, with the help of [requirementslib](https://github.com/sarugaku/requirementslib)
just saying
```python
from requirementslib import Lockfile, Requirement
from pathlib import Path

def dump_non_editable_deps_as_requirements_txt() -> bool:
    # argument to Lockfile.create should be the dir that hosts Pipfile*
    requirements = [Requirement.from_pipfile(package_name, config) for package_name, config in Lockfile.create('.').get_deps().items()]
    with open('docker/requirements.txt', 'w') as f:
        f.writelines([r.as_line(include_hashes=False) + '\n' for r in requirements if not r.editable])
```

eddieantonio: I'm not sure I understand why I'd use this approach rather than a Makefile?

eddieantonio:
This is still a work in progress. There are a few issues:

 - [ ] the production server makes some hard assumptions that it will use mod-wsgi on Sapir; hopefully, we can remove this dependency
 - [ ] there are other hardcoded things like using `/cree-dictionary/` in the URLs. Again, Sapir only.

---

A few annoying things:

 - [ ] marisa-trie does not have a pre-built binary for Linux 64 bit. This means that the Docker build requires g++. I'm trying to keep the container small, so I'm install g++ only for installing Python deps, then removing it as soon as deps are built.
 - [ ] pip takes FOREVER to clone the git repository for the fork of `sortedcontainers`. What's the status on merging in the type hints for that? Did the maintainer ever get back to you, @Madoshakalaka?

Makefile
Madoshakalaka: Not saying this is superior. But I love the aspect of encapsulating editable packages in an `r.editable` attribute rather than depending on string parsing `grep -v -- '-e'`. Actually your script is perfectly fine!

Madoshakalaka:
@eddieantonio I love the tarball suggestion.

Current status: I added type hints as `.pyi` stem files for `sortedset` `sortedlist`  but NOT `sorteddict` partly cuz python already has OrderDictionary which demotivates me a bit. I thought stem files would be more acceptible to the author. But the author replied in my PR that he wants inline type hints on function definition :expressionless: I can spent 40 minutes doing that when I'm free but somehow I was never free in the past few months

Makefile
eddieantonio: Good point. String parsing is fragile, and can easily backfire! 😅

Madoshakalaka:
closed as I'm making a new pr

PR #383, opened 2020-04-05
FST update prompts

Madoshakalaka:
In a perfect world, we would have full on automation, but this is not a perfect world

https://www.youtube.com/watch?v=Kl3H4vMqYNo


Madoshakalaka:
I mean, we can set my password to Sapir as one of the GitHub secrets, and just use SFTP to acess Sapir.

PR #386, opened 2020-04-14
Add "Results not found" message

### LINKED ISSUE #294: Inconsistent pages when no results are found
Visiting `/search/poopoo` gives you the page with the welcoming message because no results are found.

While typing `poopoo` gives you a blank page.

It should be one way or another. Preferably, I think we need to inform the user with messages like "No search results found"

eddieantonio:
Shows a dedicated message when there are no results for a given query, like this:

Searching for 'notaword' with no results

Fixes #294.


Madoshakalaka:
Looks good!!

kobexamoh:
Looks good to me! Sorry for the initial mistake – I think I commented instead of reviewing explicitly 😅

Issue #415, opened 2020-05-31
Help page should open its own tab/window


aarppe:
When you click on Help, that should open its own tab/window and retain the search page, so that one can go easily back and forth.

eddieantonio:
Please use cmd+click instead! See the email I privately sent you.

PR #416, opened 2020-05-31
Update base.html to open new tab/window for Help

aarppe:
Following up on #415, it would be desirable for clicking on 'Help' to open a new tab/window.


aarppe:
Yes, we can use cmd+click, but perhaps not all users know, or remember this (that was the case for myself). Not that this is somehow a must feature for me.

Madoshakalaka:
LGTM, btw I always use middle click (with the scroll wheel) to open a page in a new tab.

eddieantonio:
@aarppe, so our workflow is, once the PR gets an approval from at least one review, then you hit the big "Merge pull request" button. You can do that now!

PR #418, opened 2020-06-02
Fix pipon definition merge

### LINKED ISSUE #412: Dictionary entries for lemmas with both noun and verb reading incorrectly conflated together
Following up on #190, so that inflected forms are now matched, the dictionary entries for the noun and verb reading are now conflated, which shouldn't take place, cf.

image

Unless this is so in the source, _winter_ (ni) and _it is winter_ (vii) should be two different dictionary entries also in itwêwina.

Madoshakalaka:
This PR fixes #412, where both the verb and the noun _pipon_ were wrongly associated with same 3 definitions. Some code didn't deal with things to enough resolution.

Now it's correct:

Screenshot from 2020-06-01 20-48-40

Potentially this also fixes some other cases.

I'll update production database after this is merged to master.

### DIFF
--- a/CreeDictionary/DatabaseManager/xml_importer.py
+++ b/CreeDictionary/DatabaseManager/xml_importer.py
@@ -513,7 +513,6 @@ def import_xmls(dir_name: Path, multi_processing: int = 1, verbose=True):
                         keyword_counter += 1
 
                 # now we create definition for all (possibly inflected) entries in xml that are forms of this lemma.
-                # d_s_dicts = xml_lemma_to_definition_src_dicts.get(generated_wordform)
 
                 # try to match our generated wordform to entries in the xml file,
                 # in order to get its definition from the entries
@@ -522,7 +521,7 @@ def import_xmls(dir_name: Path, multi_processing: int = 1, verbose=True):
                 # first get homographic entries from the xml file
                 pos_lc_tuples_in_xml = xml_lemma_to_pos_lc.get(generated_wordform)
 
-                # The case when we do have holographic entries in xml,
+                # The case when we do have homographic entries in xml,
                 # Then we check whether these entries' pos and lc agrees with our generated wordform
                 if pos_lc_tuples_in_xml is not None:
                     for xml_pos, xml_lc in pos_lc_tuples_in_xml:
@@ -534,8 +533,6 @@ def import_xmls(dir_name: Path, multi_processing: int = 1, verbose=True):
                             )
                 # The case when we don't have holographic entries in xml,
                 # The generated inflection doesn't have a definition
-                else:
-                    pass
 
                 for d_s_dict in d_s_dicts:
 
--- a/CreeDictionary/DatabaseManager/xml_importer.py
+++ b/CreeDictionary/DatabaseManager/xml_importer.py
@@ -13,6 +13,7 @@ from API.models import Definition, DictionarySource, EnglishKeyword, Wordform
 from DatabaseManager import xml_entry_lemma_finder
 from DatabaseManager.cree_inflection_generator import expand_inflections
 from DatabaseManager.log import DatabaseManagerLogger
+from DatabaseManager.xml_consistency_checker import does_lc_match_xml_entry
 from constants import POS
 from utils import fst_analysis_parser
 from utils.crkeng_xml_utils import extract_l_str, parse_xml_lc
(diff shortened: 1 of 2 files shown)


Madoshakalaka:
@eddieantonio pretty technical changes. I do have comments if you want to scrutinize.

CreeDictionary/DatabaseManager/xml_importer.py
```diff
@@ -462,22 +463,14 @@ def import_xmls(dir_name: Path, multi_processing: int = 1, verbose=True):
...
-        xml_lemmas = {t[0] for t in xml_lemma_pos_lcs}
-
-        xml_lemma_to_definition_src_dicts: Dict[str, List[Dict[str, Set[str]]]] = {
-            xml_lemma: [
-                xml_lemma_pos_lc_to_str_definitions[(xml_lemma,) + pos_lc_tuple]
-                for pos_lc_tuple in xml_lemma_to_pos_lc[xml_lemma]
-            ]
-            for xml_lemma in xml_lemmas
-        }
-
```
Madoshakalaka, line 475: A major task below is to find definitions from entries in the xml and associate them with machine generated wordforms.

`xml_lemma_to_definition_src_dicts` here is built and used in the process. The old way negalects pos/lc in the association process and uses the same definitions for homographs (the pipon case for e.g.).

Also this is mad comprehension to an extend of unneccessary. Thanks god it's finally deleted.

CreeDictionary/DatabaseManager/xml_importer.py
```diff
@@ -462,22 +463,14 @@ def import_xmls(dir_name: Path, multi_processing: int = 1, verbose=True):
...
         # build wordforms and definition in db
         for generated_analysis, generated_wordforms in expanded[true_lemma_analysis]:
 
-            generated_lemma = fst_analysis_parser.extract_lemma(generated_analysis)
-            assert generated_lemma is not None
+            generated_lemma_lc = fst_analysis_parser.extract_lemma_and_category(
+                generated_analysis
+            )
+            assert generated_lemma_lc is not None
+            generated_lemma, generated_lc = generated_lemma_lc
```
Madoshakalaka, line 473: just to have an extra `generated_lc` to use below

CreeDictionary/DatabaseManager/xml_importer.py
```diff
@@ -520,22 +513,43 @@ def import_xmls(dir_name: Path, multi_processing: int = 1, verbose=True):
...
+                        if does_lc_match_xml_entry(generated_lc, xml_pos, xml_lc):
+                            d_s_dicts.append(
+                                xml_lemma_pos_lc_to_str_definitions[
+                                    (generated_wordform, xml_pos, xml_lc)
+                                ]
+                            )
+                # The case when we don't have holographic entries in xml,
+                # The generated inflection doesn't have a definition
+                else:
+                    pass
```
Madoshakalaka, line 538: The new matching.

CreeDictionary/tests/DatabaseManager_tests/test_xml_importer_regressions.py
```diff
@@ -23,17 +23,42 @@ def test_import_xml_common_analysis_definition_merge(shared_datadir):
...
+        .count()
+        == 1
+    )
+    assert (
+        Wordform.objects.filter(text="pipon", is_lemma=True, pos="V")
+        .get()
+        .definitions.all()
+        .count()
+        == 2
+    )
```
Madoshakalaka, line 64: unit/regression test

CreeDictionary/DatabaseManager/xml_importer.py
```diff
@@ -520,22 +513,43 @@ def import_xmls(dir_name: Path, multi_processing: int = 1, verbose=True):
...
+                            text=str_definition,
+                            wordform=db_inflection,
+                        )
+                        assert definition_counter not in citations
+                        citations[definition_counter] = set(source_strings)
 
-                            definition_counter += 1
-                            db_definitions.append(db_definition)
+                        definition_counter += 1
+                        db_definitions.append(db_definition)
```
Madoshakalaka, line 552: black formatting

CreeDictionary/DatabaseManager/xml_importer.py
eddieantonio: Deleting code is always great!

CreeDictionary/DatabaseManager/xml_importer.py
```diff
@@ -520,22 +513,43 @@ def import_xmls(dir_name: Path, multi_processing: int = 1, verbose=True):
                         keyword_counter += 1
 
                 # now we create definition for all (possibly inflected) entries in xml that are forms of this lemma.
-                d_s_dicts = xml_lemma_to_definition_src_dicts.get(generated_wordform)
+                # d_s_dicts = xml_lemma_to_definition_src_dicts.get(generated_wordform)
```
eddieantonio, line 516:
```suggestion
```

CreeDictionary/DatabaseManager/xml_importer.py
```diff
@@ -520,22 +513,43 @@ def import_xmls(dir_name: Path, multi_processing: int = 1, verbose=True):
...
+
+                # try to match our generated wordform to entries in the xml file,
+                # in order to get its definition from the entries
+                d_s_dicts: List[dict] = []
+
+                # first get homographic entries from the xml file
+                pos_lc_tuples_in_xml = xml_lemma_to_pos_lc.get(generated_wordform)
+
+                # The case when we do have holographic entries in xml,
```
eddieantonio, line 525:
```suggestion
                # The case when we do have homographic entries in xml,
```

well holo everyone

CreeDictionary/DatabaseManager/xml_importer.py
eddieantonio, line 538:
```suggestion
```

CreeDictionary/DatabaseManager/xml_importer.py
Madoshakalaka: @eddieantonio LOOLOOLOLL I did notice that line of undeleted commented out code when I was reviewing. And I was thinking "fine I'll just leave it there Eddie won't notice"

aarppe:
Yes, it now works as it should for _kîsikâw_ as well, see:

image

This also seems to mostly be the case for also _kimiwan_,

image

... though there I think the major cause of the remaining mix-up is in the XML source.

N.B. This should **NOT** be fixed programmatically, but in the source (XML or otherwise), by linguists. The reason for this is probably that the comparison script I used didn't properly disambiguate word-class-wise ambiguous dictionary entries.

aarppe:
Generally, there is a small but non-insignificant number of dictionary entries which are ambiguous for the head word (e.g. II verbs and nouns, some II and TA verbs). Because of this, any combination of the dictionary entries may need to be sensitive as deep down as the inflectional subcategory (sometimes the general word class is sufficient, but not always). In the XML, the inflectional subcategory is provided in the `lc` variable. Something similar(ly named) will be used in whatever format the new, aggregated crk dictionary source will look like.

Madoshakalaka:
@aarppe get it

PR #419, opened 2020-06-02
homographic entry paradigm fix, fixes #395

### LINKED ISSUE #395: Ambiguous words do not get paradigms
Searching for _nôhtêpayiw_, I do get a match with two dictionary entries:

image

But I cannot get a paradigm for either. Neither can I get a full analysis, even though the computational model does provide that:

```
hfst-lookup -q src/analyser-gt-desc.hfst
nôhtêpayiw
nôhtêpayiw	nôhtêpayiw+V+AI+Ind+3Sg	0.000000
nôhtêpayiw	nôhtêpayiw+V+II+Ind+3Sg	0.000000
```
The same applies for _ayâw_:

image

I suspect this is connected with _ayâw_ being either an II, AI, or TI verb.

Madoshakalaka:
#395 describes the bug that clicking on entries of ayâw and nôhtêpayiw doesn't show the user their paradigms.

The cause is twofold, the ayaw case and the nôhtêpayiw case have different direct causes. But they can be classified both as improper handling of the "extra constraints" query param.

### DIFF
--- a/CreeDictionary/API/models.py
+++ b/CreeDictionary/API/models.py
@@ -37,6 +37,7 @@ from constants import POS, ConcatAnalysis, FSTTag, Label, Language, ParadigmSize
 from fuzzy_search import CreeFuzzySearcher
 from paradigm import Layout
 from shared import paradigm_filler
+from urllib.parse import quote
 from utils import fst_analysis_parser, get_modified_distance
 from utils.cree_lev_dist import remove_cree_diacritics
 from utils.fst_analysis_parser import (
@@ -186,9 +187,9 @@ class Wordform(models.Model):
             "cree-dictionary-index-with-lemma", kwargs={"lemma_text": self.text}
         )
         if self.homograph_disambiguator is not None:
-            lemma_url += f"?{self.homograph_disambiguator}={getattr(self, self.homograph_disambiguator)}"
+            lemma_url += f"?{self.homograph_disambiguator}={quote(getattr(self, self.homograph_disambiguator))}"
 
-        return iri_to_uri(lemma_url)
+        return lemma_url
 
     def serialize(self) -> SerializedWordform:
         """
@@ -332,7 +333,9 @@ class Wordform(models.Model):
         super(Wordform, self).save(*args, **kwargs)
 
     @staticmethod
-    def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
+    def fetch_lemma_by_user_query(
+        user_query: str, **extra_constraints
+    ) -> "CreeAndEnglish":
         """
         treat the user query as cree and:
 
@@ -346,7 +349,7 @@ class Wordform(models.Model):
         Give a list of matched lemmas
 
         :param user_query: can be English or Cree (syllabics or not)
-        :param kwargs: additional fields to disambiguate
+        :param extra_constraints: additional fields to disambiguate
         """
         # Whitespace won't affect results, but the FST can't deal with it:
         user_query = user_query.strip()
@@ -376,7 +379,7 @@ class Wordform(models.Model):
             ids_by_suffix = Wordform.affix_searcher.search_by_suffix(user_query)
 
             for wf in Wordform.objects.filter(
-                id__in=set(chain(ids_by_prefix, ids_by_suffix))
+                id__in=set(chain(ids_by_prefix, ids_by_suffix)), **extra_constraints
             ):
                 cree_results.add(CreeResult(wf.analysis, wf, wf.lemma))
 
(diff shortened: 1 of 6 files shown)


CreeDictionary/API/models.py
```diff
@@ -186,9 +187,9 @@ def get_absolute_url(self) -> str:
             "cree-dictionary-index-with-lemma", kwargs={"lemma_text": self.text}
         )
         if self.homograph_disambiguator is not None:
-            lemma_url += f"?{self.homograph_disambiguator}={getattr(self, self.homograph_disambiguator)}"
+            lemma_url += f"?{self.homograph_disambiguator}={quote(getattr(self, self.homograph_disambiguator))}"
 
-        return iri_to_uri(lemma_url)
+        return lemma_url
```
Madoshakalaka, line 192: query params were not escaped correctly.

Specifically, the plus sign in `?analysis=xx+xx+xx+xx` wasn't escaped, which affected the ayaw case

CreeDictionary/API/models.py
```diff
@@ -332,7 +333,9 @@ def save(self, *args, **kwargs):
         super(Wordform, self).save(*args, **kwargs)
 
     @staticmethod
-    def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
+    def fetch_lemma_by_user_query(
+        user_query: str, **extra_constraints
+    ) -> "CreeAndEnglish":
```
Madoshakalaka, line 338: renamed `kwargs` to `extra_constraints`. I think it's more descriptive, as the kwargs are exclusively used for that.

CreeDictionary/API/models.py
```diff
@@ -376,7 +379,7 @@ def fetch_lemma_by_user_query(user_query: str, **kwargs) -> "CreeAndEnglish":
             ids_by_suffix = Wordform.affix_searcher.search_by_suffix(user_query)
 
             for wf in Wordform.objects.filter(
-                id__in=set(chain(ids_by_prefix, ids_by_suffix))
+                id__in=set(chain(ids_by_prefix, ids_by_suffix)), **extra_constraints
```
Madoshakalaka, line 382: `**extra_constraints` was missed here --- it should be included inside every Queryset.filter call.

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-entries.html
```diff
@@ -10,7 +10,7 @@ <h2 class="definition-title definition-title--search-result">
 
         <dfn class="definition-title__word" data-cy="definition-title">
           {% if result.is_lemma %}
-          <a class="definition-title__link" href="{{ result.lemma_wordform.lemma_url }}">
+          <a class="definition-title__link" data-cy="lemma-link" href="{{ result.lemma_wordform.lemma_url }}">
```
Madoshakalaka, line 13: for better integration testing. `[data-cy="definition-title"]` definied above gives both (unclickable) matched wordforms and (clickable) lemmas.

while `[data-cy="lemma-link]"`should be 100% a link and directs to the paradigm page

CreeDictionary/CreeDictionary/views.py
```diff
@@ -23,7 +23,9 @@ def lemma_details(request, lemma_text: str = None):  # pragma: no cover
     :param lemma_text: the exact form of the lemma (no spell relaxation)
     """
     extra_constraints = {
-        k: v for k, v in request.GET.items() if k in {"pos", "lc", "analysis", "id"}
+        k: v
+        for k, v in request.GET.items()
+        if k in {"pos", "full_lc", "analysis", "id"}
```
Madoshakalaka, line 28: This was responsible for the "nôhtêpayiw" case. Wordform objects generates constraints with `full_lc` rather than `lc`.

So to here's a simple explanation the bug: 

1. On the search page, one of the two _nôhtêpayiw_ lemma has `href=/word/nôhtêpayiw?full_lc=VII-3`
2. When the user clicks on the link, the `full_lc` is dropped here by this faulty code.
3. Without the `full_lc` constraint, the app finds out there are two potential matches, thus redirects the user to the same search page with the two results.To the user, it's as if the server isn't responding and that the entries do not have paradigms.

cypress/integration/regressions.spec.js
```diff
@@ -226,6 +226,66 @@ context('Regressions', () => {
...
+        cy.get('[data-cy=lemma-link]').its('length').then(length=>{
+
+          for (let i = 0; i< length; i++){
+            cy.visitSearch('nôhtêpayiw')
+            cy.get('[data-cy=lemma-link]').eq(i).click()
+            cy.get('[data-cy=paradigm]').should('be.visible')
+          }
+
+
+        })
```
Madoshakalaka, line 272: regression/integration test

CreeDictionary/CreeDictionary/views.py
eddieantonio: Good detective work!

cypress/integration/regressions.spec.js
eddieantonio: Lol, why is there so much whitespace?

Madoshakalaka: oneday eslint will correct it all

PR #421, opened 2020-06-03
JavaScript cleanup

eddieantonio:
This is a series of refactors, style tweaks, and changes, including:

 - [x] phase out use of jQuery: not 100% complete
 - [x] apply ESLint on all of our regular files
 - [x] use `fetch()` consistently instead of `XMLHTTPRequest`

I mostly worked on `index.js`, not really doing anything more than running ESLint on `recordings.js`.

**EDIT**: one casualty is waiting for the `fetch()` in our loading bar tests because Cypress doesn't _currently_ support `fetch()` BUT, [they're working on it!](https://github.com/cypress-io/cypress/issues/95#issuecomment-637567912)


eddieantonio:
> @eddieantonio why shouldn't we use jquery

It's one more dependency that we don't really need when we're doing DOM manipulation that can be done without jQuery. Do the same with a lot less code!

src/index.js
```diff
@@ -10,87 +10,102 @@ import './css/styles.css'
...
-  })
+  let tooltips = searchResultList.querySelectorAll('.definition-title__tooltip-icon, .preverb-breakdown__tooltip-icon')
+  for (let icon of tooltips) {
+    createTooltip($(icon), $(icon).next('.tooltip'))
+  }
 }
 
 /**
- * use xhttp to load search results in place
+ * use AJAX to load search results in place
```
Madoshakalaka, line 102: @eddieantonio why is ajax favoured over http xhttp?

eddieantonio: AJAX refers to any _asynchronous_ HTTP request, whether it's done with the crusty old `XMLHttpRequest` API or the newer, less error-prone `fetch()` API.

Madoshakalaka: aah was asking why use fetch over XMLHttpRequest

kobexamoh: > aah was asking why use fetch over XMLHttpRequest

I've personally always went with XMLHttpRequest because it's what I was taught/learned first! Fetch is new to me but I'm aware that it is the future so I'll try to work at using it more go forward.

If I remember right, fetch handles some things better than XMLHttpRequest does (did?). I think I have a short tutorial tucked away for why the newer is better: just holler and I can pull it over the weekend or something 💪

PR #423, opened 2020-06-03
Recordings on the search page (inefficient implementation)

### LINKED ISSUE #404: Add recording link to the search page
You should be able to hear at least ONE recording for each word that has recordings STRAIGHT FROM the search page!

So the speaker icon should be beside the (i) icon.

**EDIT**: this will avoid the question "_which words have audio?_"

eddieantonio:
This adds recordings to the search page. No more clicking through to the paradigm page to hear a recording!

Screen Shot 2020-06-03 at 4 33 51 PM

Note: For _n_ search results, it hammers the recording-validation app with _n_ queries. The better way to do this is if the recording-validation app had a bulk API, but alas, it does not (yet).

Fixes #404.

### DIFF
--- a/CreeDictionary/CreeDictionary/templates/CreeDictionary/base.html
+++ b/CreeDictionary/CreeDictionary/templates/CreeDictionary/base.html
@@ -18,6 +18,15 @@
   {% include './svg-sprites.html' %}
   {% include './components/header.html' %}
 
+  {# The 🔊 button template. Gets inflated by JavaScript. #}
+  <template id="template:play-button">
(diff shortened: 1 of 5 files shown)


kobexamoh:
Oooh! Excited to test this!

aarppe:
This is excellent! - especially as a feature from the end-user perspective, not knowing the hairiness of the implementation ...

Obviously, next steps would be to add the recording icon next to more dictionary entries - where Erin, Alyssa, Daniel and myself have our work cut out for us, in reviewing and finishing the annotations.

PR #425, opened 2020-06-04
Cleanup tooltips

eddieantonio:
Some sprucing up of the tooltips:

 - [x] phase out jQuery in tooltip.js — we'll soon be able to remove 87KiB from our JavaScript bundle!
 - [x] change styles to more closely match Diane's designs

Screen Shot 2020-06-04 at 9 15 07 AM
Screen Shot 2020-06-04 at 9 15 28 AM


kobexamoh:
Peeking now!

PR #428, opened 2020-06-04
New presentation for inflected form search results

eddieantonio:
Fixes #399 

 - [x] shift word class down to "elaboration"
 - [x] add Cree explanation for word class
 - [x] ensure the word class is still visible for lemmas

This is what it looks like...

~~for the lemma form of verbs~~: 
Screen Shot 2020-06-05 at 1 36 16 PM

**EDIT**: this is what it looks like for verbs now (consistent with non-lemma forms):
Screen Shot 2020-06-05 at 3 13 09 PM

for inflected forms of verbs:
Screen Shot 2020-06-05 at 1 36 28 PM

for nouns:
Screen Shot 2020-06-05 at 1 37 40 PM

for particles/preverbs:
Screen Shot 2020-06-05 at 1 35 59 PM

PR #430, opened 2020-06-05
Word entry cleanup

eddieantonio:
Just making some more systematic class names. This is the schematic of nesting of the classes now:

```
.
└── .definition
    └── .definition__header
        ├── .definition__title
        │   ├── .definition__matched-head
        │   └── .definition_icons
        │       └── .definition_icon
        └── .definition__elaboration
            ├── .definition__reference-to-lemma
            │   └── .definition__matched-lemma
            └── .wordclass
                ├── .wordclass__general
                └── .wordclass__help
```

(I realize that `.wordclass` can be used outside of the definition)

~~Blocked by #428~~

### DIFF
--- a/CreeDictionary/CreeDictionary/templates/CreeDictionary/base.html
+++ b/CreeDictionary/CreeDictionary/templates/CreeDictionary/base.html
@@ -20,7 +20,7 @@
 
   {# The 🔊 button template. Gets inflated by JavaScript. #}
   <template id="template:play-button">
-    <button aria-label="Play recording" class="definition-title__play-button" data-cy="play-recording">
+    <button aria-label="Play recording" class="definition__icon definition-title__play-button" data-cy="play-recording">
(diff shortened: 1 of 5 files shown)


kobexamoh:
Getting into this just now but will be 🔙 to it later in the evening!

kobexamoh:
The schematic nesting of the classes – this isn't for the frontend, correct? It's more for the way we order items behind the scenes?

(I ask because it seems like part of it maps to the way things are structured in the browser, but part of it doesn't 🤔 )

eddieantonio:
> The schematic nesting of the classes – this isn't for the frontend, correct? It's more for the way we order items behind the scenes?
> 
> (I ask because it seems like part of it maps to the way things are structured in the browser, but part of it doesn't 🤔 )

Oh no! Part of it doesn't map to the browser? I'll have to investigate!

PR #435, opened 2020-06-08
Fix #213 missing entries

### LINKED ISSUE #213: Cree word forms not getting some or all analyses
Pursuant #100 and #209, now the following Cree word forms are not recognized, either at all, or only receiving one of multiple possible analyses (`-`: analysis apparently missing; `+`: analysis apparently correct):

- [ ] `maskos (-maskwa+N+A+Der/Dim+N+A+Sg)`
- [ ] `maskwak (-maskwa+N+A+Pl)`
- [ ] `nimaskom (-maskwa+N+A+Px1Sg+Sg)`
- [ ] `niskak (-niska+N+A+Pl; +nîskâw+V+II+Cnj+Prs+3Sg)`
- [x] `nitas (-mitâs+N+A+D+Px1Sg+Sg; -mitâs+N+I+D+Px1Sg+Sg; -nitâs+N+A+D+Px1Sg+Sg; -nitâs+N+I+D+Px1Sg+Sg)` (two results with _nitâs_ as lemma, but `Sg` feature not showing)
- [x] `mitas (-mitâs+N+A+D+PxX+Sg; -mitâs+N+I+D+PxX+Sg)` (two results with _mitâs_ as lemma, but `Sg` feature not showing)
- [x] `mitasak (-mitâs+N+A+D+PxX+Pl)`

Dependent nouns in their lemma form are recognized, but other inflected forms such as plurals and obviatives are not.

Madoshakalaka:
#213 is majorly caused by the importer's inability to identify some entries lemma:

```
niska
niska   niska+N+A+Sg
niska   niska+N+A+Obv

maskwa 
maskwa  maskwa+N+A+Sg
maskwa  maskwa+N+A+Obv
```

We need a way to specifily which one is the lemma, which one isn't.

I introduced `CreeDictionary/res/linguistic_knowledge/` folder for that.

To embrace future support of multiple languages (gunaha), the directory currently has two folders: `crk` and `gunaha` (naming help needed)

`linguistic_knowledge/crk/lemma-disambiguation.txt` is created with the following content:

```
# disambiguation for maskwa
+ maskwa+N+A+Sg
- maskwa+N+A+Obv

# disambiguation for niska
+ niska+N+A+Sg
- niska+N+A+Obv
```

A preceding plus sign means the analysis is for a lemma
A preceding minus sign meanings the analysis is not for a lemma


CreeDictionary/DatabaseManager/xml_entry_lemma_finder.py
```diff
@@ -3,19 +3,57 @@
...
 from DatabaseManager.xml_consistency_checker import does_lc_match_xml_entry
 from constants import ConcatAnalysis, FSTLemma, SimpleLC
 from shared import strict_analyzer
+from utils import shared_res_dir
 
 init()  # for windows compatibility
 
 
+class SpecialLemmaDisambiguator:
+    def __init__(self, language: Literal["crk", "gunaha"]):
```
Madoshakalaka, line 22: @eddieantonio naming suggestion needed

cypress/integration/recordings.spec.js
```diff
@@ -3,7 +3,7 @@
  */
 context('Recordings', function () {
   describe('On the search page', () => {
-    it('should display for words  on the search page', () => {
+    it('should display for words', () => {
```
Madoshakalaka, line 6: just a nitpick. "on the search page" was already in `describe`

Madoshakalaka:
@aarppe You might be interested in the change

Madoshakalaka:
So this is probablly not a good idea...I realized this is a common phonemon for many NA words.

i.e.  Singular form and Obviative form looks the same for many NA words.

Will do a more general solution. I'd like to use our old firend: [lemma-tags.tsv](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/79c0aff0a00cfef43ba1dc6e3bcd740ca5341ec3/CreeDictionary/res/lemma-tags.tsv) instead when this happens.

aarppe:
This ambiguity applies for so-called single-syllable NA words, which differ from the other NA words in that they take a singular suffix _-a_, which is the same as the obviative suffix _-a_.

When importing dictionary content, we can presume that a noun has the singular feature (rather than obviative), if there is ambiguity. But as far as I can judge, the only nouns of this type are: NA-4.

aarppe:
@Madoshakalaka Yes, this [lemma-tags.tsv](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/79c0aff0a00cfef43ba1dc6e3bcd740ca5341ec3/CreeDictionary/res/lemma-tags.tsv) seems like the way to go, because we can then language-specifically decide with one of ambiguous analyses is appropriate as the lemma.

There are some exceptions, in that there are e.g. plural-only verbs (e.g. reciprocals), but normally those would not be structurally ambiguous with the singular forms, see e.g. https://sapir.artsrn.ualberta.ca/cree-dictionary/search?q=they

PR #436, opened 2020-06-09
C ons tants

Madoshakalaka:
Just a nice lil refactoring, it wasn't so sure what "constants" is.

`constants/__init__.py` is divided to four separate files and they are put inside `utils` folder now.

no-more-constants


CreeDictionary/utils/data_classes.py
```diff
@@ -0,0 +1,27 @@
...
+            result += self.raw_prefixes + "+"
+        result += f"{self.lemma}+{self.raw_suffixes}"
+        return ConcatAnalysis(result)
+
+
+ORTHOGRAPHY_NAME = {
+    "Latn": "SRO (êîôâ)",
+    "Latn-x-macron": "SRO (ēīōā)",
+    "Cans": "Syllabics",
+}
```
eddieantonio, line 27: Good, so far, but why did this get moved here? Wouldn't `utils.vars` be better?

Madoshakalaka: This reminds me of TypedDicts/NamedTuple...but yeah, I'll move it to vars

eddieantonio:
LGTM 👍

PR #439, opened 2020-06-09
Fix #213 missing entries | Default Lemma Analyses for Word Classes 

### LINKED ISSUE #213: Cree word forms not getting some or all analyses

Madoshakalaka:
#213 is related to our code's ability to identify xml entries with their fst analyses.

We used to rely purely on `lemma-tags.tsv` to identify xml entries with their fst analyses. It was flawed as there are idiomatic counterexamples. And for pronouns, there is no easy rule.

The newer code relied on comparing `<l> <pos> lc` in xml to fst analyzer outputs. There are cases this approach fails. e.g. the code has no way to know which one of `niska+N+A+Sg` or `niska+N+A+Obv` is the lemma analysis for _niska_ as both analyses look the same and has the same word class.

This PR added `lemma-tags.tsv` back, but as a fallback this time. As it's useful in disambiguating these cases.

I'll upload built production database when this pr is approved. (There is no `models.py` change so just a database swap and `$ touch wsgi.py` will work.)


cypress/integration/recordings.spec.js
```diff
@@ -3,7 +3,7 @@
  */
 context('Recordings', function () {
   describe('On the search page', () => {
-    it('should display for words  on the search page', () => {
```
Madoshakalaka, line 6: nitpick, "on the search page" was already inside `describe`

CreeDictionary/DatabaseManager/xml_entry_lemma_finder.py
```diff
@@ -3,20 +3,65 @@
...
 
 
+class DefaultLemmaPicker:
+    """
+    When two analyses have the same looking lemma and word class, this helps choose the preferred lemma
+
+    i.e. It helps solve this question: "Which one of maskwa+N+A+Sg and maskwa+N+A+Obv is the lemma?"
+    """
+
+    def __init__(self, language: Literal["crk", "gunaha"]):
```
eddieantonio, line 29:
```suggestion
    def __init__(self, language: Literal["crk", "srs"]):
```

If we're using the ISO 639 code for Tsuut'ina.

| Language | ISO 639-3 | App Name |
| - | - | - |
| Plains Cree | crk | itwêwina |
| Tsuut'ina | srs | Gúnahà |

CreeDictionary/tests/DatabaseManager_tests/test_xml_importer_regressions.py
```diff
@@ -62,3 +62,23 @@ def test_import_pipon_of_different_word_classes(shared_datadir):
...
+
+
+@pytest.mark.django_db
+def test_import_maskwa(shared_datadir):
+
+    # https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/213
+    # searching with maskwak didn't yield an inflection of "maskwa" as "maskwa" wasn't disambiguated and was marked as_is
+    migrate_and_import(shared_datadir / "crkeng-maskwa")
+
+    assert not Wordform.objects.filter(text="maskwa", is_lemma=True).get().as_is
```
eddieantonio, line 84: 👍

cypress/integration/recordings.spec.js
eddieantonio: Haha, good point!

CreeDictionary/DatabaseManager/xml_entry_lemma_finder.py
Madoshakalaka: yyeaeeah thanks

PR #449, opened 2020-06-11
Refactor altlabel

eddieantonio:
Converted `FST_TAG_LABELS` to a `Relabelling` class. I changed the API, so instead of doing this:

```python
FST_TAG_LABELS.get(LabelFriendliness.ENGLISH, {}).get(tag, None)
```

You can do this instead:

```python
FST_TAG_LABELS.english[tag]
```

or

```python
FST_TAG_LABELS.english.get(tag, None)
```


CreeDictionary/utils/fst_analysis_parser.py
```diff
@@ -14,16 +14,53 @@
 )
 
 
-class LabelFriendliness(Enum):
-    LINGUISTIC_SHORT = 0
-    LINGUISTIC_LONG = 1
-    ENGLISH = 2
-    NEHIYAWEWIN = 3
+class LabelFriendliness(IntEnum):
```
Madoshakalaka, line 17: What's the motivation of using IntEnum here. I think `auto` works inside just Enum class

eddieantonio: Honestly, I didn't have a huge motivation for doing this. I think it's just because I saw that the enumeration elements are ints, and I'm like "well, obviously this is an `int` enum". I'm not sure this detail really matters

PR #451, opened 2020-06-11
Create (faster?) lemmas() strategy

eddieantonio:
The old `random_lemmas()` strategy would frequently cause Hypothesis health check warnings. I believe this is because it did a full database scan for EVERY example that Hypothesis would request.

This introduces a strategy that does only ONE full database scan, and this scan is executed lazily — it only happens when the first example is requested by Hypothesis.


CreeDictionary/tests/conftest.py
```diff
@@ -22,36 +23,37 @@ def topmost_datadir():
     return Path(dirname(__file__)) / "data"
 
 
-@composite
-def analyzable_inflections(draw) -> Wordform:
+class WordformStrategy(SearchStrategy[Wordform]):
```
Madoshakalaka, line 26: wow hypothesis has some nice generics typing

PR #462, opened 2020-06-17
Revamp search results header

### LINKED ISSUE #358: Proposal: use emoji on search results page instead of general/specific word class
During our design QA with Diane and @kobexamoh, we considered the option of dropping the general word class text (e.g., Noun, Verb) and replacing it with the simplified emoji. I decided to implement it just for fun as can be seen in this video: https://youtu.be/55_oFnzqQas

Screen Shot 2020-03-23 at 3 46 52 PM

Replacing the general word class with emoji helps in a few ways:

 * there was too much text on the screen — too much horizontal space; this saves a bit of horizontal space
 * "Noun" and "Verb" don't tell you enough information about a word in Cree
 * It's easy to scan for the form you want when looking up an English word. This is my personal opinion, but I would much rather stare at this emoji than try to work out if I want a VTI or a VAI. I've never internalized these terms :/

The downsides are that it is a "cute" way of display word class; some people may rather see "Verb" or even "VTA". The other is that I think this is a new thing for Plains Cree, but it is not unprecedented in Algonquian languages: I got the idea from an Anishinabemowin educator: https://twitter.com/Niigaatikwe/status/1113128787334610945

@aarppe, please share your thoughts. I think this is a whole lot more usable.

**EDIT**: the experimental implementation is in this branch: https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/tree/experiment/wordclass-emoji. It is NOT production-ready!

### LINKED ISSUE #434: Display help text for inflectional category instead of word class
Instead of displaying just the simple re-labelling of the word class as introduced in #428, display the help text for the inflectional category. This data should be in `crk.altlabel`!

---

Two questions, which can be addressed now, or later:
1. For the inflectional category, are we using Arok's inflectional classes that should be in our internal dictionary database, i.e. VII/VAI/VTI/VTA-[1-5], and their relabelings in `crk.altlabel.tsv`? (Rather than the specific wordclass (that comes from the linguistic analysis, and which also has relabelings in `crk.altlabel.tsv`)
2. Sh/Couldn't we provide the inflectional category for nouns as well? Their inflectional category codes should be available, and the relabelings for them should exist in `crk.altlabel.tsv`.

Also note that the relabeling `like: wîcihêw, itêw` is rather the plain English label for the specific word class, than the Cree one - this Cree labeling would rather be: `tâpiskôc: wîcihêw, itêw`. But I wouldn't use these specific wordclasses, but rather the inflectional categories, since they tell more about how the word inflectes, and the relabelings are actually shorter - see for VTAs:

```
VTA-1	VTA-1	transitive animate verb – class 1: regular	like: wîcihêw	tâpiskôc: wîcihêw
VTA-2	VTA-2	transitive animate verb – class 2	like: nitonawêw	tâpiskôc: nitonawêw
VTA-3	VTA-3	transitive animate verb – class 3	like: kîskiswêw	tâpiskôc: kîskiswêw
VTA-4	VTA-4	transitive animate verb – class 4	like: nakatêw	tâpiskôc: nakatêw
VTA-5	VTA-5	transitive animate verb – class 5	like: itêw	tâpiskôc: itêw
```
As far as I understand, this information is already available, since it is used to disambiguate paradigm generation.

_Originally posted by @aarppe in https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/pull/428#issuecomment-639972427_

### LINKED ISSUE #445: Updated mockups for presenting search results
Since the descriptions in several previous issues are out-of-date or do not take into account later developments, this is an attempt at an up-to-date aggregation of all relevant previous ideas, and thus to replace #296, #224, #232 and even #101:

**1. Inflected form with definition – non-emoji**

image

**2. Inflected form without definition – non-emoji**

image

**3. Lemma definition – non-emoji**

image

**4. Inflected form with definition – emoji**

image

**5. Inflected form without definition – emoji**

image

**6. Lemma definition – emoji**

image

### LINKED ISSUE #447: Help button with pop-up info needed for inflectional category
As illustrated in #445, we need a new help button to give access to spelled out additional information concerning the word-class and inflectional category (this is the red `(i)` button in the mock-ups in #445).

So, an underlying code from the dictionary source, e.g. `VAI-1,` which is now to be spelled out at `Verb - like: nipâw` or `🧑🏽➡️ - like nipâw` would be associated with an (i)-button, which would then produce a pop-up with:  `VAI-1: vowel-final animate intransitive verb` OR `vowel-final animate intransitive verb (VAI-1)`

The information that would be provided is in effect relabeling that can be extracted from `crk.altlabel.tsv`.

This would look something like the following:

image

Clicking the (i) button would provide the following information:

image

eddieantonio:
This tackles the following:

 - [x] emoji for specific wordclass (closes #358)
 - [x] replaces English wordclass with English inflectional category (closes #434)
 - [x] popup for inflectional category (closes #447 )
 - [x] Updates layout to match latest mockups (closes #445)

---

**EDIT**: it's ready! Here are some screenshots:

non-lemma form with definition:
Screen Shot 2020-06-18 at 11 22 42 AM

its linguistic breakdown pop-up:
Screen Shot 2020-06-18 at 11 26 13 AM

its inflectional category pop-up:
Screen Shot 2020-06-18 at 11 26 44 AM

non-lemma:
Screen Shot 2020-06-18 at 11 23 05 AM

lemma:
Screen Shot 2020-06-18 at 11 23 19 AM

### DIFF
--- a/CreeDictionary/API/models.py
+++ b/CreeDictionary/API/models.py
@@ -223,14 +223,15 @@ class Wordform(models.Model):
 
     def get_emoji_for_cree_wordclass(self) -> Optional[str]:
         """
-        Attempts to get a description, e.g., "like: wîcihêw" or "like:
-        micisow" for this wordform.
+        Attempts to get an emoji description of the full wordclass.
+        e.g., "👤👵🏽" for "nôhkom"
         """
         maybe_full_word_class = self.full_word_class
         if maybe_full_word_class is None:
             return None
-        word_class = FSTTag(maybe_full_word_class.without_pos())
-        return LABELS.emoji.get(word_class)
+        fst_tag_str = maybe_full_word_class.to_fst_output_style().strip("+")
+        tags = [FSTTag(t) for t in fst_tag_str.split("+")]
+        return LABELS.emoji.get_longest(tags)
 
     @cached_property
     def homograph_disambiguator(self) -> Optional[str]:
--- a/CreeDictionary/API/models.py
+++ b/CreeDictionary/API/models.py
@@ -21,8 +21,6 @@ from urllib.parse import quote
 
 import attr
 from attr import attrs
-
-import CreeDictionary.hfstol as temp_hfstol
 from cree_sro_syllabics import syllabics2sro
 from django.conf import settings
 from django.db import models, transaction
@@ -30,10 +28,12 @@ from django.db.models import Max, Q, QuerySet
 from django.forms import model_to_dict
 from django.urls import reverse
 from django.utils.functional import cached_property
+from sortedcontainers import SortedSet
+
+import CreeDictionary.hfstol as temp_hfstol
 from fuzzy_search import CreeFuzzySearcher
 from paradigm import Layout
 from shared import paradigm_filler
-from sortedcontainers import SortedSet
 from utils import (
     ConcatAnalysis,
     FSTTag,
@@ -199,6 +199,15 @@ class Wordform(models.Model):
         result["lemma_url"] = self.get_absolute_url()
         result["wordclass_help"] = self.get_user_friendly_wordclass_help_for_cree()
 
+        # Displayed in the word class/inflection help:
+        result["inflectional_category_plain_english"] = LABELS.english.get(
+            self.inflectional_category
+        )
+        result["inflectional_category_linguistic"] = LABELS.linguistic_long.get(
+            self.inflectional_category
+        )
+        result["wordclass_emoji"] = self.get_emoji_for_cree_wordclass()
+
         return result
 
     def get_user_friendly_wordclass_help_for_cree(self) -> Optional[str]:
@@ -211,6 +220,17 @@ class Wordform(models.Model):
             return None
         word_class = FSTTag(maybe_full_word_class.without_pos())
         return LABELS.english.get(word_class)
+
+    def get_emoji_for_cree_wordclass(self) -> Optional[str]:
+        """
+        Attempts to get a description, e.g., "like: wîcihêw" or "like:
+        micisow" for this wordform.
+        """
+        maybe_full_word_class = self.full_word_class
+        if maybe_full_word_class is None:
+            return None
+        word_class = FSTTag(maybe_full_word_class.without_pos())
+        return LABELS.emoji.get(word_class)
 
     @cached_property
     def homograph_disambiguator(self) -> Optional[str]:
--- a/CreeDictionary/API/models.py
+++ b/CreeDictionary/API/models.py
@@ -197,7 +197,6 @@ class Wordform(models.Model):
             definition.serialize() for definition in self.definitions.all()
         ]
         result["lemma_url"] = self.get_absolute_url()
-        result["wordclass_help"] = self.get_user_friendly_wordclass_help_for_cree()
 
         # Displayed in the word class/inflection help:
         result["inflectional_category_plain_english"] = LABELS.english.get(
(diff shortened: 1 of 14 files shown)


eddieantonio:
@aarppe, note: once this is merged, I'd shorten the inflectional category information for N+*+D forms. They're a bit... noisy:

Screen Shot 2020-06-18 at 11 31 59 AM
Screen Shot 2020-06-18 at 11 32 21 AM

CreeDictionary/API/models.py
```diff
@@ -197,20 +197,29 @@ def serialize(self) -> SerializedWordform:
...
 
-    def get_user_friendly_wordclass_help_for_cree(self) -> Optional[str]:
+    def get_emoji_for_cree_wordclass(self) -> Optional[str]:
         """
-        Attempts to get a description, e.g., "like: wîcihêw" or "like:
-        micisow" for this wordform.
+        Attempts to get an emoji description of the full wordclass.
+        e.g., "👤👵🏽" for "nôhkom"
         """
         maybe_full_word_class = self.full_word_class
```
Madoshakalaka, line 217: I have some thoughts regarding this `full_word_class` property.

Our Wordform model has a non-standard `pos` field (as part of speech is depreacted), which should be replaced to a `word_class` field.

But if we are switching to a JIT paradigm, the way of calculating `full_word_class` on the fly is a good way to start.

BTW, shouldn't it be just "word_class"? The "full" is unnecessary to me.

eddieantonio: Sure! I'll rename it to `word_class`!

CreeDictionary/API/templatetags/inflection_extras.py
```diff
@@ -77,3 +79,36 @@ def presentational_pos(wordform: Union[Wordform, dict]) -> str:
         f"can not determine presentational pos for {wordform_dict}, id={wordform_dict['id']}"
     )
     return ""
+
+
+# XXX: mypy can't infer this type?
+PER_REQUEST_ID_COUNTER = WeakKeyDictionary()  # type: WeakKeyDictionary
```
Madoshakalaka, line 85: does `WeakKeyDictionary` help with memory leak because the keys are garbaged collected as soon as the reference counter to the dictionary is 0?? Want some explanation.

CreeDictionary/API/templatetags/inflection_extras.py
```diff
@@ -77,3 +79,36 @@ def presentational_pos(wordform: Union[Wordform, dict]) -> str:
...
+
+
+@register.filter
+def unique_id(context: Any) -> str:
```
Madoshakalaka, line 89: https://docs.djangoproject.com/en/3.0/howto/custom-template-tags/

So I think "filters"( `{{ ]}`) takes arbitrary value, and "tags" (`{% %}`) are the ones that are meant to take in context? You may want to use django's `@register.simple_tag(takes_context=True)` here

CreeDictionary/API/templatetags/inflection_extras.py
eddieantonio: Correct! When the `Response` object reaches a reference count of 0, the key in the dictionary is invalid, and thus, its data can be deleted

CreeDictionary/API/templatetags/inflection_extras.py
eddieantonio: I debated A LOT about this. You're right; context makes more sense, but the fact that I needed to assign and recall the ID twice makes creating a template tag... annoying... I'll give it some more thought :/

Issue #373, opened 2020-03-28
Prefix features should also end up on their own lines


aarppe:
When searching with a Cree word like _ê-kî-nitawi-kâh-kîmôci-kotiskâwêyâhk_, with preverbs (`PV/...+`) and reduplication elements (`RdplW+` and `RdplS+`), the preverbs currently do nt end up on their own line, in contrast to the suffix features:

![image](https://user-images.githubusercontent.com/15003407/77935914-59b21200-726f-11ea-8ca4-6675d391e1b0.png)

So, the visual presentation should be updated so that also prefixal features in the pop-up are shown each on their own line. In the above screenshot, that is NOT yet the case for the preverbs - they should appear similar to the suffix features, each on their own line.

aarppe:
@kobexamoh This fix is for you.

eddieantonio:
@aarppe, does this screenshot demonstrate what the issue you're talking about?

aarppe:
I've fixed the screenshot above, which illustrates the current undesired state of affairs, where the presentation of prefixal features (preverbs and reduplication, and initial change) is NOT YET  correct - they should appear similar to the suffix features, each on their own line.

eddieantonio:
Blocked by #462 — it will cause a merge conflict if I try to fix this simultaneously :/

PR #464, opened 2020-06-18
Each prefix feature is on its own line

eddieantonio:
~~Blocked by #462~~

Looks like this:

Screen Shot 2020-06-18 at 2 45 52 PM

There's probably a better way to present ALL of this, but this is better than how it used to be! 😄 

Fixes: #373

### DIFF
--- a/CreeDictionary/CreeDictionary/templates/CreeDictionary/components/_definition-title.html
+++ b/CreeDictionary/CreeDictionary/templates/CreeDictionary/components/_definition-title.html
@@ -43,16 +43,17 @@
 
         <div class="tooltip" role="tooltip">
           <span
-            data-cy="linguistic-breakdown">{% if result.linguistic_breakdown_head %}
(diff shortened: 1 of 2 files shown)


eddieantonio:
> Three questions/comments:
> 
> 1. Is the ê- preverb actually missing from the list of preverbs, or is that just a result of the screenshot?

FOR WHATEVER REASON, the ê- preverb is missing in the test database, but it's present in the production database. ¯\_(ツ)_/¯ 

> 2. At times, I've been considering whether one should show the prefixal features before/above the lemma/stem - but in any event this makes things much better in comparison to the current state.

Yeah! We should iterate on the presentation of this for sure, but just putting things on their own line is a huge improvement!

> 3. Though it may not be easy, we might want to consider shorter English relabelings for the preverbs, when presenting the linguistic analysis.

I think this is a great idea!

PR #471, opened 2020-06-21
Slightly better paradigm display

### LINKED ISSUE #470: Weird layout on paradigms
image

Not functionality breaking, but I thought I’d mention it!

eddieantonio:
- refactors the paradigm generation template to its own file
 - now all of the subtables are vertically aligned
 - adds a scrollbar to paradigms that are too wide for the screen (resolves #470)
 - renames the word detail template from "paradigm.html" to "word-detail.html"


src/css/utilities.css
```diff
@@ -36,3 +36,17 @@
...
   cursor: pointer;
 }
+
+/**
+ * Don't use this class.
+ *
+ * Instead, find a better way to wrap content in whatever this class is
+ * applied to :/
+ */
+.HACK-overflow-x-scroll {
```
eddieantonio, line 46: So the `div` that applies this class should ideally not exist, but it alleviates the problem reported in #470 for now.

CreeDictionary/CreeDictionary/views.py
```diff
@@ -105,7 +106,7 @@ def search_results(request, query_string: str):  # pragma: no cover
 @require_GET
 def lemma_details_internal(request):
     """
-    Render paradigm.html for a lemma. `index` view function renders a whole page that contains paradigm.html too.
+    Render word-detail.html for a lemma. `index` view function renders a whole page that contains word-detail.html too.
```
Madoshakalaka, line 109: much more apt naming!

PR #478, opened 2020-06-24
change: Delete fuzzy search

eddieantonio:
It's not a problem anymore!


CreeDictionary/API/models.py
```diff
@@ -325,6 +311,7 @@ class Meta:
         # text index also benefits fast lemma matching in function fetch_lemma_by_user_query
         indexes = [
             models.Index(fields=["analysis"]),
+            # TODO: Is this required?
```
Madoshakalaka, line 314: I think it is. We do have filter(text=xxx) in our code.

it's commented above

> text index also benefits fast lemma matching in function fetch_lemma_by_user_query

PR #483, opened 2020-06-26
refactor: move orthography related things to morphodict

eddieantonio:
This leaves Cree-specific orthography stuff in `CreeDictionary.orthography` and refactors language-agnostic stuff to the `morphodict` app. `morphodict` is configured in `settings.py`.

### DIFF
--- a/CreeDictionary/CreeDictionary/settings.py
+++ b/CreeDictionary/CreeDictionary/settings.py
@@ -162,7 +162,10 @@ MORPHODICT_ORTHOGRAPHY = {
             "name": "SRO (ēīōā)",
             "converter": "CreeDictionary.orthography.to_macrons",
         },
-        "Cans": {"name": "Syllabics", "converter": "cree_sro_syllabics.sro2syllabics",},
+        "Cans": {
+            "name": "Syllabics",
+            "converter": "CreeDictionary.orthography.to_syllabics",
+        },
     },
 }
 
--- a/CreeDictionary/CreeDictionary/settings.py
+++ b/CreeDictionary/CreeDictionary/settings.py
@@ -152,11 +152,23 @@ AUTH_PASSWORD_VALIDATORS = [
 
 ############################### Morphodict configuration ###############################
 
-# Configure the orthographies
+# The ISO 639-1 code is used in the lang="" attributes in HTML.
+MORPHODICT_ISO_639_1_CODE = "cr"
+
+# What orthographies -- writing systems -- are available
+# Plains Cree has two primary orthographies:
+#  - standard Roman orthography (e.g., nêhiyawêwin)
+#  - syllabics (e.g., ᓀᐦᐃᔭᐍᐏᐣ)
+#
+# There may be further sub-variants of each orthography.
+#
+# Morphodict assumes that the `text` of all Wordform are written in the default
+# orthography.
 MORPHODICT_ORTHOGRAPHY = {
-    # 'Latn' is Okimāsis/Wolvegrey's SRO
+    # All entries in Wordform should be written in SRO (êîôâ)
     "default": "Latn",
     "available": {
+        # 'Latn' is Okimāsis/Wolvegrey's SRO
         "Latn": {"name": "SRO (êîôâ)"},
         "Latn-x-macron": {
             "name": "SRO (ēīōā)",
--- a/CreeDictionary/CreeDictionary/settings.py
+++ b/CreeDictionary/CreeDictionary/settings.py
@@ -150,6 +150,19 @@ AUTH_PASSWORD_VALIDATORS = [
     {"NAME": "django.contrib.auth.password_validation.NumericPasswordValidator"},
 ]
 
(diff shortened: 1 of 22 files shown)


eddieantonio:
@Madoshakalaka, so this isn't quite ready yet, but I'd appreciate you review it anyway, because there's a lot of changes here already.

CreeDictionary/CreeDictionary/settings.py
```diff
@@ -150,6 +150,25 @@
...
+        "Latn-x-macron": {
+            "name": "SRO (ēīōā)",
+            "converter": "CreeDictionary.orthography.to_macrons",
+        },
+        "Cans": {
+            "name": "Syllabics",
+            "converter": "CreeDictionary.orthography.to_syllabics",
+        },
+    },
+}
```
eddieantonio, line 170: This is how orthographies in languages are configured. Note: given our talks with MESC, I think I might end up adding some more orthographies to this list for Plains Cree...

eddieantonio:
Lol, nm, it's ready. I think I should add a bit more documentation and comments, but it works, and `morphodict` is _almost_ entirely language-agnostic.

CreeDictionary/morphodict/templatetags/morphodict_orth.py
```diff
@@ -0,0 +1,76 @@
...
+        code: ORTHOGRAPHY.converter[code](original_text)
+        for code in ORTHOGRAPHY.available
+    }
+    inner_text = conversions[orthography]
+    data_attributes = " ".join(f'data-orth-{code}="{{}}"' for code in conversions)
+    values = tuple(conversions.values()) + (inner_text,)
+
+    return format_html(
+        # TODO: do not hardcode lang="cr" here
+        '<span lang="cr" data-orth ' + data_attributes + ">{}</span>",
```
eddieantonio, line 63: The `lang="cr"` is the only part that was too lazy to make language-agnostic. I'm not entirely sure where this should be configured, as the language code is not just about orthography.

Madoshakalaka: AppConfig is the class that :
> """representing a Django application and its configuration."""

maybe each language specific django app need to have a `lang` attribute in there `apps.py`> `AppConfig` class

Then each orthography option in `settings.MORPHODICT_ORTHOGRAPHY['available']` should specify the apps it work for. So that the lang code can be found in the app config.

Or if we have the implicit assumption that each converting function works only for the language of the app, the current `settings.MORPHODICT_ORTHOGRAPHY` will be totally fine as the App can be extracted from the module path.

CreeDictionary/morphodict/tests/test_orth_templatetags.py
```diff
@@ -0,0 +1,14 @@
...
+
+from morphodict.templatetags.morphodict_orth import orth
+
+
+def test_orth_requires_two_arguments():
+    """
+    orth() original only took one argument, but now it must take two.
+    """
+    with pytest.raises(TypeError):
+        orth("wâpamêw")
```
Madoshakalaka, line 14: glad to know tags/filters can be tested this way!!

CreeDictionary/tests/CreeDictionary_tests/test_creedictionary_extras.py
```diff
@@ -2,23 +2,14 @@
...
-def test_orth_requires_two_arguments():
-    """
-    orth() original only took one argument, but now it must take two.
-    """
-    with pytest.raises(TypeError):
-        orth("wâpamêw")
-
-
 def test_produces_correct_markup():
     context = Context({"wordform": "wâpamêw"})
```
Madoshakalaka, line 11: not sure why I missed this test earlier, but thanks for showing me how to test these filters and tags

CreeDictionary/tests/test_usage_of_morphodict.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+# -*- coding: UTF-8 -*-
+
+"""
+Integration test of settings.py and morphodict application.
```
Madoshakalaka, line 5: more like Unit test?

CreeDictionary/morphodict/templatetags/morphodict_orth.py
eddieantonio: Lol, I just made a new setting with the catchy name of `MORPHODICT_ISO_639_1_CODE`, since that's what it is.

I figure there would be a distinct `settings` module for each language that we support; I thinking having multiple languages support one `settings` file is just going to be a world of hurt :/

CreeDictionary/tests/test_usage_of_morphodict.py
eddieantonio: This is technically an integration test, because I made two units here: the `morphodict` application itself, and the settings defined in `CreeDictionary.settings`. In my view, this is testing the integration of the two. Does that make sense?

PR #515, opened 2020-07-22
docs: define lacuna (terribly)

eddieantonio:
I tried defining it but came up with a bad example. Suggestions?

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -263,15 +263,17 @@ lacuna
 [lacuna]: #lacuna
 [lacunae]: #lacuna
 
-"Gaps" in a [paradigm]. Any form that does not exist in a paradigm, where you would otherwise think it exists. For example, in Standard English, there isn't a "consistent" second person plural pronoun.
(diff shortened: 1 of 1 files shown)


aarppe:
The English pronoun _you_ is rather considered a historical development where it has come to denote both the singular and plural second person, since _thou/thee/thy_ has fallen out of use. Similarly, earlier English had different verb forms for the first, second, and third form singular, e.g. `I take, thou takest, s/he taketh/takes`, but the _-st_ and _-th_ forms have become archaic and replaced by the bare infinitive or the _-s_ - the results of this change are not usually understood as lacunae.

Pluralia tantum such as _pants_, _trousers_ is perhaps the closest you get to lacuna in English, in that generally English nouns can occur both in the singular and plural, but for some nouns things have evolved so that they synchronously no longer have a singular form.

A crucial perspective in this is that we are considering language as it is written (and spoken) at this moment, rather than contrasting the situation with previously existing states, though the latter can make some features of current English more understandable.

PR #517, opened 2020-07-27
Lacunae in the paradigm table

Madoshakalaka:
Em dash is shown in cells with FST analyses that generate nothing.

Cells without FST analyses are still shown with white space.


cypress/integration/paradigm.spec.js
```diff
@@ -59,6 +59,14 @@ describe(' I want to know if a form is observed inside a paradigm table', () =>
...
 })
 
+describe(' I want to see a clear indicator that a form does not exist', () => {
+
+  it.only('shows cells that does not exist as a em dash', ()=>{
+    cy.visitLemma('minôs')
+    cy.getParadigmCell('One', {colLabel: 'Smaller/Lesser/Younger'}).contains('—')
+  })
+})
+
```
Madoshakalaka, line 69: I am proud of `cy.getParadigmCell` @eddieantonio  :sunglasses:

PR #522, opened 2020-08-12
Prettier multiple recordings

kobexamoh:
At long last, prettier multiple recordings. This fix is based on the mockups previously demoed and is a bit more space efficient. As well, it's less JS written!

I'm leaving this in draft _for now_ because I'd love some feedback around things before it gets merged.

Anything you bring up is fair game but I'm especially interested in:

- are the buttons (play button, external link button) large enough to be interacted with? Should they be bigger for different screen sizes?
- How can the code structure fit BEM conventions better? I made use of an ID in one area – suggestions on working around this? (Eddie, this one's specifically for you 😅 )
- is there harm in hardcoding the SVG icons into the HTML (rather than using SVG sprites)?
- should the individual actions (playback of the recording, opening the recording in a new tab) be pulled out into their own functions?

(Might be still in vacation brain mode so if something's funky, let me know and I'll update – thaaanks)


kobexamoh:
Ninja edit: omg, the tests aren't passing. Gonna fix that 💀

src/recordings.js
```diff
@@ -18,14 +20,20 @@ export async function fetchFirstRecordingURL(wordform) {
...
 export function retrieveListOfSpeakers() {
   // get the value of the wordform from the page
   let wordform = document.getElementById('data:head').value
   let derivedURL = `${BASE_URL}/recording/_search/${wordform}`
 
-  let recordingsList = document.querySelector('.recordings-list')
+  let recordingsList = document.querySelector('.recordings-list') // to be deleted, but not yet because it'll break the refactoring 💀
+  
+  // select for our elements for playback and link-generation
+  let recordingsDropdown = document.getElementById('recordings-dropdown')
```
eddieantonio, line 33: So my philosophy with regards to IDs is that we should use them cautiously (or when HTML doesn't make give us any better option 💀 ).

I think the same thing can be accomplished by having just **ONE** ID for the entire `multiple-recordings` thing, then use child selectors to get the`<select>` and the `<button>`s and whatnot.

src/recordings.js
```diff
@@ -47,71 +55,41 @@ export function retrieveListOfSpeakers() {
...
-      // ...and place the newly clicked speaker's name into it
-      newSpeakerNode.slot['speaker-name'] = speakerName
-
-      // get the URL again and reinsert into the newly created node
-      newSpeakerNode.element.href = insertedURL
-
-      // place said node into the DOM
-      container.appendChild(newSpeakerNode.element)
-    }
+        for (let recordingData of recordings) {
```
eddieantonio, line 58: This is really minor, but what's the deal with the sudden huge indentation here?

src/recordings.js
```diff
@@ -47,71 +55,41 @@ export function retrieveListOfSpeakers() {
...
-      // place said node into the DOM
-      container.appendChild(newSpeakerNode.element)
-    }
+        for (let recordingData of recordings) {          
+          // TODOkobe: if all is working/Eddie has approved, perhaps refactor into a function?
+          // create a option element
+          let listOption = document.createElement('option')
+          
+          // create a new textnode that is the speaker's name
+          let nodeSpeakerName = document.createTextNode(recordingData.speaker_name + ', ' + recordingData.dialect)
```
eddieantonio, line 64: I would use a `<template>` for the code that generates the `<option>` ¯\\\_(ツ)\_/¯

src/recordings.js
```diff
@@ -47,71 +55,41 @@ export function retrieveListOfSpeakers() {
...
+          // TODOkobe: if there was to be a check for "if there's speakers with the same name, add something to the way they're rendered", it would end up right here as a loop within this loop,,,
+        }
+      
+        // audio playback for the specific speaker
+        recordingsPlayback.addEventListener('click', () => {
+          let speakerPosition = recordingsDropdown.selectedIndex
+          let audioURLPosition = recordings[speakerPosition].recording_url
+
+          // play the audio associated with that specific index
+          let audio = new Audio(audioURLPosition) // poorly named...position isn't it 💀
```
eddieantonio, line 81:
```suggestion
          let audioURL = recordings[speakerPosition].recording_url

          // play the audio associated with that specific index
          let audio = new Audio(audioURL)
```

src/recordings.js
```diff
@@ -47,71 +55,41 @@ export function retrieveListOfSpeakers() {
...
+        })
+
+        // link for the specific speaker
+        recordingsLink.addEventListener('click', () => {
+          let speakerPosition = recordingsDropdown.selectedIndex
+          let speakerBioLink = recordings[speakerPosition].speaker_bio_url
+          
+          // open the speaker's link in a new tab!
+          window.open(speakerBioLink, '_blank')
+        })
```
eddieantonio, line 93: Now that I think of it, perhaps this should just be a `<a target="_blank">` in the HTML, and require zero JavaScript code. We can use CSS to make the link look like the button, but act like a link.

src/recordings.js
kobexamoh: Oof, this is the only one I left as ✨ outstanding✨ . I can squeeze it in pre-general meeting I think but if you think it can come after the fact, I can modify it later too. Out of curiosity, why a template over this ~~hamfisted version~~?

src/recordings.js
kobexamoh: So I ended up going with a text link over the SVG icon: will that work? I figured it knocked out one of the pieces of design feedback (not immediately understanding what the external link button did).

I can always revert back to the SVG if you think that's better, though!

eddieantonio: Yes, text works for this!

src/recordings.js
eddieantonio: The `<template>` is in the HTML, which is easier to localize, and easier to edit by, e.g., @aarppe. Left as is, to change this text we'd have to dig into the JavaScript code, when this particular matter is more about the _presentation_ of the speaker name/dialect. JavaScript should, in my mind, avoid presentational concerns as much as possible. Does that make sense?

CreeDictionary/CreeDictionary/templates/CreeDictionary/word-detail.html
```diff
@@ -21,14 +21,17 @@ <h1 id="head" class="definition-title">
...
-      </div>
+    <section class="multiple-recordings definition__recordings--not-loaded" id="recordings-dropdown" data-cy="multiple-recordings">
+      <p class="multiple-recordings__help-text explainer">Choose a name from the dropdown to hear the word said by the speaker.</p>
+      <select name="recordings-dropdown" data-cy="recordings-dropdown" class="multiple-recordings__dropdown"></select>
+      <button class="multiple-recordings__action-button" data-action="play-current-recording" aria-label="Play recording" data-cy="play-selected-speaker">
+        <svg class="definition__icon speaker-playback-action-button" focusable="false">
+          <use xlink:href="#play-solid"/>
+        </svg>
+      </button>
+      <a class="multiple-recordings__action-button" data-action="learn-about-speaker" aria-label="Learn more about speaker" data-cy="learn-about-speaker" target="_blank" rel="noopener">
```
eddieantonio, line 32: 👍

src/recordings.js
kobexamoh: Yeah that makes sense! You want the JS to handle behaviour and not presentation – I gotchu 👌🏿

Issue #142, opened 2019-10-24
Prefix search should work for English as well


aarppe:
Now, if I search _sleep_, I only get _nipâw_, but not _nipâwin_ (sleeping).

I need to type _sleeping_ to get _nipâwin_.

We would want to get both _nipâw_ and _nipâwin_, when typing in the English _sleep_.

Another separate matter is matching also the English inflected forms of the search string, e.g. matching any English gloss with _slept_, while typing in _sleep_, or matching any English gloss with _sleep_, when typing _slept_ (or _sleeping_). This would seem integrally linked with the planned feature of being able to search Cree inflected word forms with sets of English keywords (a future feature to be considered later on).

Also, if one searches with _stove_, one should get the results for _stovepipe_ or _stove-pipe_ or _stove pipe_ (with a space between _stove_ and _pipe_) as well.

aarppe:
@Madoshakalaka When indexing the words in the English translations, what if we store not just the forms as they are, e.g. with a hyphen, but also without a hyphen, e.g. _stove-pipe_ is indexed as `stove-pipe` and `stovepipe`. Of course, we'd be indexing the crude stems of the words as well.

Then, when the user types any English, the hyphens and spaces are removed when searching the English search index, using a prefix search. Thus, _stove_ should match _stovepipe_ and _stove-pipe_, and _stovepipe_ should match _stovepipe_,  _stove-pipe_ should match _stovepipe_, and _stove pipe_ should match _stovepipe_. And as Eddie suggested, we'd run the stemmer on the search string as well, to match with the stemming in the index.

PR #523, opened 2020-08-19
Deprecate engcrk.xml

Madoshakalaka:
engcrk.xml now is extracted at runtime and no longer needed. Also with documentation cleanup.

### DIFF
--- a/CreeDictionary/API/models.py
+++ b/CreeDictionary/API/models.py
@@ -42,6 +42,7 @@ from utils import (
     get_modified_distance,
 )
 from utils.cree_lev_dist import remove_cree_diacritics
+from utils.english_keyword_extraction import stem_keywords
 from utils.fst_analysis_parser import LABELS, partition_analysis
 from utils.types import ConcatAnalysis, FSTTag, Label
 
@@ -534,11 +535,11 @@ class Wordform(models.Model):
         # todo: allow inflected forms to be searched through English. (requires database migration
         #  since now EnglishKeywords are bound to lemmas)
         english_results: Set[EnglishResult] = set()
-        if " " not in user_query:  # a whole word
+        for stemmed_keyword in stem_keywords(user_query):
 
             # this requires database to be changed as currently EnglishKeyword are associated with lemmas
             lemma_ids = EnglishKeyword.objects.filter(
-                text__iexact=user_query, **extra_constraints
+                text__iexact=stemmed_keyword, **extra_constraints
             ).values("lemma__id")
             for wordform in Wordform.objects.filter(
                 id__in=lemma_ids, as_is=False, **extra_constraints
--- a/CreeDictionary/DatabaseManager/english_keyword_extraction.py
+++ b/CreeDictionary/DatabaseManager/english_keyword_extraction.py
@@ -1,0 +1,19 @@
+from typing import Set
+import re
+
+word_pattern = re.compile(r"[\w.\-/']+")
+
+stop_words = {"a", "an", "the", "s/he", "s.o.", "s.t."}
+
+
+def extract_keywords(translation: str) -> Set[str]:
+    """
+    returns lowercased keywords from the translation, by stripping stop words like a/an s/he s.o. s.t.
+
+    >>> extract_keywords("s/he is of a that kind (previously mentioned)") == {'is', 'of', 'that', 'kind', 'previously', 'mentioned'}
+    True
+    """
+    words = set(word_pattern.findall(translation.lower()))
+
+    # .strip('.') is because of the trailing periods in some translation
+    return {word.strip(".") for word in words if word not in stop_words}
--- a/CreeDictionary/DatabaseManager/english_keyword_extraction.py
+++ b/CreeDictionary/DatabaseManager/english_keyword_extraction.py
@@ -1,19 +1,27 @@
-from typing import Set
 import re
+from typing import List
+
+import snowballstemmer
 
 word_pattern = re.compile(r"[\w.\-/']+")
 
 stop_words = {"a", "an", "the", "s/he", "s.o.", "s.t."}
 
+stemmer = snowballstemmer.EnglishStemmer()
 
-def extract_keywords(translation: str) -> Set[str]:
+
+def extract_keywords(translation: str) -> List[str]:
     """
     returns lowercased keywords from the translation, by stripping stop words like a/an s/he s.o. s.t.
 
-    >>> extract_keywords("s/he is of a that kind (previously mentioned)") == {'is', 'of', 'that', 'kind', 'previously', 'mentioned'}
-    True
+    returned list has no duplicates
+
+    >>> sorted(extract_keywords("s/he is of a that kind (previously mentioned)"))
+    ['is', 'kind', 'mention', 'of', 'previous', 'that']
     """
     words = set(word_pattern.findall(translation.lower()))
 
     # .strip('.') is because of the trailing periods in some translation
-    return {word.strip(".") for word in words if word not in stop_words}
+    return stemmer.stemWords(
+        {word.strip(".") for word in words if word not in stop_words}
+    )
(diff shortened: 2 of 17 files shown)


CreeDictionary/DatabaseManager/english_keyword_extraction.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    returns lowercased keywords from the translation, by stripping stop words like a/an s/he s.o. s.t.
+
+    >>> extract_keywords("s/he is of a that kind (previously mentioned)") == {'is', 'of', 'that', 'kind', 'previously', 'mentioned'}
+    True
+    """
+    words = set(word_pattern.findall(translation.lower()))
+
+    # .strip('.') is because of the trailing periods in some translation
+    return {word.strip(".") for word in words if word not in stop_words}
```
Madoshakalaka, line 19: @aarppe this is the incarnation of your awk script, as you may wanted, the `stop_words` set is the same.

cypress/integration/regressions.spec.js
```diff
@@ -40,11 +40,6 @@ context('Regressions', () => {
 
   // https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/158
   it('should display relevant English results', () => {
-    cy.visitSearch('see')
-    cy.get('[data-cy=search-results]')
-      .should('contain', 'wâpiw')
-      .and('contain', 'wâpahtam')
-      .and('contain', 'wâpamêw')
```
Madoshakalaka, line 47: @aarppe @eddieantonio wapiw has translation "s/he sees, s/he has vision CW", with the current runtime keyword extraction, you'll have to type in "sees" have the result.

The English stemming/lemmatization can be improved by an English FST

libexec/crkeng_stop_word_stat.py
```diff
@@ -0,0 +1,135 @@
...
+ ('things', 1233),
+ ('with', 1228),
+ ('by', 1201),
+ ('or', 1136),
+ ('people', 1083),
+ ('on', 1064),
+ ('makes', 999),
+ ('e.g.', 867),
+ ('as', 859),
+ ('his/her', 712),
```
Madoshakalaka, line 57: @eddieantonio @aarppe I wrote a script that does statistics about stop words. Currently only five are used. Do you think we can be less conservative and use more than five of them?

cypress/integration/regressions.spec.js
Madoshakalaka: @aarppe sorry, after some thinking I realized this is really bad. I'll add a dumb English stemmer real quick, in the process of English keyword extraction

libexec/crkeng_stop_word_stat.py
eddieantonio: I'd prefer a search index with tf.idf or Okapi BM25 instead of stop words, as it takes into account these statistics automatically.

CreeDictionary/tests/API_tests/model_test.py
```diff
@@ -102,12 +102,22 @@ def test_search_for_exact_lemma(lemma: Wordform):
...
+
+    # s = snowballstemmer.EnglishStemmer()
+
+    # >>> s.stemWord('story')
+    # 'stori'
+    # >>> s.stemWord('stories')
+    # 'stori'
+
     # This should match "âcimowin" and related words:
     search_results = Wordform.search("story")
```
Madoshakalaka, line 122: limitation of our stemmer

libexec/crkeng_stop_word_stat.py
Madoshakalaka: sounds cool!

cypress/integration/regressions.spec.js
Madoshakalaka: @aarppe @eddieantonio now a stemmer is used and things goes well :)

CreeDictionary/DatabaseManager/english_keyword_extraction.py
```diff
@@ -0,0 +1,27 @@
...
+import snowballstemmer
+
+word_pattern = re.compile(r"[\w.\-/']+")
+
+stop_words = {"a", "an", "the", "s/he", "s.o.", "s.t."}
+
+stemmer = snowballstemmer.EnglishStemmer()
+
+
+def extract_keywords(translation: str) -> List[str]:
```
eddieantonio, line 13: Why was this changed to `List`?

CreeDictionary/tests/API_tests/model_test.py
eddieantonio: Wait, it doesn't match story?

cypress/integration/search.spec.js
```diff
@@ -370,7 +371,7 @@ context('Searching', () => {
   describe('display of the header', function () {
     const lemma = 'nîmiw'
     const wordclassEmoji = '➡️' // the arrow is the most consistent thing, which means verb
-    const inflectionalCategory = 'VAI-1'
+    const inflectionalCategory = 'VAI-v'
```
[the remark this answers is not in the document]
aarppe: `VAI-v` is an old CW code which has been replaced in the latest CW version with `VAI-1/2/3`.

However, there are a few dictionary entries where I've also retained the old entry and the old wordclass code, so recognizing `VAI-v` is a useful fallback option, though on the longer term that old coding will no longer be used.

libexec/crkeng_stop_word_stat.py
aarppe: For future reference, the short stop word list was motivated by ruling out articles/determiners `a`, `an`, `the` and the most formulaic of CW conventions, i.e. `s/he`,` s.o.` and `s.t.`. The motivation was the users might have good reasons to search with common prepositions like `of` and `with`.

On the longer term, such elements which are grammatical in English and which do not always have a correspondent in Cree might be replaced with a "grammar entry", explaining how the grammatical function conveyed by the English preposition (or the like) can be (morpho-syntactically) communicated in Cree (for lack of a proper matching word).

Madoshakalaka:
@aarppe @eddieantonio There's also a nice little behaviour change with this PR.

Spaces used to stop English search like this, simply no result will be presented based on the English:

# Old Behavior

Screenshot from 2020-08-19 15-57-28

It will now at least try and match each word:

# New Behavior

Screenshot from 2020-08-19 15-57-19

I imagine presenting _atim_ when somebody searches for "labrador dog" will make them somewhat happy, rather than presenting nothing.

aarppe:
Concerning _stove_ above, we'd still want all of the below, as enumerated in #142, i.e.

- [ ] `stove` should match _stovepipe_
- [ ] `stove-pipe` should match _stovepipe_
- [ ] `stove pipe` should match _stovepipe_

eddieantonio:
> Concerning _stove_ above, we'd still want all of the below, as enumerated in #142, i.e.
> 
> * [ ]  `stove` should match _stovepipe_
> * [ ]  `stove-pipe` should match _stovepipe_
> * [ ]  `stove pipe` should match _stovepipe_

This one is going to require more thought and consideration, so I'm making it a separate issue! #524

CreeDictionary/API/models.py
```diff
@@ -534,11 +535,11 @@ def fetch_lemma_by_user_query(
         # todo: allow inflected forms to be searched through English. (requires database migration
         #  since now EnglishKeywords are bound to lemmas)
         english_results: Set[EnglishResult] = set()
-        if " " not in user_query:  # a whole word
+        for stemmed_keyword in stem_keywords(user_query):
 
             # this requires database to be changed as currently EnglishKeyword are associated with lemmas
             lemma_ids = EnglishKeyword.objects.filter(
-                text__iexact=user_query, **extra_constraints
+                text__iexact=stemmed_keyword, **extra_constraints
```
eddieantonio, line 542: This is great!

Issue #529, opened 2020-08-21
Debounce queries when typing in the search bar


eddieantonio:
Right now, we send queries per every keystroke, even if the user continues to type more. Here's an idea:

We [debounce][] the input events, and send a search request when it appears the user has stopped typing. I also feel like this would make our Cypress tests run faster! 

[debounce]: https://levelup.gitconnected.com/debounce-in-javascript-improve-your-applications-performance-5b01855e086?gi=2679f37f3d2e#:~:text=The%20debounce%20function%20delays%20the,calls%20sent%20to%20your%20server.

Madoshakalaka:
@dwhieb Thanks! I'm gonna yoink it :^)

Madoshakalaka:
@eddieantonio 

I don't think I executed it good enough. I checked on production and typed into the search bar _wapamat_, it appears to me now all requests are just delayed, not debounced --- we now have 7 simultaneous requests for _wapamat_.

![Screenshot from 2020-09-01 15-23-58](https://user-images.githubusercontent.com/44753941/91907502-37ca1a80-ec67-11ea-8350-9cae5b119b2e.png)

I'll have a follow up on this, please keep this issue open

PR #538, opened 2020-09-01
Debouncing User Queries

Madoshakalaka:
As #529 mentions, typing on the search bar overwhelms our server by sending too many requests.

Now search requests will only go off after 450 miliseconds of inactivity. 

Notably, you can still press Enter / use the magnifying glass button to speed up and save the 450 miliseconds. I actually contributed nothing towards this. As a surprising satisfaction, thanks to our good HTML practices. The search bar is coded in HTML as a form with action=submit, so entering/pressing button will just refresh the page with the already changed url on the page.

Why 450 miliseconds? I did some research and wrote a [blog post](https://madoshakalaka.github.io/2020/08/31/how-hard-should-you-debounce-on-a-responsive-search-bar.html) about it :sunglasses:

# refactoring to integration test

After the debouncing, one integration test involving typing on the search bar times out cuz of the extra 450 miliseconds. It could be prevented by pressing `{enter}` or waiting after the typing. Either ways, we want searching with enter key / wait to be consistent in the tests. 

Included is a refactoring to cypress tests, we have an extra command `cy.search` that types on the search bar:

`cy.search('minos', {pressEnter:false, waitTime: 500})`

 Old test cases were inconsistent about how search were performed, some test cases for example uses `cy.visitSearch()` to visit the url, some test case uses `cy.get([data-cy=search-bar]).type()`. Now all of them uses `cy.visitSearch()` except the ones that explicitly tests for url/typing one by one, which now uses the new command `cy.search`

There is also a QoL command `cy.searchResultsContain`, so this is possible:

`cy.search('minos').searchResultsContain('cat')`

### DIFF
--- a/cypress/integration/search.spec.js
+++ b/cypress/integration/search.spec.js
@@ -1,30 +1,49 @@
 context('Searching', () => {
+  
+  describe('I want to see search results showing dynamically while I type', () => {
+    // https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/120
+    it('should search for first minos, then minosis', () => {
+      cy.visit('/')
+      cy.search('minos')
+        .searchResultsContain('cat')
+
+      // makes it minos+is = minosis
+      cy.search('is')
+        .searchResultsContain('kitten')
+    })
+  })
+
+  describe('I want to see search results showing immediately after pressing enter', () => {
+    // https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/120
+    it('shows results for minosis immediately after enter is pressed', () => {
+      cy.visit('/')
+      cy.search('minosis', {pressEnter:true})
+        .searchResultsContain('kitten')
+
+    })
+  })
+  
+  
+  describe('I want to see search results by the url', () => {
+    it('perform the search by going directly to the URL', () => {
+      cy.visitSearch('minos')
+        .searchResultsContain('cat')
+    })
+  })
+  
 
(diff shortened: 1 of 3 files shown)


cypress/integration/search.spec.js
```diff
@@ -1,30 +1,49 @@
...
+      cy.search('minos')
+        .searchResultsContain('cat')
 
-      cy.get('[data-cy=search-results]')
-        .should('contain', 'cat')
+      // makes it minos+is = minosis
+      cy.search('is')
+        .searchResultsContain('kitten')
     })
+  })
```
Madoshakalaka, line 14: This is a new test that directly tests for typing on the search bar

eddieantonio: This is SOOOO NICE 🤩  🤩  🤩

cypress/support/commands.js
```diff
@@ -244,4 +244,45 @@ Cypress.Commands.add('getParadigmCell', {prevSubject: false}, (rowLabel, {colLab
...
+Cypress.Commands.add('clearSearchBar', {prevSubject: false}, () => {
+  cy.get('[data-cy=search]').clear()
+})
+
+/**
+ * this command clears the search bar
+ */
+Cypress.Commands.add('searchResultsContain', {prevSubject: false}, (expectedString) => {
+  return cy.get('[data-cy=search-results]')
+    .should('contain', expectedString)
```
eddieantonio, line 287: These are VERY nice!

PR #539, opened 2020-09-01
Fix Debouncing

Madoshakalaka:
previously everytime on input, `debounce(loadsearch)` is called to instantiate a new function. This is simply not how debounce is meant to be used. A debounced-function should be instantiated and reused. The wrong usage results in delayed but not debounced requests as mentioned in #529 

Now it's fixed.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -224,7 +224,7 @@ function debounce(func, wait, immediate) {
 const debouncedLoadSearchResults = debounce(() => {
   const searchBar = document.getElementById('search')
   loadSearchResults(searchBar)
-}, 700)
+}, 450)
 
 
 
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -221,16 +221,18 @@ function debounce(func, wait, immediate) {
   }
 }
 
+const debouncedLoadSearchResults = debounce(() => {
+  const searchBar = document.getElementById('search')
+  loadSearchResults(searchBar)
+}, 450)
```
eddieantonio, line 227: While we're at it, could you please extract this into a top-level `const` and add a comment on how you got this value (i.e., link to your blog post!)

Madoshakalaka: sure!

PR #552, opened 2020-10-08
change: make the other Cree words links

eddieantonio:
I was upset that only ê-kî-nitawi-kâh-kîmôci-kotiskâwêyâhk was the only Cree word that was a link, so I made the other ones links too


aarppe:
@eddieantonio Should _itwêwina_ itself be linkable?

eddieantonio:
Good point @aarppe! I also added links to for us and our partners on the front page!

PR #556, opened 2020-10-21
English Prefix Suffix Search

anAntelope:
So sleep yields sleeping now. And berry yields bearberry.

SharedScreenshot

----
It's Matt. I have to use serena's account for the time being. I have problems with two-factor auth for github login cuz my sim card doesn't work in China. I've requested an account recovering and it takes human verification and time.


CreeDictionary/CreeDictionary/hfstol.py
```diff
@@ -31,7 +31,7 @@
 def analyze(wordform: str) -> Iterable[Analysis]:
     with write_temporary_file(f"{wordform}\n") as input_file:
         output = check_output(
-            [_hfstol, "-q", _analyzer_path],
+            [_hfstol, "-q", str(_analyzer_path)],
```
anAntelope, line 34: windows compatibility

anAntelope:
This should resolve #142

CreeDictionary/morphodict/search.py
```diff
@@ -0,0 +1,840 @@
+"""
```
eddieantonio, line 1: Was this commited by mistake? It looks like an old version of `CreeDictionary/API/models.py` that was copy-pasted into `CreeDictionary/morphodict`?

eddieantonio (changes requested):
Uh... please remove `CreeDictionary/morphodict/search.py`???

CreeDictionary/morphodict/search.py
anAntelope: removed, accidentally commited

eddieantonio (approved):
LGTM!

PR #570, opened 2020-11-17
Use `ENABLE_DJANGO_DEBUG_FLAG` instead of relying on CI

eddieantonio:
I'm making this option explicit, as there is reason to run in `DEBUG` mode and not want the toolbar on a local dev machine. Also, it should _always_ be turned off in CI!


Madoshakalaka (approved):
LP(erfect)TM!

PR #575, opened 2020-12-04
fix: use latest FSTs (2020-12-04)

### LINKED ISSUE #573: Inflection typo
There seems to be a regression introduced recently for VII verbs in the obviative forms:

screenshot 2020-12-02 at 21 32 04

`kiwêmakanyiw` should be `kiwêmakaniyiw`, etc. I'm *think* I checked this a week or so ago and it was correct?

eddieantonio:
fixes #573


eddieantonio:
Actually, this doesn't fix it at all :/ 

```sh
 $ echo "kîwêmakan+V+II+Ind+4Sg" | hfst-lookup -s ./crk-normative-generator-with-morpheme-boundaries.hfstol
kîwêmakan+V+II+Ind+4Sg	kîwêmakan>yiw	0.000000
```

PR #580, opened 2020-12-12
Refactor affix search test

eddieantonio:
I was curious to see what would happen if I used the `client` fixture from `pytest-django`. It works! Though maybe that's because the two requests were placed into two separate functions :/

### DIFF
--- a/CreeDictionary/tests/API_tests/test_views.py
+++ b/CreeDictionary/tests/API_tests/test_views.py
@@ -1,8 +1,9 @@
+import time
+
 import pytest
+from django.conf import settings
 from django.test import Client
 from django.urls import reverse
-from django.conf import settings
-import time
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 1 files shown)


CreeDictionary/tests/API_tests/test_views.py
```diff
@@ -21,71 +22,49 @@ def django_db_setup():
...
-    # It seems like something closes the database connection after the first client.get()
-    # And I can't find related bug/issue reports on the web
-    # The moral is, whenever you need multiple client.get()'s, split into multiple function
-
-    # Think positively too! The above class pattern is so much more readible and clearer!
-
-    # @pytest.mark.django_db
-    # def test_click_in_text_no_affix_search():
-    #
-    #     c = Client() # tried: seperate client instances c1, c2 won't help
```
eddieantonio, line 82: Honestly, I think it's just an issue of using any amount of client instances more than once. But what could be causing it...?

Madoshakalaka:
All fine. I am happy that I wrote stuff you can delete. It's happiness shared.

PR #587, opened 2020-12-22
CI/CD for @altlab/types package 

Madoshakalaka:
A new public package [@altalb/types](https://www.npmjs.com/package/@altlab/types) is published to npm, which has a single `index.ts` and `REAMDE.md`

The package is managed by Github Actions and generated from source `schema.py`. A very considerate test is added to `test-and-deploy.yml` that checks if the generated file becomes different from the published version, in which case we should bump the package version according to semvar.


.github/workflows/check_types_version_bump.py
```diff
@@ -0,0 +1,51 @@
+"""
+check if we need to bump the version of @altlab/types package
+The fact that this is only needed during github actions make me not place it under libexec directory.
+"""
+import argparse
+import filecmp
+import json
+import sys
+from difflib import context_diff
```
Madoshakalaka, line 9: difflib and filecmp exists as built-in modules?????

Madoshakalaka:
@eddieantonio 
Here is an example of when our new action covers your back when you forget to bump the package version.

https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/runs/1592148283?check_suite_focus=true#step:5:40

package.json
```diff
@@ -22,26 +22,26 @@
...
-    "rollup-plugin-postcss": "^2.0.3",
-    "rollup-plugin-terser": "^5.3.0",
+    "rollup-plugin-postcss": "^2.9.0",
+    "rollup-plugin-terser": "^5.3.1",
     "stop-only": "^3.1.0",
     "wait-on": "^3.3.0"
   },
   "dependencies": {
-    "@popperjs/core": "^2.0.5"
+    "@popperjs/core": "^2.6.0"
```
Madoshakalaka, line 45: this is from an npm install. No dependency is upgraded.

pyproject.toml
```diff
@@ -11,5 +11,12 @@ skip = [
...
\ No newline at end of file
+
+
+[project_paths]
+# A place to store paths to different components in our project.
+# Here they are all written relative to project root.
+# we have get_project_path function in CreeDictionary.utils module that resolves to these paths
+
+# the file we write documentation for directories
+dir_doc_file = "docs/directory-structure.md"
```
Madoshakalaka: This is a refactory. An utility is added to resolve project paths.

CreeDictionary/utils/project_path.py
```diff
@@ -0,0 +1,39 @@
...
+                                       "project_paths"].items()}
+
+
+def get_project_path(name: str) -> Optional[Path]:
+    """
+    reads in project paths from pyproject.toml
+
+    :param name: name as documented in pyproject.toml
+    """
+    return _PROJECT_PATHS.get(name)
```
Madoshakalaka, line 39: This is a refactory

CreeDictionary/API/schema.py
```diff
@@ -30,7 +30,7 @@ class SerializedWordform(TypedDict):
     wordclass_emoji: str
 
     # ---- foreign keys ----
-    definition: List[SerializedDefinition]
+    definitions: List[SerializedDefinition]
```
Madoshakalaka, line 33: schema was actually WRONG, actually API gives definitions instead of definition

CreeDictionary/utils/project_path.py
eddieantonio: I have a VERY COOL but VERY EXTRA and VERY UNNECESSARY idea for this. It involves creating a class with `__getattr__` that exposes all of the paths in pyproject.toml as a `Path` instance, if it exists.

CreeDictionary/utils/project_path.py
```diff
@@ -0,0 +1,39 @@
...
+    bubbles up until we see a directory with .git folder
+
+    :return: The root directory of the repository
+    """
+    # Uhh I actually don't know any reliable way to do this
+    # this sounds like a reasonable method. You are Welcome to fix it.
+    current = FILE_DIR
+    while not (current / ".git").is_dir():
+        current = current.parent
+    return current
```
eddieantonio, line 23: This makes the assumption that the code is always in a git repository, which is not always the case. I'm not sure of a fool-proof way of doing this without making some assumption that can easily change :/

I think a better assumption is the `pyproject.toml` since it IS a runtime dependency — `.git` is technically a development thing.

```suggestion
    # Uhh I actually don't know any reliable way to do this
    # this sounds like a reasonable method. You are welcome to fix it.
    current = FILE_DIR
    while not (current / "pyproject.toml").is_file():
        current = current.parent
    return current
```

CreeDictionary/tests/test_docs.py
```diff
@@ -105,15 +86,14 @@ def format_fix_suggestion_as_lines_relative_to_project(
...
-    parsed_tml = toml.loads((PROJECT_ROOT / "pyproject.toml").read_text())
 
     # if we were to use the vanilla set, the suggestions would come out in alphabetical order
     # which makes less sense than depth first order in terms of directory presentation
     ones_with_docs = SortedSet()
     non_existent_dirs = SortedSet()
     non_dir_docs = SortedSet()
 
-    doc_file = Path(PROJECT_ROOT / parsed_tml["tool"]["dirdoc"]["doc_file"])
+    doc_file = get_project_path("dir_doc_file")
```
eddieantonio, line 96: Ah, much cleaner! ☺️

.github/workflows/check_types_version_bump.py
```diff
@@ -0,0 +1,51 @@
+"""
+check if we need to bump the version of @altlab/types package
+The fact that this is only needed during github actions make me not place it under libexec directory.
```
eddieantonio, line 3: lol, you could see my review comment coming! 😂

eddieantonio (changes requested):
A few changes, but I can make them if you can't.

CreeDictionary/utils/project_path.py
Madoshakalaka: That's a really cool idea

PR #601, opened 2021-01-11
Use cy.intercept()

eddieantonio:
There were a few integration tests that made network requests. It used to be that these could not tested due to limitations with `cy.server()`/`cy.route()`; now, these can use `cy.intercept()` to mock!

Resolves #591 

(Edit: better wording)


cypress/integration/search/result-header.spec.js
```diff
@@ -0,0 +1,243 @@
...
+        .click()
+      cy.get('@elaboration')
+        .get('[role="tooltip"]')
+        .should('be.visible')
+      cy.get('@elaboration')
+        .contains('[role="tooltip"]', inflectionalCategory)
+    })
+
+    // See: https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/445#:~:text=6.%20Lemma%20definition
+    it('should display an inflected form and its lemma', function () {
```
nienna73, line 174: The name of this test is identical to the one above it--is this intentional?

eddieantonio: Probably not! Thanks for this!

nienna73 (approved):
I was able to run the tests locally. Some offline discussions lead to updated docs for clarity. The refactor and code changes look good!

PR #608, opened 2021-01-13
Let logging from non-Django code reach console

andrewdotn:
- With a root handler, loggers from other code, e.g., in `utils`, will
    produce output on the console

  - Django installs its own handler by default, but if we let it propagate
    to the root handler, the `django` logger doesn’t need its own handler

This does eliminate some default features of Django that we’re not using,
such as mailing admins a copy of any error messages. But if we want to use
that, we can explicitly set it up later.

### DIFF
--- a/CreeDictionary/CreeDictionary/settings.py
+++ b/CreeDictionary/CreeDictionary/settings.py
@@ -300,20 +300,28 @@ else:
         "django.contrib.staticfiles.storage.ManifestStaticFilesStorage"
     )
 
+log_level = env.log_level("LOG_LEVEL", default="INFO")
+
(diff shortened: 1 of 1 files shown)


eddieantonio (approved):
LGTM!

This logging configuration is pretty cargo cult for me, so I'm glad you cleaned it up!

eddieantonio:
Side note: you can make branches directly on upstream; no need to branch! Is there a reason you're working in a fork?

andrewdotn:
Oh, I’m just so used to working on repos that I don’t have (or necessarily want) push access to, that forking is my default mode of operation.

PR #609, opened 2021-01-13
refactor: everything affix search

eddieantonio:
Refactoring A LOT about affix search. Let me know if I should squash any commits :/

### DIFF
--- a/CreeDictionary/API/apps.py
+++ b/CreeDictionary/API/apps.py
@@ -68,12 +68,9 @@ def initialize_affix_search():
         logger.exception("Cannot build tries: Wordform table does not exist (yet)!")
         return
 
-    lowered_no_diacritics_cree_with_id = [
-        (text, wf_id)
-        for text, wf_id in Wordform.objects.filter(is_lemma=True).values_list(
-            "text", "id"
-        )
-    ]
+    lowered_no_diacritics_cree_with_id = list(
+        Wordform.objects.filter(is_lemma=True).values_list("text", "id")
+    )
 
     lowered_english_keywords_with_wf_id = [
         (kw, wf_id)
--- a/CreeDictionary/API/apps.py
+++ b/CreeDictionary/API/apps.py
@@ -1,6 +1,5 @@
 import logging
 from pathlib import Path
-from itertools import chain
 
 from django.apps import AppConfig
 from django.db import OperationalError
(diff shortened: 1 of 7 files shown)


eddieantonio:
So that's what happens when I do some INTENSE refactors. There's... a lot of commits...

Let me know if I should squash some and force push. I... don't like doing that... but this is definitely one of my more absurd PRs in terms of number of commits! 😅

CreeDictionary/API/models.py
```diff
@@ -236,13 +238,30 @@ def save(self, *args, **kwargs):
...
+    @staticmethod
+    def search_with_affixes(query: str) -> SortedSet["SearchResult"]:
+        """
+        Search for wordforms matching:
+         - the wordform text
+         - the definition keyword text
+         - affixes of the wordform text
+         - affixes of the definition keyword text
+        """
+        from .search import WordformSearchWithAffixes
```
andrewdotn, line 250: Is this here because of the import cycle issue?

andrewdotn (approved):
Looks good overall.

If this happens to fix the duplicate-definition problem, we should have a test for that.

I am curious about the import cycle issue, but that’s probably a lot easier to talk about in a meeting than in GitHub comments.

No concerns from me about the number of commits.

CreeDictionary/API/models.py
eddieantonio: Yes. Although now that I reflect on it, I think having static methods for `Wordform.simple_search()` and `Wordform.search_with_affixes()` is not worth the ugly import cycle issues, so perhaps we should move that elsewhere!

CreeDictionary/API/apps.py
```diff
@@ -56,47 +65,45 @@ def read_morpheme_rankings():
...
         # exists. Then an OperationalError will occur.
-    except OperationalError:
+        logger.exception("Cannot build tries: Wordform table does not exist (yet)!")
         return
 
-    Wordform.affix_searcher = AffixSearcher(
-        lowered_no_diacritics_cree_with_id + lowered_english_keywords_with_wf_id
-    )
+    set_affix_searcher_for_cree(AffixSearcher(fetch_cree_lemmas_with_ids()))
+    set_affix_searcher_for_english(AffixSearcher(fetch_english_keywords_with_ids()))
```
nienna73, line 89: This is super nit-picky:
```suggestion
    set_affix_searcher_for_english(AffixSearcher(fetch_english_lemmas_with_ids()))
```
The use of "lemmas" in one function and "keywords" in the other makes me think they do different things. If that's the case, ignore this suggestion.

eddieantonio: They do different things! So this confusing wording is entirely intentionally. The results that can be matched are _lemmas_ — base form of words defined in the dictionary and FSTs, while the results that can be matched for English are _keywords_, mined from the dictionary definitions.

nienna73 (approved):
I made a few comments that are optional to address, but might provide some clarity to future developers. Otherwise, it looks good!

PR #610, opened 2021-01-14
refactor: more affix search refactoring

eddieantonio:
Addresses some things that have been bothering me:

 - [x] global affix search instances should not be part of `Wordform`; instead, they're created on a specific `AppConfig` instance (these are globally available, because Django)
 - [x] change ONE of the  `fetch_cree_and_english_results()` to `fetch_bilingual_results()`


nienna73 (approved):
These look good to me, thanks for changing that function name!

PR #617, opened 2021-01-18
refactor: use shared generator FST

eddieantonio:
The paradigm filler (the thing that fills in all the cells in a paradigm) used to instantiate its own copy of the normative generator (the thing that spits out perfect SRO, given an analysis). But only one copy is needed in the application, so use that one copy!


andrewdotn (approved):

aarppe:
Can this be modified easily to use the normative generator with morpheme boundaries? Because it is the morpheme boundaries we'd want to eventually see in the paradigms, and in the standardized representation of the search string (as an option the user can choose to select).

eddieantonio:
> Can this be modified easily to use the normative generator with morpheme boundaries? Because it is the morpheme boundaries we'd want to eventually see in the paradigms, and in the standardized representation of the search string (as an option the user can choose to select).

Yes and no. It's out of the scope of this particular PR (the PR is already merged and running on Sapir!). We could plop in the FST with morpheme boundaries but nothing in itwêwina knows how to deal with morpheme boundaries, so it will treat it like alphabetic text (i.e., we'll see results like "ni<wâpam>âw" with no option to remove the "<" and ">). We could add some further hacks  to the FST loader code to remove morpheme boundaries before they make their way into the rest of itwêwina, but I'd like to get rid of existing hacks first! (see #620)

The better option is simply to integrate the FST with morpheme boundaries when we start to implement morpheme boundaries in itwêwina proper. That way, we wouldn't require any hacks!

aarppe:
Ok.

My comment had the longer term in mind, when I'd think we could get by using only the normative generator FST with boundaries, and then have the code on the itwêwina side reinterpret the morpheme boundaries in whichever way in its internal structure, and then either represent the boundaries appropriately, e.g. with the middle-dot, or removing them entirely, based on user settings.

PR #621, opened 2021-01-19
remove: admin interface

eddieantonio:
Remove the entire admin interface. It is unused and untested. I think we used it in summer of 2019, but the current database does not even have any superusers.


andrewdotn:
I know this code is unused, so go ahead and delete it, but I do dream of the admin interface being useful for directly marking distinct definitions from CW/MD as “close enough” to combine, for quickly adding alternate spellings (amiskwaciy-wâskahikan = amiskwacîwâskahikan = amiskwaciwâskahikan), and so on.

eddieantonio:
> I know this code is unused, so go ahead and delete it, but I do dream of the admin interface being useful for directly marking distinct definitions from CW/MD as “close enough” to combine, for quickly adding alternate spellings (amiskwaciy-wâskahikan = amiskwacîwâskahikan = amiskwaciwâskahikan), and so on.

I believe @aarppe would like this functionality to exist _outside_ of itwêwina (the dictionary database).

aarppe:
Yes, the intention is that determinations that 1) various dictionary entries from multiple Cree lexical sources, with the same head word/phrase, concern the same lexical item, and that 2) the (English) definitions are similar enough that only one needs only to present one definition, should be implemented in the dictionary database model.

aarppe:
It is the dictionary database model where we should also record alternative (allowable) and dialectal spellings. Arok's database actually already contains some, and of course the various orthographical conventions represented in MD and AECD can be retained in the dictionary database model as well.

PR #631, opened 2021-01-25
Skip duplicates when loading dictionary

andrewdotn:
Currently searches for ‘awa’ and ‘nîminâniwan’ return a lot of duplicates where duplicate wordforms with the same analysis have been loaded into the database by the import process.

As a first step for improving search, we simply track which wordforms have already been inserted and avoid inserting duplicates.

before-after


CreeDictionary/DatabaseManager/xml_importer.py
```diff
@@ -243,6 +251,13 @@ def generate_english_keywords(
         fst_lemma_text, word_class = lemma_text_and_word_class
         generated_pos = word_class.pos
 
+        p_entry = ProcessedEntry(
+            stem=entry.stem, ic=entry.ic, lemma_analysis=lemma_analysis
+        )
+        if p_entry in seen_lemmas:
+            continue
+        seen_lemmas.add(p_entry)
```
andrewdotn, line 259: This is the key bit

CreeDictionary/tests/DatabaseManager_tests/xml_importer_test.py
```diff
@@ -57,19 +57,6 @@ def test_import_xml_common_analysis_definition_merge(shared_datadir):
...
 
-@pytest.mark.django_db
-def test_import_xml_crkeng_small_duplicate_l_pos_ic(shared_datadir):
-    # This test shows the behavior of the importer when entries with duplicate (l, pos, ic) in the xml file exists
-    # It's a rare case
-
-    # two Wordform objects will be created, each with the pooled translations from the two entries
-    migrate_and_import(shared_datadir / "crkeng-small-duplicate-l-pos-ic")
-
-    lemmas = Wordform.objects.filter(text="asawâpiwin", is_lemma=True)
```
eddieantonio, line 68: @aarppe this test is a left over from old days where the dictionary content would have duplicate entries with _sometimes_ different definitions. This caused some entries to have duplicates here. In addition `is_lemma` used to refer to _all dictionary heads_ — not lemmas, due to a misunderstanding in terminology 😬

PR #634, opened 2021-01-27
Turn off mypy ignore_missing_imports

andrewdotn:
Python doesn’t always require the file `__init__.py`, but without it, mypy
doesn’t know how to trace imports even between files in the same directory,
so all imports of types turn into `Any`, and `ignore_missing_imports` hides
that anything is amiss.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,7 +1,7 @@
 [mypy]
 
 [mypy-API.migrations.*]
-# Auto-generated code
+# Skip checking Django-auto-generated migration code
 ignore_errors = True
 
 # External packages that we haven’t set up types for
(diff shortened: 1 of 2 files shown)


mypy.ini
```diff
@@ -1,5 +1,44 @@
...
+ignore_missing_imports = True
+
+[mypy-dawg.*]
+ignore_missing_imports = True
+
+[mypy-debug_toolbar.*]
+ignore_missing_imports = True
+
+[mypy-django.*]
+# TODO: try https://pypi.org/project/django-stubs/
```
eddieantonio, line 25: Could you try it in this PR?

mypy.ini
```diff
@@ -1,5 +1,44 @@
 [mypy]
-ignore_missing_imports = True
 
 [mypy-API.migrations.*]
+# Auto-generated code
```
eddieantonio, line 4: This comment confuses me, because I usually write a comment like `# AUTO-GENERATED` in code that my scripts generate. I acknowledge that this is mostly a me problem! :joy:

How about...

```suggestion
# Do not type-check Django migrations (automatically created by Django)
```

mypy.ini
andrewdotn: If you have more patience that I do, here’s a start: https://gist.github.com/andrewdotn/902b8062bf35bd3a3eb1cbb530500392

  - The `NewType`-casts, e.g., `FSTTag(foo)` do nothing at runtime if passed a completely different type
  - Not sure how to annotate a model `CharField` as yielding a `NewType()`-created class
  - Building up a `TypedDict` starting from `{}` really upsets mypy, and I would prefer not to have to drag in whatever is the mypy equivalent of TypeScript’s `Partial<>`

PR #635, opened 2021-02-02
Introduce search_quality

andrewdotn:
- There is a list of sample queries
  - The sample queries can be run through the current search algorithm
    using a management command or a web form
  - A summary of results is displayed on a web page
  - To compare and track performance over time, serialized search results
    from previous versions can be checked in as `.json.gz` files, currently
    about 1.3MB in size
  - I’m not checking in such a results file just yet; someone who knows
    more Cree should look over my `sample.csv` first

The newer mypy version supports `int | str` notation with
`__future__.annotations`, and its `dmypy` daemon doesn’t crash as much on
my machine. Pipenv upgraded some other packages too when upgrading mypy.


CreeDictionary/search_quality/sample.csv
```diff
@@ -0,0 +1,64 @@
...
+car,pôsîw,pôsiy,,,,,
+rabbit,wâpos,,,,,,
+water,nipiy,,,,,,
+telling a story,âcimêw,,,,,,
+on water,nipîhk,,,,,,
+hello,tânisi,,,,,,
+difficult thing,âyiman,,,,,,
+many snakes,kinêpikoskâw,,,,,,
+dances,nîmiw,,,,,,
+i see you,kinêpikoskâw,,,,,,
```
eddieantonio, line 58:
```suggestion
i see you,kiwâpamitin,,,,,,
```

_kinêpikoskâw_, or _kinêpik_ ("snake) + _-iskâ_ ("an abundance of") means THERE ARE MANY SNAKES.

eddieantonio (commented):
It's going to take me a while to review this, but in the meantime, I'm gonna rope in the linguists to review `sample.csv`!

CreeDictionary/search_quality/sample.csv
```diff
@@ -0,0 +1,64 @@
...
+he sees him,wâpamêw,,,,,,
+she sees him,wâpamêw,,,,,,
+starblanket,atâhkakohp,acâhkosa kâ-otakohpit,,,,,
+star blanket,atâhkakohp,acâhkosa kâ-otakohpit,,,,,
+being taught,kiskinwahamâkosiw,,,,,,
+they see us,niwâpamikonânak,kiwâpamikonawak,,,,,
```
eddieantonio, line 64: @aarppe @katieschmirler @atticusha Please review this list and make suggestions!

@andrewdotn: is there a documented format for this file? Also, do you know how to get GitHub to render a pretty-printed version of this. I know it's capable!

andrewdotn: Click the `CreeDictionary/search_quality/sample.csv` label in the conversation view to go to https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/pull/635/files/c35622e5cdae3b6ca0a17ccf99dd1257589ee7fd#diff-1048d6e8db6891c0b02807610cba69b649b00ecd35517d5b8629211c8570fd1a

Then click the “…” menu → View file to go to https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/c35622e5cdae3b6ca0a17ccf99dd1257589ee7fd/CreeDictionary/search_quality/sample.csv

CreeDictionary/search_quality/sample.csv
andrewdotn: Oops, yes, you tried to tell me about this before and I didn’t catch on to what you were referring to. But I do see it and the copy-paste problem now!

CreeDictionary/search_quality/urls.py
```diff
@@ -0,0 +1,15 @@
+from django.urls import path, re_path
+
+from . import views
+
+app_name = "search_quality"
+urlpatterns = [
+    path("", views.IndexView.as_view(), name="index"),
+    re_path(
+        # the strict regex is to prevent directory traversal attacks
```
eddieantonio, line 9: 👍🏼

CreeDictionary/search_quality/views.py
```diff
@@ -0,0 +1,116 @@
...
+        ]:
+            context[f"average_{k}"] = mean(r[k] for r in analysis.values())
+
+        return context
+
+
+@staff_member_required
+def run(request):
+    if request.method != "POST":
+        raise Http404()
```
eddieantonio, line 87: Pedantic suggestion: this should respond with status [`405 Method Not Allowed`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405).

CreeDictionary/search_quality/views.py
```diff
@@ -0,0 +1,116 @@
...
+            "search_quality:results-analysis",
+            kwargs={"path": out_file.name},
+        )
+
+        yield f'</pre><a href="{conditional_escape(done_url)}">Done!</a>'
+
+    # Sadly, although this old trick for sending out a live-updating web page
+    # works with Firefox and Chrome, it doesn’t seem to work with Safari. And it
+    # doesn’t seem worth it to set up websockets or a ‘progress_message’
+    # database table just for this.
```
eddieantonio, line 115: FAIR! 👍🏼 

Also, TIL about `StreamingHttpResponse`!

CreeDictionary/search_quality/views.py
```diff
@@ -0,0 +1,116 @@
...
+
+        for status in gen_run_sample(out_file=out_file):
+            yield conditional_escape(status) + "\n"
+
+        done_url = reverse(
+            "search_quality:results-analysis",
+            kwargs={"path": out_file.name},
+        )
+
+        yield f'</pre><a href="{conditional_escape(done_url)}">Done!</a>'
```
eddieantonio, line 110:
```suggestion
        yield f'</pre><a href="{conditional_escape(done_url)}">View results!</a>'
```

CreeDictionary/search_quality/__init__.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
+
```
eddieantonio, line 2: Since mypy was updated to 0.800, this is no longer needed!

```suggestion
```

CreeDictionary/search_quality/sample.py
```diff
@@ -0,0 +1,62 @@
...
+        start_time = time.time()
+        results = [r.serialize() for r in Wordform.search_with_affixes(query)]
+        time_taken = time.time() - start_time
+
+        combined_results[query] = {
+            "time_taken_seconds": time_taken,
+            "results": results,
+        }
+
+        count = sum(1 if r["is_lemma"] else 2 for r in results)
```
eddieantonio, line 56: Is this the same as `analyze_results.count_results()`?

CreeDictionary/search_quality/analyze_results.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+Analysis = dict[str, Optional[SingleQueryAnalysis]]
+
+
+def count_results(results: list[SerializedSearchResult]):
+    return sum(1 if r["is_lemma"] else 2 for r in results)
+
+
+class OptionalPartialSerializedWordform(TypedDict, total=False):
```
eddieantonio, line 47: Why both `Optional` and `Partial`? Wouldn't it just be a partial (non-total) `SerializedWordform`?

CreeDictionary/search_quality/analyze_results.py
```diff
@@ -0,0 +1,191 @@
...
+    if not word.strip():
+        return ""  # n/a
+
+    index = 1
+    for r in results:
+        # We also have the definition here, could include it
+        if word == r["matched_cree"]:
+            return index
+        index += 1 if r["is_lemma"] else 2
+    return "☹"
```
eddieantonio, line 119: Haha!

CreeDictionary/search_quality/analyze_results.py
```diff
@@ -0,0 +1,191 @@
...
+                "unique_duplicate_count": dupe_info["unique"],
+            }
+
+            # How many top expected results does the sample specify for this
+            # query?
+            specified_result_count = 0
+            result_found_count = 0
+            threeten_good_result_count = 0
+
+            for i in [i + 1 for i in range(3)]:
```
eddieantonio, line 166: Better to explicitly specify which indices?

```suggestion
            for i in (1, 2, 3):
```

CreeDictionary/search_quality/analyze_results.py
```diff
@@ -0,0 +1,191 @@
...
+                # mypy does not like dynamic keys
+                word = cast(dict, row)[f"Nêhiyawêwin_{i}"]
+
+                if word:
+                    cast(dict, query_analysis)[f"word_{i}"] = word
+                    rank = find_rank(word, result_list)
+                    cast(dict, query_analysis)[f"word_{i}_rank"] = rank
+
+                    specified_result_count += 1
+                    if type(rank) == int:
```
eddieantonio, line 176: I think this is the more common way to check the type:

```suggestion
                    if isinstance(rank, int):
```

CreeDictionary/search_quality/analyze_results.py
```diff
@@ -0,0 +1,191 @@
...
+            # query?
+            specified_result_count = 0
+            result_found_count = 0
+            threeten_good_result_count = 0
+
+            for i in [i + 1 for i in range(3)]:
+                # mypy does not like dynamic keys
+                word = cast(dict, row)[f"Nêhiyawêwin_{i}"]
+
+                if word:
```
eddieantonio, line 170: If you feel like making this neater:

```suggestion
                if not word:
                    break  # there are e no more Nêhiyawêwin_x words
```

eddieantonio (approved):
LGTM! Just a few, very, very pedantic notes!

CreeDictionary/search_quality/__init__.py
andrewdotn: `CreeDictionary/search_quality/analyze_results.py:31: error: X | Y syntax for unions requires Python 3.10`

eddieantonio: Do you need it in both `__init__.py` _and_ `analyze_results.py`?

CreeDictionary/search_quality/sample.csv
andrewdotn: When *everyone* asked what the column headings meant, I took that as a signal that they were poorly named. I have renamed the first column to ‘Query’, and have added a `README.md` file that explains them in more detail.

CreeDictionary/search_quality/views.py
andrewdotn: I went with `@require_http_methods(["POST"])`, which does sent that status. I was a little disappointed though that it sends blank pages to the client, instead of at least a little “405 Method not allowed” in the body.

CreeDictionary/search_quality/__init__.py
andrewdotn: Nope, just in `analyze_results.py`.

CreeDictionary/search_quality/sample.py
andrewdotn: It was. Had to introduce a new file to break a circular import to reuse it, but it’s fixed.

CreeDictionary/search_quality/analyze_results.py
andrewdotn: Good point.

CreeDictionary/search_quality/analyze_results.py
andrewdotn: Yup, thanks

CreeDictionary/search_quality/analyze_results.py
andrewdotn: I was thinking, what if one day there are *more* columns, but your code is much clearer. If there are many more columns some day, https://stackoverflow.com/a/29596068/14558 suggests creating an ‘inclusive_range` method.

CreeDictionary/search_quality/analyze_results.py
andrewdotn: Thanks for pointing this out. There was actually a pesky but subtle indentation-related bug with the previous use of `if word: \n <long block of code>` in this file.

CreeDictionary/search_quality/sample.py
```diff
@@ -1,16 +1,9 @@
...
-from os import PathLike
 from typing import Iterable, TypedDict
 
-from API.models import Wordform
-from . import SampleSearchResultsJson, DEFAULT_SAMPLE_FILE
-
 
 class SampleCsvRow(TypedDict):
-    English: str
+    Query: str
```
eddieantonio, line 6: This makes a lot more sense! 😄

aarppe:
This sample list should be supplemented from the more structured list (of the 100 most frequent) of English words, with the top 2-5 Cree matches indicated.

Since that list has many function words (prepositions, pronouns, conjunctions, adverbs), we might want to supplement it with the next 100-200 words, to get more verbs and nouns.

In any event, even the list with 100 frequent English words and their expected Cree matches should gives us a more reliable quantitative view of how well the English search works.

PR #640, opened 2021-02-04
Add docker-compose config

andrewdotn:


CreeDictionary/CreeDictionary/settings.py
```diff
@@ -59,6 +59,7 @@
     "API.apps.APIConfig",
     "CreeDictionary.apps.CreeDictionaryConfig",
     "morphodict.apps.MorphodictConfig",
+    "DatabaseManager",
```
eddieantonio, line 62: I love splitting this up into apps!

CreeDictionary/setup.py
```diff
@@ -1,11 +0,0 @@
-from setuptools import find_packages, setup
```
eddieantonio, line 1: THANK.
YOU.

docker/Dockerfile
```diff
@@ -46,25 +50,39 @@ RUN set -ex \
...
 
-# Setup Python deps
-ADD Pipfile /app/Pipfile
-ADD Pipfile.lock /app/Pipfile.lock
-
 WORKDIR /app/
 
+# Setup Python deps
+ADD --chown=${WSGI_USER} Pipfile .
+ADD --chown=${WSGI_USER} Pipfile.lock .
```
eddieantonio, line 57: Ooh! Much cleaner than doing this in a `RUN` command!

docker/Dockerfile
```diff
@@ -46,25 +50,39 @@ RUN set -ex \
...
+
 RUN PIPENV_VENV_IN_PROJECT=1 pipenv install --deploy
 
+ADD --chown=${WSGI_USER} package.json .
+ADD --chown=${WSGI_USER} package-lock.json .
+# cypress is a big, heavy package that takes a long time to install and
+# that we absolutely don’t need in production
+RUN grep -Ev '"@?cypress' package.json | sponge package.json
```
eddieantonio, line 65: Excellent!

docker/README.md
```diff
@@ -0,0 +1,27 @@
+# Docker for itwêwina
+
+The current setup was created with the intent of using it in production,
+not for development.
+
+  - Run `sqlite3 db.sqlite3` to create an empty database file
+
+    Otherwise `docker-compose` will create a *directory* instead
```
eddieantonio, line 8: wat

docker/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+  - Populate the test dictionary with
+
+        ./docker-django-manage.py xmlimport import \
+            ../CreeDictionary/res/test_dictionaries/crkeng.xml
+
+## Future work
+
+  - smaller docker image without dev dependencies through multi-stage
+    builds
```
eddieantonio, line 27: 👍🏼

eddieantonio (approved):
This is some great stuff! 👍🏼

docker/README.md
andrewdotn: When docker is mounting host paths, those paths can be either directories or files. Docker will create the path if it does not exist. Creating an empty directory makes a bit more sense than creating a zero-byte file. If docker were being designed anew, a trailing slash could be required to indicate a directory, but there are already a ton of things out there assuming that specifying a non-existent path without a trailing slash will result in directory creation.

PR #641, opened 2021-02-05
Fix paradigm audio

nienna73:
Fix #636 . Instead of replacing entire word details page when pressing "load more", just replace the paradigm section.

PR #646, opened 2021-02-08
change: use .js-* classes to select things in the paradigm

eddieantonio:
An improvement to #641. I felt uneasy using the existing (unweildly!) [BEM](http://getbem.com/) class names, and chose instead to use the `js-*` prefix for the class names to use in JavaScript.

I also refactored and cleaned up variable names because I can't help myself ¯\\\_(ツ)\_/¯

### DIFF
--- a/CreeDictionary/CreeDictionary/templates/CreeDictionary/components/paradigm.html
+++ b/CreeDictionary/CreeDictionary/templates/CreeDictionary/components/paradigm.html
@@ -17,6 +17,8 @@
     JavaScript hooks:
      - .js-replaceable-paradigm: encapsulates the ENTIRE paradgim so that
        JavaScript can replace the contents with a different paradigm.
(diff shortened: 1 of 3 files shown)


CreeDictionary/CreeDictionary/templates/CreeDictionary/components/paradigm.html
```diff
@@ -14,11 +14,19 @@
         | You  | kinipân | ê-nipâyan |
         | [       + show more      ] |
 
+    JavaScript hooks:
+     - .js-replaceable-paradigm: encapsulates the ENTIRE paradgim so that
```
andrewdotn, line 18:
```suggestion
     - .js-replaceable-paradigm: encapsulates the ENTIRE paradigm so that
```

andrewdotn (approved):
I like the `.js-blah` convention; that makes it clearer how safe things are to rename.

PR #652, opened 2021-02-12
fix/refactor: metatags

eddieantonio:
> _Part 87 in the ongoing saga:_ Eddie is Disappointed in Django's Templating System

So, a few things:
 - I extracted the metatag in its own `{% include %}` component
 - the image URLs need to be **absolute** URIs, so, with _great effort_ I made them as such
 - and a sneaky update to the footer font colours; let me know if I should cherry-pick that into a different PR

### DIFF
--- a/CreeDictionary/CreeDictionary/templates/CreeDictionary/base.html
+++ b/CreeDictionary/CreeDictionary/templates/CreeDictionary/base.html
@@ -3,26 +3,10 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>itwêwina: the online Cree dictionary</title>
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="theme-color" content="#A20010">
   {% block metatags %}
-  <!-- Primary Meta Tags -->
-  <meta name="description" content="A smart, online dictionary of Plains Cree (Y-Dialect) | nêhiyawi-masinahikan kâ-iyinîsit">
-
-  <!-- Open Graph / Facebook -->
-  <meta property="og:type" content="website">
-  <meta property="og:url" content="https://itwewina.altlab.app/">
-  <meta property="og:title" content="itwêwina: the Plains Cree dictionary">
-  <meta property="og:description" content="A smart, online dictionary of Plains Cree (Y-Dialect) | nêhiyawi-masinahikan kâ-iyinîsit">
-  <meta property="og:image" content="{% static 'CreeDictionary/images/itwewina-social.svg' %}">
-
-  <!-- Twitter -->
-  <meta property="twitter:card" content="summary_large_image">
-  <meta property="twitter:url" content="https://itwewina.altlab.app/">
-  <meta property="twitter:title" content="itwêwina: the Plains Cree dictionary">
-  <meta property="twitter:description" content="A smart, online dictionary of Plains Cree (Y-Dialect) | nêhiyawi-masinahikan kâ-iyinîsit">
-  <meta property="twitter:image" content="{% static 'CreeDictionary/images/itwewina-social.svg' %}">
+  {% include 'CreeDictionary/components/metatags.html' %}
   {% endblock %}
   <link rel="icon" type="image/png" href="{% static 'CreeDictionary/images/itwewina-32.png' %}">
   <link rel="alternate icon" type="image/svg+xml" href="{% static 'CreeDictionary/images/itwewina.svg' %}">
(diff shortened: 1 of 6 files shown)


CreeDictionary/CreeDictionary/templates/CreeDictionary/components/metatags.html
```diff
@@ -0,0 +1,35 @@
...
+    Parameters:
+     - title — the title of the webpage -- as this goes in the tab bar, it's
+       the link in search results, and is visible in social media links is the
+       MOST important part!
+     - description — a short summary of the current page
+  {% endcomment %}
+
+  {% load url_extras %}
+
+  {% with title_="itwêwina: the online Cree dictionary" desc_="A smart, online dictionary of Plains Cree (Y-Dialect) | nêhiyawi-masinahikan kâ-iyinîsit" %}
```
eddieantonio, line 14: @aarppe please suggest here what the defaults should be!

eddieantonio:
🤦🏼  https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image#twitter-widget-0:~:text=SVG%20is%20not%20supported.

CreeDictionary/CreeDictionary/templatetags/url_extras.py
```diff
@@ -0,0 +1,31 @@
...
+
+register = template.Library()
+
+
+class AbsoluteStaticNode(StaticNode):
+    def url(self, context) -> str:
+        path = super().url(context)
+        if not path.startswith("/"):
+            # Assume it's already an absolute URI:
+            return path
```
andrewdotn, line 18: This makes my head hurt to think about. You have tests, so it’s probably ok? 🤷 

In trying to untangle this, to my great and shocking dismay I just learned from [Wikipedia](https://en.wikipedia.org/wiki/URL#Syntax) that the domain name *is* optional and `` does work :O

Given the enormous potential confusion between absolute *paths* which start with a `/` and absolute *URLs* which do not, can we:

 1. Have a block comment defining the terms “absolute path” and “absolute URL,” possibly quoting RFC1808 “Relative Uniform Resource Locators”:

    > Absolute URLs contain a great deal of information which may already be known from the context of the base document's retrieval, including the scheme, network location, and parts of the URL path.

    I know it says RFC 3986 obsoletes 1808, but I don’t think it does so meaningfully, in that the newer, supposedly-better RFC is so abstract that it *doesn’t even contain the term ‘absolute URL’*

  2. Any mention of ‘absolute’ in the code/comments must include a specifier about which absolute it’s referring to, i.e., ‘absolute URL’ or ‘absolute path’

CreeDictionary/CreeDictionary/templates/CreeDictionary/components/metatags.html
nienna73:
```suggestion
  {% with title_="itwêwina: the online Cree dictionary" desc_="A smart, online dictionary of Plains Cree (Y-Dialect) | nêhiyawi-masinahikan kâ-iyiniwâk" %}
```

Or I think Antti also suggested "nêhiyaw-itwêwina ayapîhk"

CreeDictionary/CreeDictionary/templates/CreeDictionary/base.html
```diff
@@ -3,26 +3,10 @@
...
-  <meta property="og:description" content="A smart, online dictionary of Plains Cree (Y-Dialect) | nêhiyawi-masinahikan kâ-iyinîsit">
-  <meta property="og:image" content="{% static 'CreeDictionary/images/itwewina-social.svg' %}">
-
-  <!-- Twitter -->
-  <meta property="twitter:card" content="summary_large_image">
-  <meta property="twitter:url" content="https://itwewina.altlab.app/">
-  <meta property="twitter:title" content="itwêwina: the Plains Cree dictionary">
-  <meta property="twitter:description" content="A smart, online dictionary of Plains Cree (Y-Dialect) | nêhiyawi-masinahikan kâ-iyinîsit">
-  <meta property="twitter:image" content="{% static 'CreeDictionary/images/itwewina-social.svg' %}">
+  {% include 'CreeDictionary/components/metatags.html' %}
```
nienna73, line 9: Yay for making this file cleaner!

nienna73 (changes requested):
I made a suggestion about the Cree description of the site that should be reviewed before merging the changes. Otherwise it looks good. I don't think you need to cherry-pick the footer change.

PR #655, opened 2021-02-13
change: use recommended Google Fonts inclusion line

eddieantonio:
It's what's recommended to you when you select the fonts on [Google Fonts](https://fonts.google.com/specimen/Open+Sans?query=open+&preview.text_type=custom&sidebar.open=true&selection.family=Open+Sans:ital,wght@0,400;0,700;1,400).

It includes:
 - `rel=preconnect` to <https://fonts.gstatic.com>, because that is where the fonts proper are downloaded; the API request is to a different origin.
 - `font-display: swap` so [rendering is not blocked while the page is downloading the fonts](https://css-tricks.com/font-display-masses/#most-of-the-time-youll-use-swap)


dwhieb (approved):
I don't have itwêwina building and testing locally yet so I didn't try this out locally, but this looks good to me. I think `font-display: swap` makes sense for us.

PR #656, opened 2021-02-13
change: improve orthography selector accessibility

eddieantonio:
It's bugged me for a while that the "SRO/syllabics" buttons (orthograpgy selectors) buttons were too small on mobile, so I made the touch target much larger.


nienna73 (approved):
I had to look up what a 'ch' unit is in CSS. Looks good!

PR #657, opened 2021-02-13
feat: sitemaps

### LINKED ISSUE #365: Dynamically generate a sitemap.xml with ALL THE WORDFORMS!
> **User story**: As a Cree language learner, when I search for a Cree word in my search engine of choice, it should link to a page on itwêwina.

How do we let search engines like Google, Bing, Duck Duck Go, etc. know that we're a Cree dictionary? By having them crawl our site! But since web crawlers need _links_ to crawl, we need to provide links to all of our _stored_ wordforms. The solution: `sitemap.xml`.

The sitemap.xml should list both all general pages (e.g., `about`, `contact-us`), but also every `/word/XXX` page.

@aarppe, I know you don't want Google to just show the definition on their search page. You can disable this by using the [`nosnippet` directive](https://support.google.com/webmasters/answer/35624#nosnippet).

We'll probably want to provide a `robots.txt` as well with a link to the `sitemap.xml`.

eddieantonio:
Uses Django's [sitemap framework](https://docs.djangoproject.com/en/2.2/ref/contrib/sitemaps/#sitemap-for-static-views) to create `sitemap.xml`.

Fixes #365


eddieantonio:
@andrewdotn the implementation is ready and works (lacking tests though :/). I would _like_ a better way to deal the the test database (e.g., git LFS or just not committing the database to the repository!), but alas, that does not exist. If you're bored, feel free to review what's here.

CreeDictionary/tests/CreeDictionary_tests/test_sitemap.py
```diff
@@ -0,0 +1,50 @@
...
+    #     <loc>http://itwewina.altlab.app/</loc>
+    #   </url>
+    #
+    #   <!-- head word forms -->
+    #   <url>
+    #     <loc>http://itwewina.altlab.app/word/atim/</loc>
+    #   </url>
+    # </urlset>
+    ns = {"sitemap": "http://www.sitemaps.org/schemas/sitemap/0.9"}
+    locations = [loc.text for loc in root.findall("./sitemap:url/sitemap:loc", ns)]
```
nienna73, line 40: I'm having a hard time following what's happening here

eddieantonio: I rewrote it. I always make the same mistake when writing XPath queries with the ElementTree library 😩

andrewdotn (approved):

CreeDictionary/tests/CreeDictionary_tests/test_sitemap.py
nienna73: Thanks! It makes more sense to me now.

nienna73 (approved):

PR #662, opened 2021-02-19
build: refinements

eddieantonio:
Getting ready for some Dockerfile reworking. Starting out with some small things first.

 - change cypress `baseUrl` to `localhost` so that it's an `ALLOWED_HOST` in `settings.py`
 - change `package.json` dependencies such that dependencies required ONLY FOR TESTING (\*cough\* Cypress \*cough\* \*cough\*) can be omitted by using `npm install --only=production`

**EDIT**:  This is the start of a [stacked pull request](https://www.michaelagreiler.com/stacked-pull-requests/). I will continue making PRs based off of THIS PR.

PR #663, opened 2021-02-19
build: multi-stage Docker build

eddieantonio:
I've rearranged the Dockerfile such that it's a multi-stage build now. The total size of the application image is about ~248 MiB. 

**EDIT**: whereas the builder image is **~657MB**! 🤯 

Also, the Cypress tests can run on the `docker-compose.dev.yml` arrangement!

This PR builds off of #662.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -24,7 +24,7 @@ ARG NODE_VERSION=v12.20.2
 # (note: default npm is too old for the apt install'd version of NodeJS??!?!?!)
 WORKDIR /tmp
 ADD https://nodejs.org/dist/${NODE_VERSION}/node-${NODE_VERSION}-linux-x64.tar.gz /tmp/node.tar.gz
-RUN tar xf node.tar.gz\
+RUN tar xf node.tar.gz \
  && mv node-${NODE_VERSION}-linux-x64 /opt/node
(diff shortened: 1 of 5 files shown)


docker/Dockerfile
```diff
@@ -1,102 +1,90 @@
...
-ARG EXPOSE_PORT=8000
+ARG NODE_VERSION=v12.20.2
+
+# Install Node + npm
+# (note: default npm is too old for the apt install'd version of NodeJS??!?!?!)
+WORKDIR /tmp
+ADD https://nodejs.org/dist/${NODE_VERSION}/node-${NODE_VERSION}-linux-x64.tar.gz /tmp/node.tar.gz
+RUN tar xf node.tar.gz \
+ && mv node-${NODE_VERSION}-linux-x64 /opt/node
+ENV PATH="/opt/node/bin:${PATH}"
```
andrewdotn, line 29: Oh cool, I did not realize you could reference `$PATH` when resetting it in a Dockerfile and was embedding the default …

andrewdotn (approved):

docker/Dockerfile
eddieantonio: I learned this too today. I _considered_ copying everything into `/usr/local` but this seemed much safer!

PR #666, opened 2021-02-20
update deployment step of test-and-deploy workflow

### LINKED ISSUE #665: update deploy step of test-and-deploy action
The `deploy-to-sapir` step in the `test-and-deploy` action should be updated to use the webhook in the deployment app, triggering deployment.

dwhieb:
closes #665

The `test-and-deploy` workflow now sends a POST request to the deployment webhook (https://deploy.altlab.dev/itwewina), which in turn triggers deployment of itwêwina.

The POST request is currently sent using `curl`. If the deployment server returns a 5xx status, the workflow will fail silently (using the `--fail` option). This could be improved by using a script rather than `curl`, so that the workflow fails on 4xx errors as well. (There's a new `--fail-with-body` option that would exit with an error code on 4xx errors, but our server isn't running a high enough version of `curl`.)

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -158,6 +158,6 @@ jobs:
     steps:
     - name: send HTTP request to deploy.altlab.dev webhook
       run: >-
-        curl -X POST https://deploy.altlab.dev/korp-frontend-dev -d '{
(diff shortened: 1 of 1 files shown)


eddieantonio (approved):
LGTM! 🎉

PR #669, opened 2021-02-22
Add google site verification

andrewdotn:
I believe that how this will work is:
  - The included code will let me see details of the production site in the search console.
  - I will then be able to share that access with others.
  - The code is not secret because it is often included in meta tags.

If/when we get DNS-based site verification, this can be removed.


eddieantonio (approved):
LGTM!

PR #670, opened 2021-02-22
Don’t check in test db.sqlite3

andrewdotn:
Instead, just before the first DB-using test runs, if the database has no wordforms, build the test DB from XML.

This takes about 30s which will add to CI run time, but at least then we’ll know that building the test DB works.

### DIFF
--- a/CreeDictionary/conftest.py
+++ b/CreeDictionary/conftest.py
@@ -1,0 +1,37 @@
+import pytest
+from django.conf import settings
+
+# See “`conftest.py`: sharing fixtures across multiple files”
+# https://docs.pytest.org/en/stable/fixture.html#conftest-py-sharing-fixtures-across-multiple-files
+
+
+@pytest.fixture(scope="module")
+def django_db_setup():
+    """
+    Normally pytest-django creates a new, empty in-memory database and runs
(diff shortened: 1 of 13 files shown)


andrewdotn:
Am thinking that the python test failure is from a cache that needs to be reset after populating wordforms.

CreeDictionary/conftest.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # If this environment variable is set, a conditional in settings.py
+    # should have pointed the database at CreeDictionary/test_db.sqlite3
+    assert settings.USE_TEST_DB
+
+    # We’d like the output from the DB setup to go directly to stdout regardless
+    # of any capture settings. This is a workaround for “ScopeMismatch: You
+    # tried to access the 'function' scoped fixture 'capsys' with a 'session'
+    # scoped request object”:
+    # https://github.com/pytest-dev/pytest/issues/2704#issuecomment-603387680
```
eddieantonio, line 51: 👍🏼

eddieantonio (approved):
This is awesome! Thank you so much for doing it! It's been a TODO for quite a while!

PR #672, opened 2021-02-25
change(css): adjust link colours

eddieantonio:
@andrewdotn and played with the colours, using [Adobe Color](https://color.adobe.com/en/create/color-wheel) to pick less arbitrary shades, based on the header color.

This is what it looks like:

Screen Shot 2021-02-25 at 10 00 59 AM

### DIFF
--- a/src/css/variables.css
+++ b/src/css/variables.css
@@ -36,9 +36,10 @@
   --header-stripe-color:    var(--app-title-color); /* Color of the "stripe" to the left of the title. */
   --link-action-color:      #A568DB; /* A link that implies an immediate action. */
   --link-active-color:      #0E6790; /* Currently active link. */
-  --link-color:             #0F76B5; /* Regular links */
+  /* Triad based on #95211E */
+  --link-color:             #1685DE; /* Regular links */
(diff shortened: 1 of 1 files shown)


dwhieb (approved):
I like it! The blue is only a AA rating for color contrast according to WCAG, but since the links are always bold I think that's fine. If we use this color for normal text we might want to consider increasing the contrast ratio with white though.

eddieantonio:
> I like it! The blue is only a AA rating for color contrast according to WCAG, but since the links are always bold I think that's fine. If we use this color for normal text we might want to consider increasing the contrast ratio with white though.
> 
> https://image-color.com/color-contrast.html#FFFFFF-1685DE

Thank you for checking the WCAG colour contrast!

aarppe:
I don't know if it's real or just my eyes, but I find the new alternative link color, which Finns would characterize as that of the berry porridge, better than the earlier pink'ish.

Nevertheless, I'd be interested in seeing at some later point what the turquoise color scheme would look like - colder colors for the summer?

dwhieb:
I wanna try berry porridge now!

PR #673, opened 2021-03-01
change: better names for docker compose

eddieantonio:
So our Docker image  for itwewina showed up in logs as "docker_app" and the container as "docker_app_1". This is bad! So I specified some explicit names.

Docs:
 - https://docs.docker.com/compose/compose-file/compose-file-v3/#build:~:text=If%20you%20specify%20image%20as%20well,and%20optional%20tag%20specified%20in%20image%3A
 - https://docs.docker.com/compose/compose-file/compose-file-v3/#container_name

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -4,6 +4,8 @@ services:
     build:
       context: ..
       dockerfile: docker/Dockerfile
+    image: itwewina
+    container_name: itwewina
     volumes:
       - "./db.sqlite3:/app/CreeDictionary/db.sqlite3"
       - "./.env:/app/.env"
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,11 +1,9 @@
 version: "3"
 services:
-  app:
+  itwewina:
     build:
       context: ..
       dockerfile: docker/Dockerfile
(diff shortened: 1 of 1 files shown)


andrewdotn:
If you change `app` on line 3 to `itwewina`, I think that’ll fix the auto-generated naming issues.

The default is `${project_name}_${service}_${n}` where:
  - [project comes from `docker-compose -p` or `$COMPOSE_PROJECT_NAME`](https://docs.docker.com/compose/reference/overview/#use--p-to-specify-a-project-name) but defaults to the name of the directory containing the `docker-compose.yml` file
  -  service name is the YAML key, and
  -  `n` is a number to distinguish different instances when scaling

This lets you run kititwêwininawa on the same host for different projects w/o collision.

eddieantonio:
> * [project comes from `docker-compose -p` or `$COMPOSE_PROJECT_NAME`](https://docs.docker.com/compose/reference/overview/#use--p-to-specify-a-project-name) but defaults to the name of the directory containing the `docker-compose.yml` file

@nienna73 👀 👀 👀 

Haha, I was trying to look for this option earlier today!

andrewdotn (approved):

PR #681, opened 2021-03-05
change: split docker-compose files for production and staging

eddieantonio:
Using multiple docker-compose files to make suitable environments for both development and production:

See: https://docs.docker.com/compose/extends/#multiple-compose-files


andrewdotn (commented):
I’d actually recommend not making this change unless someone is going to be actively using docker for development and will be maintaining it.

  - For development, normally `/app` itself would be a mount so that changes in an editor/IDE on the host would show up inside the container
  - Doing that would suggest other changes to the Dockerfile, so that we might end up with several of them
  - UIDs will be an issue, especially if we’d want to support development on both macOS and linux docker hosts

I do of course want to make sure that your intended use case is supported, so please let me know if I am missing something here that will make the complexity worth it.

eddieantonio:
> I do of course want to make sure that your intended use case is supported, so please let me know if I am missing something here that will make the complexity worth it.

My intended use case was making it easier to see what changes in the `docker-compose` deployment do before trying it in production (my thoughts were on switching to uswsgi). So... somewhat like staging? Not quite development though.

eddieantonio:
> I’d actually recommend not making this change unless someone is going to be actively using docker for development and will be maintaining it.

I'm torn on this because I don't want to increase the maintance footprint of the project, but I also want to easily experiment with Docker locally without trying it on production 🙃 

What would be a good way to do this _without_ increasing our maintannce tasks?

andrewdotn:
Oh, ok, yup, as a staging docker-compose deployment this totally makes a lot of sense: an easier way to run the same container as in prod. Dev docker stuff is typically fairly different because it’ll include all the dev packages and such.

Note that for the port stuff, it may be easier to set [`CYPRESS_BASE_URL=http://localhost:8001`](https://docs.cypress.io/guides/guides/environment-variables.html)

eddieantonio:
@andrewdotn I changed the word "development" to "staging" and made a few other changes. Let me know if it looks good! I want to experiement with, e.g., uwsgi on my machine before pushing it to production! 😬

andrewdotn (approved):

PR #685, opened 2021-03-11
feat: add dictionary sources as a data migration

eddieantonio:
Inserts bibliographic information as a data migration. Why a data migration?

 - so it automatically appears when creating a test database
 - it appears in the production site the next time `migrate` is run


CreeDictionary/DatabaseManager/xml_importer.py
```diff
@@ -115,15 +115,6 @@ def find_latest_xml_file(dir_name: Path) -> Path:
     )
 
 
-def import_sources():
```
eddieantonio, line 118: This was never called anywhere, which explains _a lot_.

andrewdotn (approved):

PR #687, opened 2021-03-11
Only show auto definitions to admins

andrewdotn:
This is a minimal set of changes so that auto-translated definitions are only shown to logged-in users for now.

While I can think of many ways to improve this code, I may just be making unnecessary work for myself.

### DIFF
--- a/CreeDictionary/API/helpers.py
+++ b/CreeDictionary/API/helpers.py
@@ -1,0 +1,7 @@
+def serialize_definitions(definitions, include_auto_definitions: bool):
+    ret = []
+    for definition in definitions:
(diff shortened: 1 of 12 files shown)


eddieantonio (approved):
Looks good, but the changes are some smelly shotgun surgery. I'm wondering if you have suggestions for how to refactor the codebase to make changes to autodefinitions involve a lot less editing things in several differen places. Perhaps replace conditionals with polymorphism? Or decorate the search results with a filter to remove auto definitions? Or to add autodefinitions?

eddieantonio (approved):
:shipit:

PR #691, opened 2021-03-12
Cypress test for auto translate

andrewdotn:
- If logged in, you see an auto-translation for acâhkosa
  - If not logged in, you do not see any auto-translations


andrewdotn:
While making the tests pass, I cleaned out a bunch of unused scripts and pipenv `[scripts]` entries. Let me know if you’d prefer that to go into a separate PR.

cypress/integration/admin.spec.js
```diff
@@ -0,0 +1,44 @@
+const { join } = require('path');
+
+context('Admin interface', () => {
+  it('should redirect anonymous users to the login page', function() {
+    cy.visit('/admin')
+    cy.location().should('match', /\/admin\/login\//)
```
dwhieb, line 6: Just a little more readable.

```suggestion
    cy.location().should(`contain`, `/admin/login`)
```

cypress/integration/admin.spec.js
```diff
@@ -0,0 +1,44 @@
+const { join } = require('path');
```
dwhieb, line 1: "join" loses its context when it's destructured from `path`—it's not clear what's being joined.

```suggestion
const path = require('path'); // and then just use path.join() elsewhere
```

```suggestion
const { join: joinPathSegments } = require('path');
```

cypress/integration/admin.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+context('Admin interface', () => {
+  it('should redirect anonymous users to the login page', function() {
+    cy.visit('/admin')
+    cy.location().should('match', /\/admin\/login\//)
+  })
+
+  function login() {
+    cy.visit('/admin')
+    cy.readFile(join(__dirname, '..', '..', 'CreeDictionary', '.cypress-user.json')).then(cypressUser => {
```
dwhieb, line 11:
```suggestion
    cy.readFile(path.join(__dirname, `../../CreeDictionary/.cypress-user.json`)).then(cypressUser => {
```

cypress/integration/admin.spec.js
```diff
@@ -0,0 +1,44 @@
...
+      cy.get('#id_username').type(cypressUser.username)
+      cy.get('#id_password').type(cypressUser.password)
+      cy.get('.submit-row > input').click()
+    })
+    cy.location('pathname').should('eq', '/admin/')
```
dwhieb, line 16: Should this assertion be inside the callback for `cy.readFile().then()`? I _think_ the way it is currently, this assertion will resolve as true before Cypress finishes logging in. Or does this timing not matter here?

cypress/integration/admin.spec.js
```diff
@@ -0,0 +1,44 @@
+const { join } = require('path');
+
+context('Admin interface', () => {
+  it('should redirect anonymous users to the login page', function() {
+    cy.visit('/admin')
+    cy.location().should('match', /\/admin\/login\//)
+  })
+
+  function login() {
```
dwhieb, line 9: We might not have to use the user interface to log in for each test, which the Cypress docs treat as an anti-pattern. (Though I vaguely remember you saying that it was difficult to log in programmatically in the app? Just ignore this if so.) We could [add a `cy.login()` command](https://docs.cypress.io/api/cypress-api/custom-commands.html#Syntax) for testing purposes.

Cypress has some useful docs on logging in during tests:

* https://docs.cypress.io/guides/references/best-practices.html#Organizing-Tests-Logging-In-Controlling-State
* https://github.com/cypress-io/cypress-example-recipes#logging-in-recipes

dwhieb (commented):
Just some minor comments / suggestions on the Cypress tests.

nienna73:
You might consider moving the URLs into their own settings file to make them easier to manage, otherwise these tests look good!

dwhieb (approved):

PR #698, opened 2021-03-16
Combine search survey results into sample.csv

andrewdotn:


CreeDictionary/search_quality/combine_samples.py
```diff
@@ -0,0 +1,145 @@
...
+            if have_header:
+                # Don’t yield first line
+                continue
+
+        if have_header:
+            if len(row) > column_count:
+                raise Exception(f"More entries in row {row} than header values")
+
+        query_term = row[0].strip()
+        max_col = column_count
```
nienna73, line 61: I think it's possible that `column_count` is still set to `None` here

CreeDictionary/search_quality/combine_samples_test.py
```diff
@@ -0,0 +1,124 @@
...
+
+def test_combine_results_use_voting_for_disagreement2():
+    # This matches the current survey man: a word in #2 position wins if also
+    # mentioned in #3 position on another list, versus a word only mentioned in
+    # a single list.
+    assert combine_results([["bar"], ["bar", "baz"], ["bar", "qux", "baz"]]) == [
+        "bar",
+        "baz",
+        "qux",
+    ]
```
nienna73, line 124: Yay for tests!

dwhieb (approved):
Documenting my official approval from Zoom 👍🏻😁

CreeDictionary/search_quality/combine_samples.py
andrewdotn: Yup, that is intentional, but I thank you for asking about that because its usage is a bit obscure. I have added a comment, and am happy to revisit if you’d like the code cleaned up further.

PR #718, opened 2021-03-24
Refactor search

andrewdotn:
This shouldn’t change any behaviour, I only moved things around.

The current search is doing a lot of different things, and I want to add more things without it becoming unmaintainable. To start, this moves some of the different pieces into separate files, and breaks the main circular-import problem that plagued users of this code.


eddieantonio (commented):
This is awesome! ..and basically something I wanted to do in the first place anyway!

nienna73 (approved):

PR #722, opened 2021-03-29
perf: faster startup

eddieantonio:
**EDIT**: In this PR, I'm trying to solve this problem:

```
itwewina_1  | WSGI app 0 (mountpoint='') ready in 44 seconds on interpreter 0x55edee7f9ff0 pid: 6 (default app)
```

44 second startup? On my SSD? No good. 🙅🏼

### DIFF
--- a/CreeDictionary/API/apps.py
+++ b/CreeDictionary/API/apps.py
@@ -35,9 +35,9 @@ class APIConfig(AppConfig):
             self.perform_time_consuming_initializations()
 
     def perform_time_consuming_initializations(self):
-        logger.debug("loading affix caches")
         from API.search import affix
 
+        logger.debug("loading affix caches")
         affix.cache.preload()
 
         logger.debug("calling initialize_preverb_search()")
--- a/CreeDictionary/API/search/affix.py
+++ b/CreeDictionary/API/search/affix.py
@@ -129,14 +129,16 @@ def fetch_english_keywords_with_ids():
     """
     Return pairs of indexed English keywords with their corresponding Wordform IDs.
     """
-    return EnglishKeyword.objects.all().values_list("text", "lemma__id")
+    # Slurp up all the results to prevent walking the database multiple times
+    return tuple(EnglishKeyword.objects.all().values_list("text", "lemma__id"))
 
 
 def fetch_cree_lemmas_with_ids():
     """
     Return pairs of Cree lemmas with their corresponding Wordform IDs.
     """
-    return Wordform.objects.filter(is_lemma=True).values_list("text", "id")
+    # Slurp up all the results to prevent walking the database multiple times
+    return tuple(Wordform.objects.filter(is_lemma=True).values_list("text", "id"))
 
 
 class _Cache:
--- a/CreeDictionary/uwsgi.ini
+++ b/CreeDictionary/uwsgi.ini
@@ -33,14 +33,6 @@ close-on-exec = true
 # With post-buffering off, the application MUST read EVERY POST body or else
 # the communication protocol could break :/
 post-buffering = true
-# 🩹 loads the wsgi application **AFTER** forking off workers.
-# The application can hang on startup and as of this writing (2021-03-17),
-# we don't know why.
-# (see: https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/697)
-# If this hang happens BEFORE forking off the workers, then THE ENTIRE SITE
-# IS INACCESSIBLE!
-# See: https://uwsgi-docs.readthedocs.io/en/latest/articles/TheArtOfGracefulReloading.html?highlight=lazy-apps#preforking-vs-lazy-apps-vs-lazy
-lazy-apps = true
 
 
 # Hooks
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -37,5 +37,4 @@ services:
       - "9191:9191"
     volumes:
       - "./.env:/app/.env"
-      # TODO: should the production database be stored in /data/application-data/itwewina?
-      - "./db.sqlite3:/app/CreeDictionary/db.sqlite3"
+      - "/data_local/application-data/itwewina/db.sqlite3:/app/CreeDictionary/db.sqlite3"


eddieantonio:
Startup is a lot faster now!

**NOTE**: this PR is only safe to merge once itwewina is moved back to `altlab-itw`!

---

I tested branch on (the newly reconfigured!) `altlab-itw` and here's the relevant time on the startup logs:

```
itwewina_1  | *** Operational MODE: preforking+threaded ***
itwewina_1  | WSGI app 0 (mountpoint='') ready in 2 seconds on interpreter 0x55cb70147ff0 pid: 6 (default app)
itwewina_1  | running "print:Application loaded!" (post app)...
itwewina_1  | Application loaded!
```


Summary of things I did:

 - copy the database on altlab-itw to the _other_ mount, namely `/data_local`
 - `VACUUM` the database
 - batch database reads: slurp all the required info from the database before using it
 - intialize the app once (and thus, initialize the affix trees once) to avoid I/O contention 😬 

Things I could still do:

 - make a [partial index](https://sqlite.org/partialindex.html) over `API_Wordform.is_lemma = 1`.
 
Though I think with the above improvements alone, getting a 2 second startup time is quite nice!

---

The big issue on startup is bullding the affix trees. Building the affix trees _requires_ the following two queries:


**Cree wordforms**:

```sql
SELECT "API_wordform"."text", "API_wordform"."id" FROM "API_wordform" WHERE "API_wordform"."is_lemma" = 1;
```

**English keywords**:

```sql
SELECT "API_englishkeyword"."text", "API_englishkeyword"."lemma_id" FROM "API_englishkeyword"
```

Both of these queries **require** a database scan.

```
sqlite> EXPLAIN QUERY PLAN SELECT "API_wordform"."text", "API_wordform"."id" FROM "API_wordform" WHERE "API_wordform"."is_lemma" = 1;
QUERY PLAN
`--SCAN TABLE API_wordform
sqlite> EXPLAIN QUERY PLAN SELECT "API_englishkeyword"."text", "API_englishkeyword"."lemma_id" FROM "API_englishkeyword";
QUERY PLAN
`--SCAN TABLE API_englishkeyword
```

---

So let's make this faster!

First of all, since database scans are _required_, I decided to run [`VACUUM`](https://www.sqlite.org/lang_vacuum.html), whcih rebuilds the database file. I _assume_ this encouraged linear accesses, due to this line in the SQLite3 docs:

> Running `VACUUM` ensures that each table and index is largely stored contiguously within the database file.

I was sloppy and lost the exact measurements, but this _did_ improve startup time. ¯\\\_(ツ)\_/¯ 

---

I changed the uwsgi conf to only load the application once (and thus, only build the affix trees once, and thus, only to do a database scan once). This will reduce the disk contention on startup!

---

The biggest startup queries were the ones used for the building the affix trees. I changed it so that the results are all slurped up at once. I did this to goad Django into doing as much IO upfront, and avoid -- doing a little bit of IO, then doing a tiny bit of work, then doing a little bit of IO, then doing a tiny bit of work. I'm assuming that operating system and driver firmware conspire to make big access all at once fast, yet they won't query as quickly if there is a pause between different read events.

@andrewdotn: let me know what you think!

eddieantonio:
One more thing!

I wanted to figure out how quick the `Wordform` and `EnglishKeyword` queries would be, so I made this hacky script:

```python
# CreeDictionary/debug_query_perf.py
"""
Usage:

    $ python manage.py shell
    >>> from debug_query_perf import *
    >>> cree_query()
    >>> english_query()

"""
from API.models import EnglishKeyword
from API.models import Wordform

import time


def timed(fn):
    from time import monotonic

    def wrapped():
        start = monotonic()
        fn()
        stop = monotonic()
        return stop - start

    return wrapped


@timed
def cree_query():
    tuple(Wordform.objects.filter(is_lemma=True).values_list("text", "id"))


@timed
def english_query():
    tuple(EnglishKeyword.objects.all().values_list("text", "lemma__id"))
```

Here are a few runs on `altlab-itw`:

```
itwewina@fd2051baa954:/app/CreeDictionary$ python3 manage.py shell
from Python 3.9.1 (default, Feb  9 2021, 07:55:26)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from debug_query_perf import *
>>> cree_query()
0.295579623998492
>>> cree_query()
0.2569739389982715
>>> cree_query()
0.2554483780004375
>>> cree_query()
0.2543556040000112
>>> cree_query()
0.25551015900055063
>>> cree_query()
0.25614316200153553
>>> cree_query()
0.25546982000014395
>>> english_query()
0.11456292599905282
>>> english_query()
0.09624667699972633
>>> english_query()
0.09584621800240711
>>> english_query()
0.0964166749981814
>>> english_query()
0.09624442299900693
>>> english_query()
0.0956228649993136
>>> english_query()
0.09627691199784749
>>> english_query()
0.096052991997567
>>> cree_query()
0.2628897219983628
```

Notably, the first call to `cree_query()` is noticably slower than every subsequent call ¯\\\_(ツ)\_/¯

eddieantonio:
> I am confused about many things here. I would enjoy spending a bunch of time investigating them but I think that would jeopardize other development goals. All I can offer are some comments.

You're right. I'll get itwêwina running on altlab-itw tomorrow morning and then we can continue.

> You got a 44-second startup _on your laptop_??? Something seems very wrong there because I have never seen startup times like that in development. Did `vacuum` fix that on your laptop?

Yes and yes, but I was not dilligent in controlling for other factors. But that number was copy-pasted from the startup occuring on my laptop. Note: the database was on my macOS boot parition, then mounted, through the Docker virtual machine, into the Docker container. So there's a lot there!

> While normally you’d want to avoid database scans and prefer index lookups, on spinning disks where you are reading a substantial portion of the table, it is faster to sequentially read everything and have the CPU throw away unnecessary data, even if does that millions of times per second, than to physically reposition the drive’s read head, which can only happen hundreds of times per second. We could do some I/O benchmarking to see if the underlying storage has the expected characteristics.
> 
> While a fancy new thing I was surprised to learn about, you don’t need a partial index. If you create a normal index on `(is_lemma, text)`, then the database engine will be easily able to skip all the `is_lemma=0` items when querying by `text`.

It may actually be worth creating this index! The affix search _must_ do a database scan. We also have this query:

https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/e83060f486c53a315320de9921dd2f59d5b4672e/CreeDictionary/CreeDictionary/views.py#L49-L53

> I am still very curious what `vmtouch` shows for whether the database file ends up getting fully cached in RAM or not. With tons of RAM, in theory `cat database > /dev/null` should load it all into the cache. You can also use [`drop_caches`](https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/sysctl/vm.rst#drop_caches) to test startup time on a cold cache.

I may try that tomorrow morning! 

> PS: `filefrag -e /opt/docker-compose/itwewina/cree-intelligent-dictionary/docker/db.sqlite3` on altlab-itw shows where the file blocks are stored on disk, and it’s in 38 pieces. This could be correlated with this other tool that would show exactly which blocks are being read on startup and how long it’s taking to read them: https://github.com/brendangregg/perf-tools/blob/master/examples/iosnoop_example.txt

I forgot that fragmenting matters on a spinning disk! It's quite possible that `VACUUM` not only had the effect of rearranging the blocks in logical database file sequentially, but _also_ in writing the blocks in sequential areas on the hard disk ¯\\\_(ツ)\_/¯

eddieantonio:
itwêwina is running on `altlab-itw` again.

I have copied the database on `altlab-kor` to `/data_local/application-data/itwewina/db.sqlite3`, `VACUUM`ed. It's in three extents:

```
easantos@altlab-itw:~$ filefrag -eX /data_local/application-data/itwewina/db.sqlite3
Filesystem type is: ef53
File size of /data_local/application-data/itwewina/db.sqlite3 is 1133850624 (276819 blocks of 4096 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..     7ff:    1336800..   1336fff:    800:
   1:      800..     fff:    1337800..   1337fff:    800:    1337000:
   2:     1000..    8fff:    1339000..   1340fff:   8000:    1338000:
   3:     9000..   10fff:    1341000..   1348fff:   8000:
   4:    11000..   18fff:    1349000..   1350fff:   8000:
   5:    19000..   20fff:    1351000..   1358fff:   8000:
   6:    21000..   28fff:    1359000..   1360fff:   8000:
   7:    29000..   30fff:    1361000..   1368fff:   8000:
   8:    31000..   38fff:    1369000..   1370fff:   8000:
   9:    39000..   40fff:    1371000..   1378fff:   8000:
  10:    41000..   43952:    1379000..   137b952:   2953:             last,eof
/data_local/application-data/itwewina/db.sqlite3: 3 extents found
```

andrewdotn (approved):
Please sync the current DB right before the deploy.

PR #723, opened 2021-03-31
xmlimport: Write out JSONL inconsistency report

andrewdotn:


eddieantonio (approved):
LGTM! This is a lot better than just printing a log message!

PR #724, opened 2021-03-31
Make morpheme rankings / preverb cache lazy

andrewdotn:
The spooky-action-at-a-distance where these are initially set to empty
values that calling code can use and make invalid decisions based upon,
unless you call the right thing at startup from tests, has been causing
issues for me for a while.

### DIFF
--- a/CreeDictionary/API/apps.py
+++ b/CreeDictionary/API/apps.py
@@ -32,15 +32,12 @@ class APIConfig(AppConfig):
             self.perform_time_consuming_initializations()
 
     def perform_time_consuming_initializations(self):
-        logger.debug("loading affix caches")
+        logger.debug("preloading caches")
         from API.search import affix
-        from API.models import WordformCache
+        from API.models import wordform_cache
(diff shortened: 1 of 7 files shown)


CreeDictionary/API/apps.py
```diff
@@ -37,14 +34,13 @@ def ready(self) -> None:
...
         affix.cache.preload()
 
-        logger.debug("calling initialize_preverb_search()")
-        initialize_preverb_search()
-
-        logger.debug("calling read_morpheme_rankings()")
-        read_morpheme_rankings()
+        # Accessing these cached properties will preload them
+        WordformCache.PREVERB_ASCII_LOOKUP
+        WordformCache.MORPHEME_RANKINGS
```
eddieantonio, line 43: Not necessary, but this matches the `affix.cache.preload()` line from above:

```suggestion
        WordformCache.preload()
```

eddieantonio (approved):
LGTM!

PR #727, opened 2021-04-01
Refactor search, part 2

andrewdotn:
Building on the previous move into separate files, this introduces more
classes and objects. The general workflow is:

  - Search methods generate candidate Result objects, recording the
    features that indicate why a given Result may be a good match for the
    query

  - Features for the same results are aggregated together and can be used
    for ranking. If *multiple* search methods select a given result, that
    can be used as a signal to prioritize that result.

  - Presentation features like relabelling are handled separately in a
    later phase of the process

Again, user-facing changes should be minimal at this time. This is just
rearranging stuff so that there is room to add new functionality later.

The main outstanding issue that I am aware of at this point is that
changing the search result serialization may break click-in-text, in which
case I will have to add legacy compatibility keys.

### DIFF
--- a/CreeDictionary/API/schema.py
+++ b/CreeDictionary/API/schema.py
@@ -1,7 +1,7 @@
 """
 this file contains `TypedDict` classes that effectively serves as json schema for serialized objects
 """
-from typing import List, Tuple, Union
+from typing import List, Tuple, Union, Optional, Sequence
 
 from typing_extensions import Literal, TypedDict
 
@@ -10,7 +10,7 @@ from utils.types import FSTTag
 
 class SerializedDefinition(TypedDict):
     text: str
-    source_ids: Tuple[str, ...]
+    source_ids: Sequence[str]
 
 
 class SerializedWordform(TypedDict):
@@ -46,7 +46,9 @@ class SerializedLinguisticTag(TypedDict):
 
 class SerializedSearchResult(TypedDict):
     # the text of the match
-    matched_cree: str
+    wordform_text: str
+    # legacy name, deprecated
+    matched_cree: Optional[str]
 
     is_lemma: bool
 
@@ -70,11 +72,5 @@ class SerializedSearchResult(TypedDict):
 
     # This omits preverbs and other features displayed elsewhere
     relevant_tags: Tuple[SerializedLinguisticTag, ...]
-
-    # TODO: there are things to be figured out for this :/
-    # Sequence of all reduplication tags present, in order
-    reduplication_tags: Tuple[str, ...]
-    # Sequence of all initial change tags
-    initial_change_tags: Tuple[str, ...]
 
     definitions: Tuple[SerializedDefinition, ...]
--- a/CreeDictionary/API/search/__init__.py
+++ b/CreeDictionary/API/search/__init__.py
@@ -1,15 +1,7 @@
-from sortedcontainers import SortedSet
-
-from .lookup import (
-    WordformSearchWithAffixes,
-    WordformSearchWithExactMatch,
-)
-from .types import SearchResult
+from .runner import search
 
 
-def search_with_affixes(
-    query: str, include_auto_definitions=False
-) -> SortedSet[SearchResult]:
+def search_with_affixes(query: str, include_auto_definitions=False):
     """
     Search for wordforms matching:
      - the wordform text
@@ -18,18 +10,20 @@ def search_with_affixes(
      - affixes of the definition keyword text
     """
 
-    search = WordformSearchWithAffixes(query)
-    return search.perform(include_auto_definitions=include_auto_definitions)
+    return search(
+        query=query, include_auto_definitions=include_auto_definitions
+    ).serialized_presentation_results()
 
 
-def simple_search(
-    query: str, include_auto_definitions=False
-) -> SortedSet[SearchResult]:
+def simple_search(query: str, include_auto_definitions=False):
     """
     Search, trying to match full wordforms or keywords within definitions.
 
     Does NOT try to match affixes!
     """
 
-    search = WordformSearchWithExactMatch(query)
-    return search.perform(include_auto_definitions=include_auto_definitions)
+    return search(
+        query=query,
+        include_affixes=False,
+        include_auto_definitions=include_auto_definitions,
+    ).serialized_presentation_results()
--- a/CreeDictionary/API/search/affix.py
+++ b/CreeDictionary/API/search/affix.py
@@ -20,11 +20,10 @@ from django.conf import settings
 from API.models import Wordform, EnglishKeyword
 from utils.cree_lev_dist import remove_cree_diacritics
 from .types import (
-    EnglishResult,
-    MatchedEnglish,
-    CreeResult,
     InternalForm,
+    Result,
 )
+from . import core
 
 # A simplified form intended to be used within the affix search trie.
 SimplifiedForm = NewType("SimplifiedForm", str)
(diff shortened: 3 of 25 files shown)


CreeDictionary/API/views.py
```diff
@@ -18,9 +15,7 @@ def click_in_text(request) -> HttpResponse:
     elif q == "":
         return HttpResponseBadRequest("query param q is an empty string")
 
-    results: List[SerializedSearchResult] = []
-    for result in simple_search(q, include_auto_definitions=False):
-        results.append(result.serialize(include_auto_definitions=False))
+    results = simple_search(q, include_auto_definitions=False)
```
eddieantonio, line 18: I just want to make sure we're not breaking the click-in-text API that's being used! So this should have tests ensuring the refactor did not, in fact, introduce existing changes.

Once that's handled, feel free to merge!

CreeDictionary/API/schema.py
```diff
@@ -46,7 +46,9 @@ class SerializedLinguisticTag(TypedDict):
 
 class SerializedSearchResult(TypedDict):
     # the text of the match
-    matched_cree: str
+    wordform_text: str
+    # legacy name, deprecated
+    matched_cree: Optional[str]
```
nienna73, line 51:
```suggestion
    matched_source_language: Optional[str]
```

CreeDictionary/API/search/__init__.py
```diff
@@ -1,15 +1,7 @@
...
-    WordformSearchWithAffixes,
-    WordformSearchWithExactMatch,
-)
-from .types import SearchResult
 
-
-def search_with_affixes(
-    query: str, include_auto_definitions=False
-) -> SortedSet[SearchResult]:
+def search_with_affixes(query: str, include_auto_definitions=False):
```
nienna73, line 4: Consider putting the return type back into this function definition

CreeDictionary/API/search/affix.py
```diff
@@ -125,15 +124,15 @@ def query_would_return_too_many_results(query: InternalForm) -> bool:
     return len(query) <= settings.AFFIX_SEARCH_THRESHOLD
 
 
-def fetch_english_keywords_with_ids():
+def fetch_target_language_keywords_with_ids():
     """
     Return pairs of indexed English keywords with their corresponding Wordform IDs.
     """
     # Slurp up all the results to prevent walking the database multiple times
     return tuple(EnglishKeyword.objects.all().values_list("text", "lemma__id"))
```
nienna73, line 132: Should this class also be changed to not reference a specific language?

CreeDictionary/API/search/lookup.py
```diff
@@ -242,25 +141,49 @@ def _fetch_results(
...
-                EnglishResult(MatchedEnglish(user_query), wordform, Lemma(wordform))
-            )  # will become  (user_query, inflection.text, wordform)
+            search_run.add_result(Result(wordform, is_preverb_match=True))
+
+
+def filter_cw_wordforms(q: Iterable[Wordform]) -> Iterable[Wordform]:
+    """
+    return the wordforms that has definition from CW dictionary
+
+    :param q: an Iterable of Wordforms
```
nienna73, line 168: Is there a better name for `q`? I'm not totally sure it needs a better name since this function is so small, but it would be nice to have a better idea of what it is based on its name. Either way, this docstring explanation is nice 👍

CreeDictionary/API/search/presentation.py
```diff
@@ -0,0 +1,177 @@
...
+
+    for tag in head_breakdown:
+        preverb_result: Optional[Preverb] = None
+        if tag.startswith("PV/"):
+            # use altlabel.tsv to figure out the preverb
+
+            # ling_short looks like: "Preverb: âpihci-"
+            ling_short = LABELS.linguistic_short.get(tag)
+            if ling_short is not None and ling_short != "":
+                # looks like: "âpihci"
```
nienna73, line 152: This comment seems to disagree with the one above. How does it go from "Preverb: âpihci-" to "âpihci"?

CreeDictionary/API/search/types.py
```diff
@@ -1,80 +1,24 @@
...
-from API.models import Wordform, Definition
-from API.schema import SerializedLinguisticTag, SerializedSearchResult
-from utils import Language
+from API.models import Wordform
+from API.schema import SerializedLinguisticTag
 from utils.fst_analysis_parser import LABELS
-from utils.types import ConcatAnalysis, FSTTag
+from utils.types import FSTTag
 
 # it's a str when the preverb does not exist in the database
```
nienna73, line 15: Is this comment still accurate/relevant?

CreeDictionary/API/search/types.py
```diff
@@ -138,86 +82,90 @@ def linguistic_tag_from_fst_tags(tags: Tuple[FSTTag, ...]) -> LinguisticTag:
...
+@dataclass
+class Result:
     """
+    A target-language wordform and the features that link it to a query.
 
-    # the text of the match
-    matched_cree: str
+    The features—is it an exact source-language match for the query? Is the edit
+    distance low? &c.—that make this a candidate result for a search query are
+    required to rank different results before showing them to the user.
```
nienna73, line 107: I'm having a hard time following what this is saying

CreeDictionary/search_quality/analyze_results.py
```diff
@@ -57,12 +57,20 @@ class DuplicateAnnotatedSearchResult(TypedDict, total=False):
...
 
-    raw_suffix_tags: tuple[FSTTag, ...]
 
-    is_duplicate_of: Optional[str]
+def get_result_text(r: DuplicateAnnotatedSearchResult | SerializedSearchResult):
+    # Use the new language-independent key if available
+    if "wordform_text" in r:
+        return r["wordform_text"]
+    # Fall back to old thing
+    return r["matched_cree"]
```
nienna73, line 73: Is the idea that we can eventually take this out?

CreeDictionary/search_quality/analyze_results.py
```diff
@@ -113,7 +120,7 @@ def find_rank(word: str, results: list[SerializedSearchResult]):
     index = 1
     for r in results:
         # We also have the definition here, could include it
-        if word == r["matched_cree"]:
+        if word == get_result_text(r):
             return index
         index += 1 if r["is_lemma"] else 2
     return "☹"
```
nienna73, line 126: 😂

nienna73 (approved):
Overall, this looks awesome! I left a few comments/questions/suggestions, but otherwise it looks good to go from my perspective.

CreeDictionary/search_quality/analyze_results.py
andrewdotn: Yeah, this is the same thing as https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/pull/727#pullrequestreview-631585858

I can remove it, but then it turns into a bigger thing with needing to update the search result sampling code used on older branches for comparison purposes.

CreeDictionary/API/search/__init__.py
andrewdotn: I prefer not to explicitly specify types that mypy can infer. I have updated the `.serialize()` method that is called here to be annotated with a type instead. Now mypy correctly infers the return type as `list[SerializedPresentationResult]`.

CreeDictionary/API/search/affix.py
andrewdotn: I’m leaving that for when the `EnglishKeyword` database table gets renamed.

CreeDictionary/API/search/lookup.py
andrewdotn: I did not write this code, I just moved it around, and this functionality really doesn’t belong in the dictionary server but in whatever process produces the source dictionary used by itwêwina; that said `queryset` is a better name than `q` here so I have renamed it.

CreeDictionary/API/search/presentation.py
andrewdotn: The line below the comment transforms `Preverb: âpihci-` to `âpihci`; I have updated the comment to try to clarify that.

CreeDictionary/API/search/types.py
andrewdotn: Good catch, thank you!

CreeDictionary/API/search/types.py
andrewdotn: Is this newer version more understandable, or does it need more work?


    """
    A target-language wordform and the features that link it to a query.

    Features of a wordform allow better inferences about how good a result is.

    Some examples of these features could include:
      - Is this Result a match between a source-language query and wordform
        text, or between a target-language query term and a word in the
        definition text?
      - What is the edit distance between the query term and the wordform?
      - What is the best cosine vector distance between a definition of the
        result wordform and the query terms?

    The best search results will presumably have better scores on more features
    of greater importance.

    Search methods may generate candidate results that are ultimately not sent
    to users, so any user-friendly tagging/relabelling is instead done in
    PresentationResult class.
    """

CreeDictionary/API/search/lookup.py
nienna73: That's fair, I was pretty sure you had just moved it like you said. Thanks for making it clearer all the same!

CreeDictionary/API/search/presentation.py
nienna73: Ohh, I see now.

CreeDictionary/API/search/types.py
nienna73: This is much clearer to me 😄

CreeDictionary/search_quality/analyze_results.py
nienna73: No need to remove it now, it's just good to know that this is supporting older code.

PR #728, opened 2021-04-01
Adjust database path and production docker paths

andrewdotn:
- Mount the database from a directory, not a file, so that we can safely
    enable WAL mode, and to make some other database maintenance tasks a
    little easier

  - Use a persistent path for storing search sample results

Before deploying this change, hard-link the current database file to the
new location so that the transition is seamless.

And all developers will also need to move any existing database files to
the new path.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -21,5 +21,10 @@ services:
       - "9191:9191"
     volumes:
       - "./.env:/app/.env"
+      # Mount directory, not file, so that -shm and -wal files will be
+      # persisted outside the container if using WAL mode
       - "/data_local/application-data/itwewina/db/:/app/CreeDictionary/db/"
+      # Use a persistent path for search sample results, so that they
+      # remain across deployments and so that additional sample results can
+      # be uploaded from other sources for comparison purposes
       - "/data_local/application-data/itwewina/search-quality/:/app/CreeDictionary/search_quality/sample_results/"
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,20 +1,4 @@
-# Base docker-compose for itwewina.
-#
-# NOTE!!!!!!!!!
-# This docker-compose file is useless standalone. You **MUST** combine it with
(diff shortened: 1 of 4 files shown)


docker/docker-compose.yml
```diff
@@ -37,4 +21,5 @@ services:
       - "9191:9191"
     volumes:
       - "./.env:/app/.env"
-      - "/data_local/application-data/itwewina/db.sqlite3:/app/CreeDictionary/db.sqlite3"
+      - "/data_local/application-data/itwewina/db/:/app/CreeDictionary/db/"
+      - "/data_local/application-data/itwewina/search-quality/:/app/CreeDictionary/search_quality/sample_results/"
```
eddieantonio, line 25: I have created the requisite directories and hard links!

```
easantos@altlab-itw:~$ ls -liRF /data_local/application-data/itwewina/
/data_local/application-data/itwewina/:
total 1107292
3034048 -rw-r--r-- 1 root     root            496 Apr  1 11:57 README
3034044 drwxr-xr-x 2 itwewina itwewina       4096 Apr  1 11:52 db/
3014661 -rw-r--r-- 2 itwewina itwewina 1133850624 Apr  1 10:19 db.sqlite3
3034045 drwxr-xr-x 2 itwewina itwewina       4096 Apr  1 11:51 search-quality/

/data_local/application-data/itwewina/db:
total 1107280
3014661 -rw-r--r-- 2 itwewina itwewina 1133850624 Apr  1 10:19 db.sqlite3

/data_local/application-data/itwewina/search-quality:
total 0
```

eddieantonio (commented):
For the `db/` folder, I think it will cause less headaches in the future to write documentation as to WHY we did it like this (like a README). Also, I believe some things in the `docker/Makefile` need to be adjusted, because they assume a file to be mounted in the container.

eddieantonio (approved):

docker/docker-compose.yml
andrewdotn: Thank you!

Issue #708, opened 2021-03-18
Implement settings menu/page for selecting interface/functionality options


aarppe:
So that users – language instructors, learners, community members, linguists – can choose exactly the functionality, dictionary sources, and labeling as well as othe behavior that they want, we should implement the following settings menu/window for selecting among all the options:

<img width="598" alt="image" src="https://user-images.githubusercontent.com/15003407/111917665-ea689b80-8a46-11eb-96a4-e8e7f6e7020e.png">

Some of Cree terms yet need some validation by Jean / Arok.

aarppe:
A task-oriented listing:
---
- [ ] **Interface language / ohtiskawi-ayamiwin**
   - [ ]	nêhiyawêwin / Plains Cree 
   - [ ]	English / âkayâsîmowin
---
- [ ] **Dictionary sources / itwêwasinahikana**
   - [ ]	Cree : Words / nêhiyawêwin : itwêwina
   - [ ]	Maskwacîs Cree Dictionary / nêhiyaw-pîkiskwêwinisa
   - [ ]	Alberta Elders’ Cree Dictionary / alperta ohci kêhtê-ayak nêhiyaw otwêstamâkêwasinahikan
---
- [ ] **Labels / tânisi ê-wîhcikâtêki**
   - [ ]	linguistic / kanawâpahtamowin ayamiwina
   - [ ]	plain English / mosci-âkayâsîmowin
   - [ ]	nêhiyawêwin / Plains Cree
   - [ ]	emojis / cikâstêpisona
---
- [ ] **Behavior / isîhtwâwin**
   - [ ]	show morpheme boundaries / wâpahtahin tânisi itwêwin ê-pâh-piskihcayâk icwêwinisihk
 	icwêwinisa 
   - [ ]	show Cree words translated into English phrases / wâpahtahin tânisi âkayâsîwastâniwan nêhiyawi-itwêwina âkayâsîwi-pîkiskwêwinihk
   - [ ]	show synthesized pronunciations of Cree words / wâpahtahin tânisi âpacihcikan nêhiyawi-itwêwina ê-itahk
---

aarppe:
Based on our Tech Team meeting discussion, the desired order of implementation would be the following, in part due to "customer" demand and in part due to the necessary components already being implemented:
[...]

PR #729, opened 2021-04-01
feat:  implement display mode buttons

eddieantonio:
~~Add a toggle for "basic" and "traditional" mode. Basic is what is up now; "traditional" mode will display something like "Noun (NAD-4)" instead of emoji on the search page.~~

EDIT: there are two modes: **community** and **linguistic**. The toggles are tucked into the bottom of the screen. Turn on "lingusitic mode" and search results change their display to this:

mîciw, linguistic mode

Compare with the default:

mîciw, community mode

Start to address #708


CreeDictionary/CreeDictionary/display_options.py
```diff
@@ -0,0 +1,18 @@
+"""
+Display options. Currently controls:
+
+ - "mode": either "basic" and "traditional"
+
+As of 2021-04-14, "mode" is a corse mechanism for affecting the display; there are a
```
andrewdotn, line 6:
```suggestion
As of 2021-04-14, "mode" is a coarse mechanism for affecting the display; there are
```

CreeDictionary/CreeDictionary/views.py
```diff
@@ -236,6 +236,41 @@ def google_site_verification(request):
...
+
+    def post(self, request) -> HttpResponse:
+        mode = request.POST.get(DISPLAY_MODE_COOKIE)
+
+        # Tried to set to an unknown display mode
+        if mode not in DISPLAY_MODES:
+            return HttpResponse(status=HTTPStatus.BAD_REQUEST)
+
+        if who_asked_us := request.headers.get("Referer"):
+            # Force the browser to refersh the page that issued this request.
```
andrewdotn, line 262:
```suggestion
            # Force the browser to refresh the page that issued this request.
```

cypress/integration/search/header-display-options.spec.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Similar to ./result-header.spec.js -- displays a more deta
```
andrewdotn, line 2:
```suggestion
 * Similar to ./result-header.spec.js -- displays more data
```

src/css/styles.css
```diff
@@ -597,6 +602,16 @@ a.multiple-recordings__action-button {
   cursor: help;
 }
 
+
+/**
+ * ELEMENT inflecional-class
```
andrewdotn, line 607:
```suggestion
 * ELEMENT inflectional-class
```

andrewdotn (approved):
Looks ok. I know that in the templates themselves it always shows up as `display.mode`, but I have a vague sense that `display` may be too generic a term to use as a default context variable. Maybe `display_mode`? `display_options.mode`? If it does become an issue, we could always change it later.

aarppe:
Would the pop-ups be in reverse?

That is, for the community mode, what you see initially is 1) the emojis and the 'like: X' text, and secondarily in the pop-up 2) the linguistic labels (long 'regular transitive animate verb' and short '(VTA-1)').

And for the linguistic mode, you'd initially see 1) the short linguistic label, e.g. VTA-1, and below that the full linguistic label but then secondarily in the pop-up 2) the example word case 'like: XXX' (but perhaps not the emoji).

Eventually I'd think we'd want to include the nêhiyawêwin labels in the mix.

PR #735, opened 2021-04-09
Remove use of @altlab/types

andrewdotn:
click-in-text no longer uses it


eddieantonio (approved):
LGTM! Less complexity is always good! :D

PR #737, opened 2021-04-14
Speed up sitemap generation

eddieantonio:
The sitemap generation used `Wordform.get_absolute_url()` which would then make many queries to `Wordform.objects.filter(text=...)` which would bog down sitemap generation. No more.

I'm not sure if the new index is useful (I have not measured its effects). Please let me know and I can remove the new index from this PR.


CreeDictionary/API/models.py
```diff
@@ -196,10 +201,19 @@ class Meta:
             models.Index(fields=["analysis"]),
             # text index benefits fast wordform matching (see search.py)
             models.Index(fields=["text"]),
+            # When we *just* want to lookup text wordforms that are "lemmas"
+            # (Note: Eddie thinks "head words" may also be lumped in as "lemmas")
+            # Used by:
+            #  - affix tree intialization
+            #  - sitemap generation
+            models.Index(fields=["is_lemma", "text"], name="lemma_text_idx"),
+            # TODO: do we need these indices?
```
andrewdotn, line 210: grep for `pos=` or `inflectional_category=` and you’ll find they’re an attempt to speed up [building the preverb cache](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/ed8a43b138ab03d45a21800940a7fec23c89376a/CreeDictionary/API/search/lookup.py#L157).

CreeDictionary/API/models.py
```diff
@@ -196,10 +201,19 @@ class Meta:
...
             models.Index(fields=["inflectional_category"]),
             models.Index(fields=["pos"]),
         ]
 
+        ordering = ["is_lemma", "text"]
```
andrewdotn, line 215: I’m curious what the intent here is, as this change will add a global ordering. If sitemap generation needs a specific order, could the ordering be specified there instead?

andrewdotn (approved):
Have you run any tests with the full dictionary for this PR? I know nginx times out the request when google tries to retrieve the sitemap, and I’d put caching / pregenerating it on my todo list.

nienna73 (approved):

eddieantonio:
> Have you run any tests with the full dictionary for this PR?
> I know nginx times out the request when google tries to retrieve the sitemap, and I’d put caching / pregenerating it on my todo list.

Yes. I was wondering why it kept timing out, so I logged all of the queries that Django was making when creating the sitemap on the _full_ database. The culprit was the homograph disambiguator. I hope that @dwhieb's work on the dictionary DB will make this hacky code a thing of the past, but for now, I just instructed the sitemap generator to allow ambiguous URLs.

CreeDictionary/API/models.py
eddieantonio: The reason I did this originally was because the sitemap generator [was emitting warnings](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/runs/2347894309#step:10:477) about an unordered object list. I thought I'd eliminate these warnings once and for all for every purpose, but now I'm realizing that this ordering is quite presumptuous. So I added an `.order_by()` clause on the queryset generated for sitemap generation instead! 👍🏼

PR #738, opened 2021-04-15
Fancy queries

andrewdotn:


CreeDictionary/CreeDictionary/templates/CreeDictionary/query-help.html
```diff
@@ -0,0 +1,49 @@
+{% extends 'CreeDictionary/base.html' %}
+
+{% load creedictionary_extras %}
+
+{% block prose %}
+{% spaceless %}
+<section id="contact-us" class="prose box box--spaced">
```
eddieantonio, line 7:
```suggestion
<section id="query-help" class="prose box box--spaced">
```

CreeDictionary/API/search/query.py
```diff
@@ -0,0 +1,46 @@
+import unicodedata
+from typing import Optional
+
+import marshmallow.fields
+
+from API.search.util import to_sro_circumflex
+
+
+class Query:
```
eddieantonio, line 9: I like this class!

One note: `SearchRun.__str__` delegates to the `__str__` of this class (which delegates to the default `__repr__`). Should this be given a simple `__repr__`. Something like?

```python
    def __repr__(self) -> str:
        return f"<Query {self._raw_query_string!r}>"
```

CreeDictionary/API/search/core.py
```diff
@@ -70,37 +70,3 @@ def serialized_presentation_results(self):
...
         return f"SearchRun<query={self.query}>"
-
-
-def to_internal_form(user_query: str) -> InternalForm:
-    """
-    Convert text to the internal form used by the database entries, tries, FSTs, etc.
-
-    In itwêwina, the Plains Cree dictionary, this means SRO circumflexes.
-    """
-    return InternalForm(to_sro_circumflex(user_query))
```
eddieantonio, line 81: This was part of a half (quarter? eighth?) finished refactor that would make the codebase language-independent. Alas, it never went anywhere.

eddieantonio (approved):
LGTM! Just a few minor notes!

PR #739, opened 2021-04-16
docs: production on whorf

eddieantonio:
# What's in this PR?
 
 - a brief document on how itwêwina is deployed on Whorf

---

We had a document on production on Sapir, but no document for Whorf. This PR deletes the old document and replaces it with a new document. Please read it and tell me if it's missing anything, or if there's anything that is too at-risk of becoming out-of-sync.


dwhieb (approved):
This looks great - I found it completely understandable and it does help me understand the overall process now.

andrewdotn (approved):

PR #752, opened 2021-04-21
refactor: relabellings

eddieantonio:
# What is in this PR?

 - Moves relabelling stuff code to its own module: `CreeDictionary.relabelling`
 - I attempted to update the code to make it a bit more... not bad? But I only did so much.
 - Try to hide the `_LabelFriendliness` enum. It's weird, has a weird name, and generally should not be known to the outside world.

Note: I fully intend to work on paradigm relabellings based on this PR.


CreeDictionary/CreeDictionary/relabelling.py
```diff
@@ -0,0 +1,200 @@
...
+
+        .linguistic_short[tag]  or .linguistic_short.get(tag, default)
+        .linguistic_long[tag]   or .linguistic_long.get(tag, default)
+        .english[tag]           or .english.get(tag, default)
+        .cree[tag]              or .cree.get(tag, default)
+        .emoji[tag]             or .emoji.get(tag, default)
+    """
+
+    # This data structure is kind of a weird, but I'm keeping it for now.
+    _DataStructure = dict[Tuple[FSTTag, ...], dict[_LabelFriendliness, Optional[Label]]]
```
eddieantonio, line 44: **MUST** use `Tuple` from typings in this context due to this issue: https://github.com/python/mypy/issues/9522

Using `tuple[FSTTag, ...]` yields an `unexpected ...` error from mypy.

CreeDictionary/CreeDictionary/relabelling.py
```diff
@@ -0,0 +1,200 @@
...
+import csv
+from enum import IntEnum
+from typing import Iterable, Optional, TextIO, Tuple, TypeVar, Union
+
+from utils import shared_res_dir
+from utils.types import FSTTag, Label
+
+CRK_ALTERNATE_LABELS_FILE = shared_res_dir / "crk.altlabel.tsv"
+
+Default = TypeVar("Default")
```
andrewdotn, line 12: By itself this is very mysterious. What kind of default is this, and for what?

nienna73 (approved):
LGTM

CreeDictionary/CreeDictionary/relabelling.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+class _LabelFriendliness(IntEnum):
+    """
+    Weird enum that I'm not sure should have ever existed.
```
andrewdotn, line 17: - `SOURCE_LANGUAGE` and `TARGET_LANGUAGE` would be more language-independent
  - You can always parse the header to get the column names and let them flow through to the HTML templates; that is, if someone adds a “foo” column in the .tsv, that would then be accessible as `{{ label.foo }}`. But not having looked at how exactly these are used, that may be unhelpfully abstract and a distraction from your present refactoring goals.

andrewdotn (approved):
Some suggestions, but it’s ok not to address them now, as the PR in its current state is definitely an improvement.

CreeDictionary/CreeDictionary/relabelling.py
eddieantonio: This is a generic type variable for the return type of the default argument of `.get(key, default)`. Honestly, Python's generics type variables are really gross! And this function doesn't need to be generic, so I'll just delete it!

CreeDictionary/CreeDictionary/relabelling.py
eddieantonio: Yeah, this is hard-coded to the structure of `crk.altlabel.tsv`. I won't fix it in this PR, because this PR is me getting things ready for paradigm relabelling, but it's a thing to consider when making this language-independent.

aarppe: But this can be more complex than just source and target language. In the case of itwêwina, you would have both linguistic and plain/non-linguistic forms in the target language, and plain/non-linguistic forms in the source language. Also, the linguistic labels can be short or long.

To keep things simple, perhaps one would have a two-way division under the source and target languages, into linguistic and non-linguistic (or plain), but how these are named and accessible in the interface may need to be specified per each language.

In the case of Cree, these relabeling schmemes would be:
1a. Target, linguistic
1b. Target, plain (English)
2b. Target, (plain) nêhiyawêwin

PR #754, opened 2021-04-21
remove: Wordform.md_only

eddieantonio:
This was hack that was required in some of the earliest iterations of itwêwina. I'm glad I can remove it!


nienna73 (approved):
Tests pass so LGTM!

Issue #755, opened 2021-04-22
update glossary with information from ALTLab dictionary database


dwhieb:
Update the glossary and style guide with information from the ALTLab dictionary database. Add quotes / citations / references in support of terms where possible. Make these documents more visible by including a link to them from the README.

dwhieb:
# Notes on the term _lemma_

> Continuous text consists of 'wordforms' (like _permitted_ or _permits_), but the headwords in a dictionary are generally 'lemmas' like _permit_, and this is the object that we usually want to study. A lemmatization program takes as its input the various word-forms and maps them on to the lemma they belong to. (Atkins & Rundell [2008](#AtkinsRundell2008): 88)

* headwords = lemmas
* lemmatization = linking inflections to lemmas

> The canonical form, sometimes called the _lemma_, is the form chosen to represent a paradigm; most headwords, with the exception of cross-references and names, are canonical forms. (Landau [2001](#Landau2001): 98)

> The lemma functions as a representative of a linguistic sign; in a dictionary it represents the lexical item described in the individual dictionary entry. (Svensén [2009](#Svensen2009): 93)

* Very occasionally, the _dictionary_ lemma (headword) will be different from the _linguistic_ lemma, namely in instances where the dictionary lemma requires capitalization (e.g. for proper names).

> A common synonym of 'lemma' is 'headword'. In this book, the term 'lemma' has been preferred mainly because 'headword' will be somewhat problematic when the lemma consists of a multi-word lexical item. (Svensén [2009](#Svensen2009): 93 fn. 1)

# References

* <p id=AtkinsRundell2008>Atkins, B. T. & Michael Rundell. 2008. <cite>The Oxford guide to practical lexicography</cite>. Oxford University Press.</p>
* <p id=Landau2001>Landau, Sidney I. 2001. <cite>Dictionaries: The art and craft of lexicography</cite>. Cambridge University Press.</p>
* <p id=Svensen2009>Svensén, Bo. 2009. <cite>A handbook of lexicography: The theory and practice of dictionary-making</cite>. Cambridge University Press.</p>

aarppe:
[...]

PR #756, opened 2021-04-22
Glossary update

andrewdotn:
- Fix all broken markdown links
  - Standardize on ‘word class’, deprecating ‘specific word class’
  - Add definitions of source and target language
  - Call out contradiction


andrewdotn:
@dwhieb

I saw your comments in #755 after writing this. If lemma = headword, we need a different term (“head” without “word”?) to describe things in dictionaries that have definitions, especially if a dictionary can include phrases as heads.

PR #758, opened 2021-04-23
Update phrase translation FST and tag map

andrewdotn:
The correct translations for these two preverbs are:

  - *ka-* Future Definite = 'will'

  - *wi-* Prospective Aspect = 'is going to'

      - Combines with the future tag, not the intentional tag.

We won’t have the intentional “want to” translation again unless PV/nôhtê
is added to the generator FST.


andrewdotn:
The FST was built from http://github.com/giellalt/lang-crk/commit/0ea2c33323e71bf28ed5bfdef4915d26e72f52ca

dwhieb (approved):
Looks ✨fantabulous✨

PR #760, opened 2021-04-23
Add featuredump manage command

andrewdotn:
The featuredump command runs searches for all the queries in the survey file, and writes out JSONL of the features, just like what you see with `verbose:1`. This allows bulk-loading search results into Jupyter for analysis and modelling.

Also includes a few semi-related things, which can be split into separate PRs if desired:
  - Converts `target_language_keyword_match` feature into a list, so that multiple keyword hits are remembered
  - Apply `black` to a migration file that PyCharm keeps re-editing. I know that `pyproject.toml` has a section to disable black for migrations, but PyCharm doesn’t seem to be following that.
  - Some refactoring to the search survey stuff

### DIFF
--- a/CreeDictionary/API/models.py
+++ b/CreeDictionary/API/models.py
@@ -3,13 +3,16 @@ from __future__ import annotations
 import logging
 from collections import defaultdict
 from pathlib import Path
-from typing import Dict, Literal, Optional
+from typing import Dict, Optional, Union
+from typing import Literal
 from urllib.parse import quote
 
 from django.db import models, transaction
 from django.db.models import Max, Q
 from django.urls import reverse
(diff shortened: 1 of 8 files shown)


CreeDictionary/API/models.py
```diff
@@ -38,6 +38,10 @@ def get_queryset(self):
         return super().get_queryset().select_related("lemma")
 
 
+# This type is the int pk for a saved wordform, or (text, analysis) for an unsaved one.
+WordformKey = Union[int, tuple[str, str]]
```
eddieantonio, line 42: I think I've been meaning to give this its own type for a while!

eddieantonio (approved):
LGTM!

PR #761, opened 2021-04-26
Remove warning about “Database is Locked”

andrewdotn:
I did not see this error the last two times I ran this process.


eddieantonio (approved):
LGTM!

Is this due to WAL mode on production?

andrewdotn:
WAL definitely helps. I thought I’d hit the error once in dev even with WAL mode, but it seems to be working in prod, so I won’t worry about it for now.

(Some web searches led me to believe that we might be able to address this by setting Django to use persistent DB connections, but it’s something I don’t think I should spend time on unless it’s a problem.)

PR #514, opened 2020-07-20
Added paradigm layout files for personal and demonstrative pronouns

eddieantonio:
Fixes #164
Blocked by #513

> These follow two strategies:
> 1. Personal pronouns are not generated but merely spelled out (not in quotes, in contrast to labels);
> 2. Demonstrative pronouns are generated based on their linguistic feature specifications (similar to nouns and verbs).
> 
> — @aarppe

Issue #164, opened 2019-11-07
Certain pronoun classes need inflectional paradigms


aarppe:
It would make sense to create "paradigm" tables for personal pronouns, demonstrative pronouns, as well as interrogative pronouns - these "paradigms" could be invoked from any of their constituent member pronouns, presenting how that pronoun is positioned in relation to the rest in the same subclass. This is related to #127. Don't know about the need to contrast between basic and full paradigms, though.

aarppe:
There are now tentative layouts for personal and demonstrative pronouns, to be found in giella/langs/crk as:

```
inc/paradigms/non-linguistic/pronoun-demonstrative-basic.layout.tsv
inc/paradigms/non-linguistic/pronoun-personal-basic.layout.tsv
```

What needs to be checked is whether one strictly needs the dynamic FST-based generation of the pronominal forms, as in this case they can be and have been easily enumerated, but will the presentation of static forms work with the current interpretation of `*.layout` files?

aarppe:
@eddieantonio You wanted these a while back. As said, for these I'm not generating the forms at all, since the forms per cells are static. Of course, we could still generate them, if that would be useful or necessary for some purpose. There's a file for that in:

`inc/paradigms/pronoun.paradigm`

eddieantonio:
Thanks for this! I'm not super sure what the most elegant way would be to incorporate this into itwêwina would be at the moment, but this is definitely useful!

eddieantonio:
A note about this:

The current layout paradigm parser considers anything in quotes to be a label or title.

So this:

```
        "One"
"I/me/my"       "niya"
"you (one)/your"        "kiya"
"s/he/it/him/her/his/her/its"   "wiya"
        "Many"
"we (but not you)/us/our"       "niyanân"
"you and we/us/our"     "kiyânaw"
"you (all)/your"        "kiyawâw"
"they/them/their"       "wiyawâw"
```

Is a table full of labels, and none of them are actual "forms".

aarppe:
I was toying with various options, first from the old itwêwina standpoint, and in a two column layout I had more abstract row/column labels for `Speaker/Listener/Other` and `One/Many`, and then one could have used the treatment of the cells as labels to provide the exact translation of the cell content (and to distinguish niya with correponding row/column labels of `Speaker + One` as 'I' and `Speaker + Many` as `we`.

[...]

PR #766, opened 2021-04-29
feat: parse static paradigm & generate analyses

eddieantonio:
# What's in this PR

Making progress on the road to using pane-based layouts! This does have a few scattered changes, unfortunately.

 - adds support for parsing "static" and "dynamic" paradigms — by adding the `LiteralCell`: it's literal text!
 - adds a really basic demonstrative pronoun paradigm; @aarppe will create a REAL one, but I made a simple static paradigm for test purposes
 - adds support for generating an (X)FST analyses string. Give it a lemma, and it will generate a big giant string of analyses that can be fed into a generator FST to generate ALL of the wordforms!
 - refactor: separate static from dynamic paradigm layouts in the directory structure — at least in the test fixture directory!
 - refactor: change a bunch of weird methods into properties — this seems to make more sense

# What will happen in ✨ future✨  PRs

 - `ParadigmTemplate` will gain a `.generate()` method that, given a lemma, will use the FST to generate all inflections, replacing its `InflectionCell`s with `LiteralCell`s containing the generated wordform.
 - a new directory in `res` called `paradigm-templates` which will contain both static and dynamic templates
 - something will maintain a cache of all parsed static and dynamic templates, know how to use the generator FST, and, when given a head wordform, it will give you the corresponding paradigm

Addresses #164 
See also: #514


CreeDictionary/CreeDictionary/paradigm/panes.py
```diff
@@ -140,6 +170,10 @@ class Row:
     has_content: bool
     num_cells: int
 
+    @property
+    def inflection_cells(self) -> Iterable[InflectionCell]:
+        yield from ()
```
nienna73, line 175: Is this one of the things to be implemented in the future?

nienna73 (approved):
I like the uses of `@property` in here and yay for tests!

CreeDictionary/CreeDictionary/paradigm/panes.py
eddieantonio: This is the super class that other classes inherit from and overwrite the property; I changed it so that superclass raises `NotImplementedError` so that subclasses can provide a clear implementation!

PR #768, opened 2021-04-30
feat: static paradigms (backend implementation)

eddieantonio:
# What's in this PR?

Added the `ParadigmManager` that loads up _static_ paradigms. Then ask the `ParadigmManager` for a paradigm for a demonstrative pronoun or a personal pronoun, and it will give you that paradigm with the appropriate wordforms!

 - added: static pronoun layouts recovered from #514
 - changed: move static layouts to `res/layouts/static` directory
 - changed: **static layouts must have valid analyses**; so no more hard-coding forms in the layouts

 - refactor: `LiteralCell` -> `WordformCell`
 - refactor: `InflectionCell` -> `InflectionTemplate`
 - refactor: extract superclass `Paradigm` from `ParadigmTemplate`.

 - added: ParadigmPane/Row/Cell classes gained some new methods:
     * `.contains_wordform()` queries whether the paradigm contains the given wordform
     * `.fill()` given a mapping of analyses to one or more wordforms, this returns a `Paradigm` with all `InflectionTemplates` replaced with `WordformCell`s
 - added: `ParadigmTemplate.fill()`, however `Paradigm` instances cannot be filled (they _should_ only have `WordformCells` — not `InflectionTemplate`s!)

Importantly, **the paradigm classes do not care who generates the actual wordforms**. That is, the paradigm pane class hierarchy does **not** know what an FST is!

 - added: `ParadigmManager`
    * responsible for loading the layouts and using an FST to fill in the layouts with wordforms!
 - added: `default_paradigm_manager()`: to access... the default `ParadigmManager`!

 - changed: old `ParadigmFiller` warns instead of crashes if there is a file found that does not match a known wordclass

# What will be done in future PRs?

 * the `default_paradigm_manager()` will be used in the "word detail" view to get paradigms for the pronouns
 * there will be an end-to-end test that will  try to visit the paradigm of... wîya and.. anima... and it should find a paradigm with all their friends!
 * HTML templates that use the new paradigm classes!

Obsoletes: #514


andrewdotn (approved):
Looks great!

PR #773, opened 2021-05-04
change: a few tweaks and fixes for styling

eddieantonio:
# What's in this PR?

 - refactor: CSS class `.linklike` is now `.link`
 - fix: **visited links in the footer are no longer purple**!
 - fix: use correct CSS for when a column label is too wide (although, frankly, I don't recall this being a problem...)
 - change: make column headers **bold**
 - change: make each pane its own `<tbody>` because semantics
 - change: make the personal pronoun layout use two panes!


andrewdotn (approved):
What will the new link footer colour be?

eddieantonio:
> What will the new link footer colour be?

The `:visited` colour will be white, which is what it was _supposed_ to be. But on the production site right now it shows up like this:

purple on maroon link

Not ideal!

PR #774, opened 2021-05-04
Bulk homograph disambiguation for search results

andrewdotn:
One reason the current search is slow is that homograph disambiguation does
at least one query—and often multiple duplicate ones—for every result
entry. Replace all that with doing a single query.

This stopgap should help us until we get stable keys into the original
dictionary database sources.

We only present links for lemmas, so filtering that down helps a bit too.

### DIFF
--- a/CreeDictionary/API/models.py
+++ b/CreeDictionary/API/models.py
@@ -93,19 +93,25 @@ class Wordform(models.Model):
         tags = [FSTTag(t) for t in fst_tag_str.split("+")]
         return LABELS.emoji.get_longest(tags)
 
-    @cached_property
+    # We do our own caching instead of using @cachedproperty so that we can
+    # disambiguate homographs in bulk when displaying search results
+    _cached_homograph_disambiguator: Optional[str]
+
+    @property
     def homograph_disambiguator(self) -> Optional[str]:
         """
         :return: the least strict field name that guarantees unique match together with the text field.
             could be pos, inflectional_category, analysis, id or None when the text is enough to disambiguate
         """
-        homographs = Wordform.objects.filter(text=self.text)
-        if homographs.count() == 1:
-            return None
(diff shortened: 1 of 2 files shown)


CreeDictionary/API/search/core.py
```diff
@@ -42,24 +42,22 @@ def sorted_results(self) -> list[types.Result]:
...
-                "lemma__definitions__citations",
-                "definitions__citations",
-            )
-        except AttributeError:
-            # This happens rarely, but is reproducible with the test suite:
-            # “'RelatedManager' object has no attribute 'citations'.” I think
-            # this is a django bug? It tries to look up a field from Definition
-            # on a DictionarySource object. Passing on the exception just means
-            # that some searches will run more slowly. To revisit when upgrading
-            # to Django 3.
```
eddieantonio, line 57: So this bug no longer exists?

CreeDictionary/API/models.py
```diff
@@ -93,19 +93,25 @@ def get_emoji_for_cree_wordclass(self) -> Optional[str]:
         tags = [FSTTag(t) for t in fst_tag_str.split("+")]
         return LABELS.emoji.get_longest(tags)
 
-    @cached_property
+    # We do our own caching instead of using @cachedproperty so that we can
+    # disambiguate homographs in bulk when displaying search results
```
eddieantonio, line 97:
```suggestion
    # We do our own caching instead of using @cachedproperty so that we can
    # disambiguate homographs in bulk when displaying search results
```

👍🏼

eddieantonio (approved):
LGTM!

CreeDictionary/API/search/core.py
andrewdotn: I’m not 100% sure the bug is gone, but it used to be triggered during the unit tests. After upgrading Django that doesn’t happen anymore. That’s good enough for me.

PR #776, opened 2021-05-05
Docker CI improvements

andrewdotn:
- Post-run, the docker caching stuff was timing out after 40 minutes with
    a full disk, so remove that for now
  - Use the registry API to update the tag instead of pulling then pushing


eddieantonio (approved):
Aaaaaa, this is awesome! LGTM!

PR #778, opened 2021-05-07
feat: add explict paradigm field on Wordforms

eddieantonio:
# What's in this PR?

Instead of trying to guess if a wordform belongs to a known static paradigm, we will let that be an **explicit field** in the wordform, as this is linguistic data.

- change: a few refactorings to `Wordform`
- added: `Wordform.paradigm` field
- added: a data migration that gives existing wordform explicit paradigm fields

*Part one of two in improving the static paradigms!*


CreeDictionary/CreeDictionary/migrations/0003_add_wordform_paradigms.py
```diff
@@ -0,0 +1,64 @@
...
+    ("ôma", "PRON"): "demonstrative-pronouns",
+    ("ôhi", "PRON"): "demonstrative-pronouns",
+    ("anima", "PRON"): "demonstrative-pronouns",
+    ("anihi", "PRON"): "demonstrative-pronouns",
+    ("nêma", "PRON"): "demonstrative-pronouns",
+    ("nêhi", "PRON"): "demonstrative-pronouns",
+    # Inanimate/Obviative inanimate demonstratives
+    ("ôhi", "PRON"): "demonstrative-pronouns",
+    ("anihi", "PRON"): "demonstrative-pronouns",
+    ("nêhi", "PRON"): "demonstrative-pronouns",
```
eddieantonio, line 29: @UAlbertaALTLab/crk-fst I would appreciate a review to make sure I didn't missing any wordforms that should belong to the given paradigms. These are the basic demonstrative pronouns.

C.f., https://plainscree.atlas-ling.ca/grammar/words/pronouns/demonstrative-pronouns/#:~:text=Inanimate-,demonstrative,-Pronouns

atticusha: do you want emphatic/contrastive pronouns like "nista" etc?

andrewdotn (commented):
About how long should this migration take to run in production?

CreeDictionary/CreeDictionary/migrations/0003_add_wordform_paradigms.py
eddieantonio: I think we'll have a different paradigm for nîsta/kîsta/wîsta. Right now, we only have paradigms for "basic" demonstratives.

aarppe: And we'd want to check that the FST analyzes those properly as well.

eddieantonio:
> About how long should this migration take to run in production?

The data migration (adding explicit paradigms) runs quite quickly on my laptop using a copy of the production database (<2s). It's adding the field that takes a while — I neglected to measure it, but it takes about 10 seconds to do the `ALTER TABLE` to add the paradigm

CreeDictionary/CreeDictionary/migrations/0003_add_wordform_paradigms.py
eddieantonio: The FST gets _funky_ with pronouns. There's a lot of inconsistency and duplicate results. Do the mappings above **look correct** such that I can merge this and get the first draft static paradigms working?

aarppe:
My take on this is that the major ambiguity concerns what we specify as the lemma/baseform for pronouns. Now in some cases the lemma is the wordform (personal pronouns); in some other cases the lemma may be the most prototypical word-form for some pronoun set (demontratives). The latter might be based on a potentially easier way to correct mismatches in grammar checking, but that's way into the future. I suspect that the former strategy may be the better one overall, but I'll want to sleep on it.

aarppe:
I discussed the personal pronouns today with Arok, and using the long vowel before -y- is an Albertan specialty, in contrast to SRO where the decision was to always code -i- as short before -y-, effectively allowing -y- to lengthen the vowel. Consequently, I'll move the nîya, kîya, etc. forms from `pronouns.lexc` to `non_standard_forms.lexc`, where they below. Subsequently, they will still be recognized, but not generated.

PR #781, opened 2021-05-07
Cosine vector distance search

andrewdotn:
This adds the machinery for doing cosine vector distance search. Currently it is only accessed by a fancy query, pending a PR to add a weighted ranking function.


CreeDictionary/API/search/cvd_search.py
```diff
@@ -0,0 +1,67 @@
...
+    # better homonym handling.
+    wordform_results = Wordform.objects.filter(
+        text__in=set(wf["text"] for wf in wordform_queries)
+    )
+
+    # Now match back up
+    wordforms_by_text = {
+        text: list(wordforms)
+        for text, wordforms in itertools.groupby(wordform_results, key=lambda x: x.text)
+    }
```
nienna73, line 56: I didn't know you could make a dictionary inline like this!

CreeDictionary/API/search/runner.py
```diff
@@ -26,4 +37,7 @@ def search(*, query: str, include_affixes=True, include_auto_definitions=False):
         do_source_language_affix_search(search_run)
         do_target_language_affix_search(search_run)
 
+    if cvd_search_type != CvdSearchType.OFF:
+        do_cvd_search(search_run)
```
nienna73, line 41: Is there a truthy test that could be done here instead? As in `cvd_search_type == blah` instead of `!=`?
It's a small thing, I just find truthy statements easier to read.

CreeDictionary/API/search/types_test.py
```diff
@@ -0,0 +1,27 @@
...
+def test_result_adding_cvd():
+    def make_wf(text: str = "foo"):
+        ret = Wordform(text=text, is_lemma=True)
+        ret.lemma = ret
+        return ret
+
+    r = Result(make_wf(), query_wordform_edit_distance=1)
+    assert r.cosine_vector_distance is None
+
+    # set CVD if cur is None
```
nienna73, line 14: What is `cur`?

CreeDictionary/CreeDictionary/templates/CreeDictionary/query-help.html
```diff
@@ -1,49 +1,57 @@
 {% extends 'CreeDictionary/base.html' %}
-
+<!-- -->
 {% load creedictionary_extras %}
-
+<!-- -->
```
nienna73, line 4: Why the empty comments?

CreeDictionary/CreeDictionary/templates/CreeDictionary/query-help.html
```diff
@@ -1,49 +1,57 @@
...
+        Use cosine vector distances. The <kbd>retrieval</kbd> setting augments search
+        results using cvd lookups but does not adjust ranking; <kbd>exclusive</kbd>
+        shows only cvd-generated results, for test and debugging purposes.
+    </dd>
+
+    <dt>verbose: (boolean)</dt>
+    <dd>Show a JSON dump of available search ranking features</dd>
+  </dl>
+  {% endif %}
+</section>
```
nienna73, line 54: Yay for having a help page!

CreeDictionary/cvd/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+def definition_vectors():
+    try:
+        return _load_vectors(definition_vectors_path())
+
+    except FileNotFoundError:
+        raise DefinitionVectorsNotFoundException
+
+
+# https://stackoverflow.com/a/48027864/14558
+def uniq(l: list) -> list:
```
nienna73, line 46:
```suggestion
def unique(l: list) -> list:
```

CreeDictionary/CreeDictionary/templates/CreeDictionary/query-help.html
andrewdotn: I have been using [prettier](https://prettier.io) as an auto-formatter for my Django templates. It’s designed for regular HTML, and it mostly works, so long as I add empty comments between things that should be on their own lines.

CreeDictionary/cvd/__init__.py
andrewdotn: It’s named after the [unix shell command](https://linux.die.net/man/1/uniq).

CreeDictionary/cvd/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+    return uniq(ret)
+
+
+def _warn(word, msg, already_warned):
+    if (
+        logger.isEnabledFor(logging.DEBUG)
+        and already_warned is not None
+        and word not in already_warned
+    ):
+        already_warned.add(word)
```
nienna73, line 98: Is this using Python pointer magic of some sort?
I'm confused as to how this actually gets stored/updated

CreeDictionary/cvd/__init__.py
nienna73: That doesn't mean unix is right...

nienna73 (approved):
A couple small points of confusion that are likely just that (me being confused about small things), otherwise looks good!

CreeDictionary/cvd/__init__.py
andrewdotn: If you pass a set to a function, that function gets a reference to the set, not a copy of the set, so it can update that set for you. In this case, some calling function *might* call `extract_keyed_words` a bunch of times and want to keep state across those calls.

I *could* rewrite this bit to do a better job of tracking build stats if you really wanted me to.

nienna73: That makes sense. Passing by reference has never been my forte, so thanks for explaining that!

CreeDictionary/cvd/__init__.py
eddieantonio: 😂  :joy: :joy: kitâpwân!

CreeDictionary/search_quality/management/commands/featuredump.py
```diff
@@ -55,7 +58,7 @@ def handle(self, *args, **options):
                 for i, r in enumerate(results):
                     ret = r.features()
                     ret["query"] = query
-                    ret["wordform_text"] = r.wordform.textq
```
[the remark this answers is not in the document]
andrewdotn: The method was not being type-checked at all, and mypy could not figure out that `def search(): search_run = SearchRun(...); …; if … : return search_run; …; return search_run` returns a `SearchRun`; but even then it doesn’t know how to type-check Django model objects, so `wordform.blah` is allowed.

CreeDictionary/cvd/management/commands/builddefinitionvectors.py
```diff
@@ -0,0 +1,94 @@
...
+    if path:
+        file = open(path, "w")
+        try:
+
+            def log(*args, **kwargs):
+                print(*args, **kwargs, file=file)
+
+            yield log
+        finally:
+            file.close
```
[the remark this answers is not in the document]
andrewdotn: Yikes! Thanks.

CreeDictionary/cvd/management/commands/builddefinitionvectors.py
```diff
@@ -0,0 +1,94 @@
...
+    vector_for_keys,
+    definition_vectors_path,
+)
+from cvd.definition_keys import definition_to_cvd_key
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = """Create a vector model from current definitions"""
```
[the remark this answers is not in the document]
andrewdotn: Yeah.

PR #783, opened 2021-05-07
change: render paradigm declared in the paradigm

eddieantonio:
# What's in this PR?

The dictionary now uses `Wordform.paradigm` to decided whether it should fetch/render a static paradigm.

- changed: **static paradigms are now explicitly declared**
- changed: `WordformCell`s can now be parsed from layouts
- added: `ParadigmLayout.as_static_paradigm()` returns a paradigm that can be rendered without any further paradigm filling
- added `ParadigmManager.static_paradigm_for()` returns a static paradigm given a name, e.g., from `Wordform.paradigm`
- removed: `ParadigmManager.paradigm_for(analysis)` — we no longer need the analysis to determine which paradigm to fetch

Follow up to #778.

### DIFF
--- a/CreeDictionary/CreeDictionary/paradigm/panes.py
+++ b/CreeDictionary/CreeDictionary/paradigm/panes.py
@@ -145,7 +145,7 @@ class ParadigmLayout(Paradigm):
         return Paradigm(pane.fill(forms) for pane in self.panes)
 
     def as_static_paradigm(self) -> Paradigm:
-        f"""
+        """
         Returns a Paradigm that can be rendered without having to explicitly fill it.
         
         :raises ParadigmIsNotStaticError: when called on a dynamic paradigm
@@ -436,7 +436,7 @@ class WordformCell(Cell):
         return self.inflection == wordform
 
     def fill_one(self, forms: dict[str, Collection[str]]) -> Cell:
-        # No need to cell that already has contents!
+        # No need to fill a cell that already has contents!
         return self
 
     def __str__(self) -> str:
--- a/CreeDictionary/CreeDictionary/paradigm/panes.py
+++ b/CreeDictionary/CreeDictionary/paradigm/panes.py
@@ -11,10 +11,9 @@ from functools import cached_property
 from itertools import zip_longest
 from typing import Collection, Iterable, Optional, TextIO
 
-from more_itertools import first
+from more_itertools import first, ilen
 
 logger = logging.getLogger(__name__)
-
 
 class ParseError(Exception):
     """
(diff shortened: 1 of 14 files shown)


CreeDictionary/CreeDictionary/paradigm/panes.py
```diff
@@ -139,6 +144,17 @@ def fill(self, forms: dict[str, Collection[str]]) -> Paradigm:
         """
         return Paradigm(pane.fill(forms) for pane in self.panes)
 
+    def as_static_paradigm(self) -> Paradigm:
+        f"""
```
andrewdotn, line 148: Is the `f` intentional?
```suggestion
        """
```

andrewdotn (dismissed):

andrewdotn (commented):
Looks good, apart from tests failing.

CreeDictionary/CreeDictionary/paradigm/panes.py
eddieantonio: Nope 🙃

eddieantonio:
> Looks good, apart from tests failing.

I can't replicate locally! 😭 :sob: 😭 

Looks like this is a Monday problem :/

andrewdotn:
It’ll fail locally if you remove `CreeDictionary/test_db.sqlite3`. Remember that it’s cached on your dev machine, but recreated every time on CI.

Maybe try to set the required `Wordform.paradigm` values in `ensuretestdb`?

Oh! if you’re counting on the migration to set stuff, then it won’t happen on a new import…

eddieantonio:
> It’ll fail locally if you remove `CreeDictionary/test_db.sqlite3`. Remember that it’s cached on your dev machine, but recreated every time on CI.
> 
> Maybe try to set the required `Wordform.paradigm` values in `ensuretestdb`?
> 
> Oh! if you’re counting on the migration to set stuff, then it won’t happen on a new import…

Ahhhh! Thanks for the insight! I will try that! ...on Monday, probably because Friday brain 😅

andrewdotn:
If we re-import the production database, I believe that will also leave us without these new paradigms.

That’s ok because there are already some manual steps for reimports, like redoing auto-translations, and building definition vectors will soon be added to that list.

But if you are feeling ambitious about this particular PR, maybe add a post-import method called by both xmlimport and ensuretestdb?

andrewdotn (approved):

eddieantonio:
> But if you are feeling ambitious about this particular PR, maybe add a post-import method called by both xmlimport and ensuretestdb?

Exactly what do you mean by "post-import"? Like a simple function call after both, or some kind of signal thingy?

andrewdotn:
Oh, just a simple function call.

nienna73 (approved):

PR #785, opened 2021-05-10
change: remove subparser for xmlimport

eddieantonio:
The `xmlimport` command used to have obligate "subcommands":

 - `import [--wipe-first] path/to/crkeng.xml`

That's it. Those are all the subcommands.

I decided to remove the subcommand parser and make `xmlimport` do what `xmlimport import` used to do. 😄

### DIFF
--- a/CreeDictionary/CreeDictionary/management/commands/ensuretestdb.py
+++ b/CreeDictionary/CreeDictionary/management/commands/ensuretestdb.py
@@ -35,7 +35,6 @@ def import_test_dictionary():
         print("No wordforms found, generating")
         call_command(
             "xmlimport",
-            "import",
             shared_res_dir / "test_dictionaries" / "crkeng.xml",
         )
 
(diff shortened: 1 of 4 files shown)


eddieantonio:
> https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/d1d59d70437e29cb2771174ecb64d514471e392c/CreeDictionary/search_quality/management/commands/runsamplequeries.py#L18

Ah, makes sense, what with Django's default options kind of poluting the `--help` screen:

```

runsamplequeries option:
  --csv-file CSV_FILE
  --max MAX             Only run this many queries
  --shuffle, --no-shuffle
                        Shuffle sample before running, useful with --max
  --append-to-query APPEND_TO_QUERY
                        Append string to every query, useful with fancy queries
  --result-file-path RESULT_FILE_PATH, --query-results-file RESULT_FILE_PATH
                        Full path to write results to, default=/Users/santoseadmin/Work/cree-intelligent-
                        dictionary/CreeDictionary/search_quality/sample_results/query-results.json.gz
  -n RESULT_FILE_NAME, --result-file-name RESULT_FILE_NAME
                        Base name to use for results file
```

🤩

PR #790, opened 2021-05-12
refactor: use relative imports in CreeDictionary package

eddieantonio:
# What's in this PR?

Converts all imports within `CreeDictionary` to relative imports, plus
other minor touch-ups.

# Why is this PR?

The `CreeDictionary` package should be split up into two:

 - `itwewina_site`: is a Django **site**, containing `settings.py`, `wsgi.py`, the _root_ `urls.py`
 - `cree_dictionary`: is a Django **app**, containing a slimmed down `urls.py` and all
   the other content (e.g., templates, Cree-specific code).

This way we can make way for Arapaho and other languages

### DIFF
--- a/CreeDictionary/CreeDictionary/migrations/0003_add_wordform_paradigms.py
+++ b/CreeDictionary/CreeDictionary/migrations/0003_add_wordform_paradigms.py
@@ -1,6 +1,6 @@
 from django.db import migrations
 
-from CreeDictionary.ensure_data import (
+from ..ensure_data import (
     set_paradigm_for_demonstrative_and_personal_pronouns,
 )
 
--- a/CreeDictionary/CreeDictionary/hfstol.py
+++ b/CreeDictionary/CreeDictionary/hfstol.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-
 """
 Run finite-state transducer analyzer and generator
 """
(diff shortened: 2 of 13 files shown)


CreeDictionary/CreeDictionary/migrations/0003_add_wordform_paradigms.py
```diff
@@ -1,6 +1,6 @@
 from django.db import migrations
 
-from CreeDictionary.ensure_data import (
+from ..ensure_data import (
```
nienna73, line 3: This one only has two dots but the others have three 🤔

nienna73 (approved):
Nice!

andrewdotn (changes requested):
Requires further discussion.

eddieantonio:
> Requires further discussion.

Could you please _state_ what the further discussion is in the comments? I believe it's because we both started refactoring the same imports (c.f., andrewdotn:import-fix). Please state this in your next review comments.

CreeDictionary/CreeDictionary/migrations/0003_add_wordform_paradigms.py
eddieantonio: The other relative imports are in `management/commands/` whereas this is in `migrations`. Each dot corresponds to a directory out. This is just one directory out.

PR #793, opened 2021-05-13
Add FST tool for admins to do analyses

andrewdotn:


andrewdotn:
Screen Shot 2021-05-13 at 12 11 01 PM

src/index.js
```diff
@@ -83,6 +83,8 @@ document.addEventListener('DOMContentLoaded', () => {
     setSubtitle(getEntryHead())
     setupAudioOnPageLoad()
     setupParadigmSizeToggleButton()
+  } else if (route === '/query-help' || route == '/admin/fst-tool') {
```
eddieantonio, line 86: Honestly, the router needs to be fixed here in a future PR :/

CreeDictionary/CreeDictionary/templatetags/creedictionary_extras.py
```diff
@@ -52,3 +52,10 @@ def url_for_query_tag(user_query: str) -> str:
         /search?q=w%C3%A2pam%C3%AAw
     """
     return url_for_query(user_query)
+
+
+@register.filter()
+def kbd_text_query_link(text):
```
eddieantonio, line 58: I like this filter!

Could you please a docstring to explain why this exists (i.e., its usage in the FST tool page)?

eddieantonio (approved):
LGTM! I would just appreciate a docstring for the new filter.

src/index.js
andrewdotn: Yeah, I just added this because Cypress was correctly complaining about the exception showing up in the JS console whenever these pages were being visited.

PR #794, opened 2021-05-13
Bump hfst_optimized_lookup version

andrewdotn:
This will fix the issue with navajo/avocado analyzing as ‘na’ and ‘a’.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -168,10 +168,10 @@
         },
         "hfst-optimized-lookup": {
             "hashes": [
-                "sha256:a6c9fa5329fc5d80b5f38023b37ccb1669b8805d49877bee298ad541fb8321f2"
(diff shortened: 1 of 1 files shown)


eddieantonio (approved):
LGTM! I'm glad we solved this bug!

aarppe:
I could imagine some logic here that one might have wanted to provide some match based on the prefix of word - but as far as I know this wasn't a deliberate design decision that we would have explicitly discussed, which is an example of the type of behavior we should have as little as possible, if at all.

andrewdotn:
Yeah, this is one case where it turns out that it’s pretty easy to tell whether the query is in Cree or English: some letters only exist in English, and not in Cree. We shouldn’t try to show Cree-language matches against queries which are definitely in English.

PR #795, opened 2021-05-13
Run black on migrations

andrewdotn:
Although the migration files were excluded from checks, something in
PyCharm sometimes runs black on these files anyway. I’m not sure how to
turn that off, so it seems easier to just have black format migration files
as the usual practice.


eddieantonio (commented):
LGTM!

PR #796, opened 2021-05-14
Greatly reduce test database size

andrewdotn:
There was a bug in the code to produce the test database. It would include
all words from the full dictionary where a *substring* of the headword
analysis occurred in `test_db_words.txt`. With this patch, the test
database will only include words from the full dictionary where the *lemma*
is in `test_db_words.txt`, based on equality and not substring matches.

This should speed up CI a bit.


eddieantonio (approved):
LGTM! Love a PR that removes 55 lines!

PR #797, opened 2021-05-14
Update directory structure documentation

andrewdotn:


andrewdotn:
@eddieantonio Two changes since we went over this yesterday:

  - morphodict is a package that can contain multiple django apps, and let’s
    call the main one, with `models.py`, `morphodict.dictapp` or something

  - `docker` folders inside each dictionary application directory

docs/directory-structure.md
```diff
@@ -1,241 +1,205 @@
...
+        │   └── CreeDictionary/     # this goes away too
+        │       ├── __init__.py
+        │       ├── models.py
+        │       ├── views.py
+        │       └── …
+        │
+        ├── morphodict/             # python package for language-independent code
+        │   ├── __init__.py
+        │   ├── cvd/
+        │   ├── dictapp/            # django app with primary database tables
```
eddieantonio, line 75: @dwhieb What do you think about calling this [Django application ](https://docs.djangoproject.com/en/3.2/ref/applications/) `lexicon`?

This will affect the default name of database tables, e.g.,

```sql
SELECT text FROM lexicon_wordform;
```

or 

```sql
SELECT publisher FROM lexicon_dictionarysource;
```

---

It will also affect how the app is enabled:

```sql
# src/crkeng/site/settings.py
INSTALLED_APPS = [
   ...,
   "morphodict.lexicon.apps.LexiconAppConfig"
]
```

---

This will be visible in the Django Admin:

Screenshot of Django Admin with  “LEXICONI” listed as the name of the primary application

eddieantonio (approved):
A few small nitpicks, but I like it for now! I'd appreciate @nienna73's  and @dwhieb's thoughts on this though.

docs/directory-structure.md
```diff
@@ -1,241 +1,205 @@
...
+  - Allow generalization to more languages than Plains Cree, and
+
+  - Address some other issues that have arisen with the source code layout.
+
+## The proposal
+
+### High-level decisions / assumptions
+
+  - In the file system, dictionary applications such as itwêwina are named
+    `sssttt`, where `sss` and `ttt` are three-letter [ISO 639-3 language
```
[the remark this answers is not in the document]
andrewdotn: Yup. I’ve added a link to `glossary.md` which will be accurate once the ‘source language’ and ‘target language’ definitions of https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/pull/756 are merged in.

docs/directory-structure.md
andrewdotn: I think ‘lexicon’ is a much, *much* better than <sub><sup>‘dictapp’</sup></sub>. Thanks!

eddieantonio (approved):

docs/directory-structure.md
```diff
@@ -1,241 +1,206 @@
...
+        │   │   ├── settings.py
+        │   │   └── urls.py
+        │   └── frontend/           # Not a python package
+        │       ├── foo.js
+        │       └── foo.css
+        │
+        ├── arpeng/
+        ├── crkfra/
+        ├── cr_shared               # for code and resources shared between Cree dialects
+        └── srseng/
```
nienna73, line 132: Is there a plan for where tests should live?
Does each individual service have its own tests as well as a larger set of tests for all the services?

nienna73: I see the explanation below now. I'm still wondering if each project should have a dedicated `tests` folder?

nienna73 (approved):
Just the one comment about where tests live, otherwise this looks great!

docs/directory-structure.md
andrewdotn: I added some more comments about testing. We were hoping to not have any separate `tests` folders, but to have the tests mixed in, but if you really like having `tests` folders we can definitely talk about that.

Some of the testing stuff is pretty speculative at this point because it looks like it could turn out to be a bit of a pain to wrangle pytest to run tests across multiple django projects at once. If there turns out to be some way to organize the tests that makes running them much easier, that could end up overriding any other preferences.

nienna73: That should be fine. I think I see the thought process behind deciding where the tests go.

docs/directory-structure.md
dwhieb: I agree _lexicon_ is a big improvement!

If you want to get super precise, _dictionary_ could be used to refer to the Django app, and _lexicon_ to the actual database. You can think of a dictionary as a particular view on a lexicon—a lexicon can be presented in different ways. I'm totally fine with just using _lexicon_ though.

docs/directory-structure.md
```diff
@@ -1,241 +1,213 @@
...
+        │   └── CreeDictionary/     # this goes away too
+        │       ├── __init__.py
+        │       ├── models.py
+        │       ├── views.py
+        │       └── …
+        │
+        ├── morphodict/             # python package for language-independent code
+        │   ├── __init__.py
+        │   ├── cvd/
+        │   ├── lexicon/            # django app with primary database tables
```
dwhieb, line 76: Will all the language data live in a single database, or will we run separate database instances for each language?

I like having a single database, because it forces us to structure language data the same way across languages.

docs/directory-structure.md
```diff
@@ -1,241 +1,213 @@
...
+        │   ├── res/                # Resources go here
+        │   │   ├── altlabels.tsv
+        │   │   ├── dictionaries/
+        │   │   ├── fst/
+        │   │   └── layouts/
+        │   ├── site/               # Django project
+        │   │   ├── __init__.py
+        │   │   ├── settings.py
+        │   │   └── urls.py
+        │   └── frontend/           # Not a python package
```
dwhieb, line 106: Some more structure here would be helpful, depending on how complex the frontend is. Is there an expectation that the entire frontend will only have one JS and one CSS asset per project? If so, these should probably be given a naming convention of `app.js` and `app.css` or some such. Or are we using a component structure, like below?

```
frontend/
└── component
    ├── component.css
    ├── component.js
    └── component.test.js
```

dwhieb (approved):
Sorry for the delay! Made a couple comments, but looks good to me!

docs/directory-structure.md
andrewdotn: We will have separate databases for each language pair so that time-consuming/risky database operations for new languages we’re experimenting with won’t affect production sites like itwêwina.

But, the code to access the database will be shared, and thanks to Django, each database will have exactly the same *schema* for the same tables.

docs/directory-structure.md
andrewdotn: I’ve updated the docs with a bit more info on this. In short:

  - There already is a bit of structure to our frontend stuff, with multiple JS and CSS files
  - That seems to be working, so we’ll keep it for now
  - Although language-specific stuff should move to language-specific directories

PR #798, opened 2021-05-14
Mount FSTs from filesystem

andrewdotn:
The foma FSTs are checked in with git-lfs. The docker build currently doesn’t do an LFS checkout. There are a few different ways to tackle this; I’m leaning towards this quick fix on the theory that, if a file is big enough to warrant git-lfs, we probably don’t want to put copies into every docker image either.


eddieantonio (approved):
LGTM!

I mean, there is the argument for "reproducable images", but our primary use case of Docker images is not necessarily for its _long-term_ reproducability.

PR #800, opened 2021-05-14
Add `CreeDictionary/__init__.py`

andrewdotn:


andrewdotn:
There’s a bit more background for this in [this commit message](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/pull/800/commits/07672a3c8e57ab1ebe0c9595d8be44e80bcaaa79).

CreeDictionary/CreeDictionary/settings.py
```diff
@@ -63,13 +63,13 @@
...
     # Internal apps
     # TODO: our internal app organization is kind of a mess 🙃
-    "API.apps.APIConfig",
-    "CreeDictionary.apps.CreeDictionaryConfig",
-    "cvd",
-    "search_quality",
-    "phrase_translate",
-    "morphodict.apps.MorphodictConfig",
-    "DatabaseManager",
+    "CreeDictionary.API",
```
eddieantonio, line 66: I love how the comment above was not deleted 💀 💀 💀 

(do not delete that comment; I still agree with it, haha!)

CreeDictionary/manage.py
```diff
@@ -5,9 +5,16 @@
 
 import os
 import sys
+from pathlib import Path
+
+# sys.path[0] is initialized to the directory containing the script, which
+# is too deep for our purposes.
+sys.path[0] = os.fspath(Path(sys.path[0]).parent)
```
eddieantonio, line 12: 👍🏼

pyproject.toml
```diff
@@ -13,24 +13,12 @@ skip = [
...
-[project_paths]
-# A place to store paths to different components in our project.
-# Here they are all written relative to project root.
-# we have get_project_path function in CreeDictionary.utils.project_path
-# module that resolves to these paths
-# TODO: use `pip install project-paths` (requires Python 3.8+)
+DJANGO_SETTINGS_MODULE = "CreeDictionary.CreeDictionary.settings"
 
-# the file we write documentation for directories
-dir_doc_file = "docs/directory-structure.md"
```
eddieantonio, line 36: Haha, it's probably best to clean this all up!

pyproject.toml
```diff
@@ -13,24 +13,12 @@ skip = [
...
+django_find_project = false
+python_paths = "."
```
eddieantonio, line 24: Could you please add a comment about why this was added? I think it has something to do with the new pytest-pythonpath dependency?

eddieantonio (approved):
LGTM! I would just like one comment to clarify the use of `pytest-pythonpath`

pyproject.toml
andrewdotn: I have added a comment that explains a bit but it seems like setting sys.path is just something that pipenv/pytest/mypy all handle badly.

This page I just found leads me to believe the problems I was running into *might* by fixed by ensuring we have `__init__.py` files in all the directories that need them, and only the ones that need them: 
https://docs.pytest.org/en/documentation-restructure/background/pythonpath.html

I haven’t been spending much time on this at the moment because I expect it’ll all have to change soon anyway when there are multiple `manage.py` scripts and associated django projects to test together.

PR #802, opened 2021-05-18
refactor: move README to right place

eddieantonio:
This README got stranded from the folder it was describing.

Note, I'd love to get rid of `CreeDictionary/static/README.md` but that will require considerable thinking, so that's left for a future PR!


eddieantonio:
Closed in favor of #803

PR #805, opened 2021-05-18
refactor: JavaScript

eddieantonio:
# What is in this PR?

In general, I tried to tidy up all of the JavaScript so that it's easier to understand and maintain.

 * move a bunch of JavaScript files to `frontend/js`
 * extract paradigm stuff to `frontend/js/paradigm.js`
 * fix a regression that nobody made a bug report about, but I noticed, and it was bothering me! 😠


frontend/js/paradigm.js
```diff
@@ -0,0 +1,159 @@
...
+  const jsonScriptElement = document.getElementById(id)
+  if (jsonScriptElement) {
+    return JSON.parse(jsonScriptElement.textContent)
+  } else {
+    return undefined
+  }
+}
+
+/**
+ * https://stackoverflow.com/a/11654596
```
andrewdotn, line 123: But but but 😢 … can’t we use `URLSearchParams` like https://stackoverflow.com/a/11654596 instead of this line noise ???

andrewdotn (approved):
Having all the `.js` files in a `js/` folder *except* for `index.js` feels a bit off, but this is definitely an improvement overall.

frontend/js/paradigm.js
andrewdotn: Oops meant https://stackoverflow.com/a/41542008/14558 was distracted by trying and failing to come up with a sufficient expression of horror at that … perl … monstrosity

andrewdotn: Thank you!

PR #811, opened 2021-05-20
fix: npm start uses correct crkeng runserver command

eddieantonio:
I use `npm start` to develop frontend things. This used a now-deleted command to start the Django server. Now we use the right command! 😄


andrewdotn (approved):

andrewdotn:
My bad!

PR #813, opened 2021-05-20
crkeng.app: Django application for purposes of overriding templates and static files

eddieantonio:
By default, Django's template engine/loader and staticfiles app search for **application-specific** files by traversing apps in the `INSTALLED_APPS` list and picking the **first** it finds.

So it's beneficial for an intelligent dictionary instance to not only have a Django project (e.g., `src/crkeng/site/settings.py`) but also an _application_ (in the Django parlance) that goes first in `INSTALLED_APPS`. This PR proposes where it should go.

Note: I feel a little dirty about putting the app config in `__init__.py`. Django would like you to create `apps.py` within the Python package containing the app, but that would lead to this path: `crkeng.app.apps.CrkEngAppConfig` which is just a bit too redundant for me.


eddieantonio:
After reading “[Configuring applications](https://docs.djangoproject.com/en/3.2/ref/applications/#configuring-applications)”, I think I _might_ want to move the `AppConfig` subclass to `src/crkeng/app/apps.py`, even though that is not _strictly_ necessary, but Django makes ✨ assumptions ✨ that require workarounds, like explicilty specifying the `AppConfig.path`.

To reduce the redundency, the app could be registered like so:

```python
# src/crkeng/site/settings.py
INSTALLED_APPS.insert(0, "crkeng.app")
```


¯\\\_(ツ)\_/¯

eddieantonio:
@andrewdotn notified me that _Django does not need an `AppConfig` for each app!_ So I deleted it. So adding an app literally requires two things:

 - creating an empty file `app/__init__.py`
 - adding it to the settings module, e.g.,

```python
# srseng/site/settings.py
INSTALLED_APPS.insert(0, "srseng.app")
```

nienna73 (approved):
LGTM

PR #814, opened 2021-05-25
fix: use both +Ipv and IPV in crk.altlabel.tsv

eddieantonio:
~~Fixed the failing test by using the form of `+IPV`/`+Ipv` that is actually used within the application code: namely, `+Ipv`~~

**EDIT**: So `+Ipv` is the FST-style tag and _IPV_ is the _inflectional category_. N.B.: the FST does not produce a tag for `+Ipv`; it's purely invented in the database importer within itwêwina.

Please see #815


andrewdotn (approved):

andrewdotn:
BTW, which test was failing??

eddieantonio:
@andrewdotn this test:

https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/a5f9bb8c19f8de1cda6a26f1d2c5c58a4b909de1/src/CreeDictionary/tests/API_tests/model_test.py#L30-L33

Because `Ipv` was deleted from `crk.altlabel.tsv`, there was no "plain English" relabelling of `pê-+Ipv`. Note, in the current state, _both_ `IPV` and `Ipv` are needed in `crk.altlabel.tsv` (see #815)

PR #818, opened 2021-05-26
feat: use dynamic, pane-based paradigms for "full" layouts

eddieantonio:
# What's in this PR?

Full paradigms use the new layouts and look like this:

pipon full layout

 - adds the full, dynamic paradigm layouts, as created by @aarppe
 - edits the dynamic layouts a little bit, to fix some syntax issues
 - adds missing `crk.altlabel.tsv` entry
 - **implements** `ParadigmLayout.dynamic_paradigm_for(lemma, word_class)`
 - changes: when looking at the **full** layouts, uses the new, pane-based dynamic layouts!

# What is NOT in this PR?

(In other words, what needs to be done in future PRs)

 - basic layouts using the new pane-based layout format
 - I think the headers are not working properly for the pane-based layouts **EDIT**: see #820 
 - allow for multiple sizes of layout
 - delete the `convertndstopane` command — the layouts have been converted! **EDIT**: see #819 
 - delete the old style "full" paradigms **EDIT**: see #821
 - refactor/move paradigm stuff to `src/morphodict`


src/CreeDictionary/CreeDictionary/paradigm/manager.py
```diff
@@ -30,21 +32,50 @@ def static_paradigm_for(self, name: str) -> Optional[Paradigm]:
...
+        template2forms = {
+            template: analysis2forms[analysis]
+            for template, analysis in template2analysis.items()
+        }
+        return layout.fill(template2forms)
 
+    @staticmethod
+    def _load_all_layouts_in_directory(path: Path):
         for layout_file in path.glob("*.tsv"):
             layout = ParadigmLayout.loads(layout_file.read_text(encoding="UTF-8"))
```
andrewdotn, line 77: Do you know any reason to explicitly specify the encoding these days?

andrewdotn (approved):
There are a few TODO comments about potentially using `ConcatAnalysis`; personally I wouldn’t bother.

src/CreeDictionary/CreeDictionary/paradigm/manager.py
andrewdotn: Genuinely curious, I never specify this and wonder if I should be.

eddieantonio: It's a habit I use because Python uses the system encoding as the default (yuck!). Although this generally isn't a problem on most modern systems, this can cause problems when the system's locales are b0rked (it happens), current locale is something you don't expect, or when you're running on _legacy systems_ (see: Windows before Windows 10—I was writing code on Windows 7 for my current job, and this line is _necessary_ for that use case 🙃 ).

PR #819, opened 2021-05-26
remove: convertndstopanes

eddieantonio:
The panes have been converted in the parent commits!


src/CreeDictionary/CreeDictionary/management/commands/convertndstopanes.py
```diff
@@ -1,201 +0,0 @@
-"""
-Converts all loaded paradigm templates and outputs them in the pane-style template.
-
-Note: this code is **temporary**! (2021-04-23)
```
andrewdotn, line 4: It’s nice when comments like this end up being true! 🎆

andrewdotn (approved):

PR #820, opened 2021-05-26
fix: header rows for new pane-based paradigm layouts

eddieantonio:


src/CreeDictionary/CreeDictionary/paradigm/panes.py
```diff
@@ -350,14 +355,17 @@ def inflection_cells(self) -> Iterable[InflectionTemplate]:
...
+
     def __eq__(self, other) -> bool:
         if isinstance(other, HeaderRow):
             return self._tags == other._tags
         return False
 
     def __str__(self):
-        tags = "+".join(self._tags)
-        return f"{self.prefix} {tags}"
+        return " ".join(f"{self.prefix} {tag}" for tag in self._tags)
```
andrewdotn, line 368: `self.prefix` will be repeated for every tag?

src/CreeDictionary/CreeDictionary/paradigm/panes.py
```diff
@@ -641,6 +640,30 @@ class ColumnLabel(BaseLabelCell):
...
+    ['1Sg', '3Sg', 'Obv']
+    """
+    splits = re.split(r" +", text)
+    if len(splits) % 2 != 0:
+        raise ParseError(f"Uneven number of space-separated segments in {text!r}")
+
+    tags = []
+    for actual_prefix, tag in pairs(splits):
+        if actual_prefix != prefix:
+            raise ParseError(f"Expected prefix {prefix!r} but saw {actual_prefix!r}")
```
andrewdotn, line 661: some minor optional golfing:

```suggestion
            raise ParseError(f"Expected {prefix=!r} but saw {actual_prefix!r}")
```

andrewdotn (approved):

src/CreeDictionary/CreeDictionary/paradigm/panes.py
eddieantonio: Yes... this is the syntax I came up with and did not fully think out 🙃 

This syntax is consistent with row labels and column labels. It's still not too late to change the syntax if you think it's silly.

PR #823, opened 2021-05-27
refactor: rename lemma_details -> entry_details

eddieantonio:
This view displays the full details of a _dictionary entry_. This _may_ display the paradigm, but it also may not! @dwhieb Is this the correct terminology? `lemma_detail` was definitely wrong!


dwhieb:
Yup, sounds correct! Just to clarify, this view is what you get when you click on a headword, correct? Like this?

https://itwewina.altlab.app/word/w%C3%A2skahikan/?inflectional_category=INM

eddieantonio:
@dwhieb That's correct!

PR #824, opened 2021-05-27
feat: choose paradigm label preference (backend only)

### LINKED ISSUE #743: Extend relabeling choice to paradigm layouts
Extending on #729, apply the choice of relabeling mode to paradigm layouts as well.

eddieantonio:
# What's in this PR?

The backend implementation of choosing relabellings for the paradigm labels only. I decided to cut-off this PR at just the backend because it was getting a bit too big to review.

 - **added**: the `morphodict` Preference "framework" — a convention for storing preferences in cookies
 - **refactor**:  "display mode" uses the preference framework instead of wacky constants
 - **added**: paradigm labels implemented using the preference framework
 - **added**: support for the `{% relabel %}` tag to use the current paradigm label cookie

# What's **not** in this PR?

 - Frontend (e.g., something to click to make it actually work!). This will be in a PR coming soon!
 - Settings page #708 

The frontend _technically_ works as of this PR, but it... needs some work:

Cree paradigm labels beta

Addresses #743


aarppe:
Looks good to me. Was going to ask where the hyphen comes from in **ê-/kâ- itwêwin**, and that appears to be in `crk.altlabel.tsv` what used to be an empty column marker that's been appended to the contents of the previous column. Will remove those separately.

Also, since this is using nêhiyawêwin mode, I presume that is also being released, but is that a separate PR?

src/morphodict/preference.py
```diff
@@ -0,0 +1,77 @@
...
+        if who_asked_us := request.headers.get("Referer"):
+            # Force the browser to refresh the page that issued this request.
+            response = HttpResponse(status=HTTPStatus.SEE_OTHER)
+            response.headers["Location"] = who_asked_us
+        else:
+            # We don't know where to redirect, so send no content.
+            # (also, this probably should never happen?)
+            response = HttpResponse(status=HTTPStatus.NO_CONTENT)
+
+        response.set_cookie(self.cookie_name, value)
```
andrewdotn, line 76: If I’m looking at the right docs, [Django defaults](https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpResponse.set_cookie) to creating a cookie that only lasts until the browser is closed.

Frequent itwêwina users may expect the preference to last a bit longer, e.g., the cookie could be regularly set to expire in two weeks so that someone using the site frequently keeps their preferences. But I know that is more work to set up.

For now maybe just a comment here about what the cookie lifetime is?

andrewdotn (approved):
Looks good.

If there may eventually be many preferences, you might want to have a single `preferences` cookie and a single view to do partial updates of that.

PR #826, opened 2021-05-28
feat: paradigm label frontend (interface)

### LINKED ISSUE #743: Extend relabeling choice to paradigm layouts

eddieantonio:
# What's in this PR?

https://user-images.githubusercontent.com/2294397/120379159-9696f000-c2dc-11eb-843f-8779dcc64367.mov

 - over-engineered `Preference` framework
   - preferences can be **registered** for the site
   - a **context processor** gives uniform access to all registered preferences on the site within templates
   - this is _less_ over-engineered than it was previously. Feel free to roast me on it still ¯\\\_(ツ)\_/¯ 
 - `paradigm-label-selector.html` component, which uses the aforementioned Preference context processor to list all of the preferences
 - tests of the functionality

Fixes #743


frontend/css/styles.css
```diff
@@ -1224,6 +1200,60 @@ a.menu-choice__label:active {
   background-color: var(--menu-selected-bg-color);
 }
 
+/********************************* MENU BAR *********************************/
+
+/**
+ * BLOCK: menu-bar
+ *
+ * Display a whole bunch of options in horizontal menu bar.
```
andrewdotn, line 1208:
```suggestion
 * Display a whole bunch of options in a horizontal menu bar.
```
or
```suggestion
 * Display a whole bunch of options in the horizontal menu bar.
```

src/morphodict/preference/test_context_processors.py
```diff
@@ -0,0 +1,79 @@
...
+    request = HttpRequest()
+    if choice is not None:
+        request.COOKIES[Pets.cookie_name] = choice
+    context_that_likes_dogs = RequestContext(request, {})
+    template = Template(
+        """
+        I like {{ preferences.pets.current_label }}.
+    """
+    )
+    rendered = template.render(context_that_likes_dogs)
```
andrewdotn, line 77: I like the variable name `context_that_likes_dogs`, but now that it’s parametrized, I suggest:

```suggestion
    context_that_likes_label = RequestContext(request, {})
    template = Template(
        """
        I like {{ preferences.pets.current_label }}.
    """
    )
    rendered = template.render(context_that_likes_label)
```

andrewdotn (approved):

aarppe:
Among the various options, I'd call the current "linguistic" one "`show short linguist tags`", and add another "`show long linguistic labels`". Moreover, I'd be partial to using only the long labels, but if we give choice, we might as well give them all. As for the fifth option, "`short linguistic labels`", I just wasn't able to come up with those for all cases.

aarppe:
Also, my general sentiment has been that choosing a labeling option would apply throughout, so not just for 1) the labels for the paradigm rows and columns, but also for 2) the presentation of the linguistic analysis, i.e. `nipâw+V+AI+Ind+1Sg`, as well as 2) the presentation of lexicographical content from the dictionary DB, i.e. `VAI` and `VAI-1`.

Maybe this will be forthcoming once all the individual bits and pieces will have been implemented.

eddieantonio:
@aarppe, this PR _only_ adds paradigm label support, hence claiming to fix only #743. I can add lingusitic long labels, but it looks like this:

wâpamêw paradigm with long linguistic labels

Is this desirable?

aarppe:
The labels in the image above would be **linguistic tags** (in the first column in `crk.altlabel.tsv`, e.g. `+1Sg+2SgO`), not **short linguistic labels**, e.g. `1 → 2`.

In principle, we could offer as alternatives any of the columns in `crk.altlabel.tsv`. But in practice not all of them might be implemented, or implementable, e.g. emojis, or the short linguistic tags. What  I had in mind is offering 1) long linguistic labels (column 3); 2) plains English labels (column 4); and 3) nêhiyawêwin labels, to start off with (column 5).

I could imagine that we could use the column names in `crk.altlabel.tsv`, or set a second row in that file as the names of the labeling types to be shown the users, and then a third row coding which labeling formats are made available. Or a separate file with this information. Anyhow, the first column would always correspond to the internal original format (tags from FST analysis, lexical coding in dictionary DB, row/column labels in the layouts/panes), and might not always even need to be shown.

Anyhow, this is a great development!

PR #829, opened 2021-05-31
change: both paradigm views use the same code path

eddieantonio:
# What is in this PR?

The code when using an AJAX call to get a different paradigm size (e.g., the _+ show more_ button) would use a different paradigm generation method than the standalone view. This makes sure they render the same content.

~~# Why is this PR a draft?~~

~~Although the implementation is simple, **the test is difficult to get right**. I wanted a regression test that would check if the AJAX call would create the same HTML that is included in the standalone HTML. However, the Django test client closes the SQLite3 connection **when returning from the first view**. So I can't visit multiple URLs within one test case. I'm not sure how to fix it or how to write a better test case ¯\\\_(ツ)\_/¯~~

The tests pass. It turns out that passing an `HttpRequest()` to another `HttpRequest()` is a bad idea in test cases!

### DIFF
--- a/src/CreeDictionary/CreeDictionary/templates/CreeDictionary/components/paradigm-size-button.html
+++ b/src/CreeDictionary/CreeDictionary/templates/CreeDictionary/components/paradigm-size-button.html
@@ -1,0 +1,29 @@
+{% spaceless %}
+
+  {% comment %}
+    The button
+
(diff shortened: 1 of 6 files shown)


nienna73 (approved):
LGTM

PR #832, opened 2021-06-02
refactor(cypress): cy.login() is available to all specs

eddieantonio:
# What's in this PR?

 - **moves** command definition of `cy.login()` from `admin.spec.js` to Cypress's support file, making it available to **all spec files**.
 - Uses `Cypress.env()` to declare URLs in `cypress.json`, [similar to the speech-db](https://github.com/UAlbertaALTLab/recording-validation-interface/blob/38b15f74be325898bcefce58b6d4834c5e279ae4/cypress.json#L3-L11)
 - Creates a new command: `cy.readCypressUserJSON()` to do exactly that!
 - **enables** a test that was skipped in the last PR — the paradigm labels require a logged-in user (for now)

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -5,5 +5,9 @@
   "projectId": "8r2xra",
   "defaultCommandTimeout": 10000,
   "supportFile": "cypress/support/commands.js",
-  "pluginsFile": false
+  "pluginsFile": false,
+  "env": {
+    "admin_url": "/admin/",
+    "admin_login_url": "/admin/login/"
+  }
 }
--- a/cypress/integration/admin.spec.js
+++ b/cypress/integration/admin.spec.js
@@ -1,35 +1,3 @@
-Cypress.Commands.add('login', () => {
-  cy.visit('/admin/login/')
-  cy.get('[name=csrfmiddlewaretoken]')
-    .should('exist')
-    .should('have.attr', 'value')
-    .as('csrfToken')
-
-  cy.readCypressUserJSON()
-    .then(({username, password}) => {
-      cy.get('@csrfToken').then(function (token) {
-        cy.request({
-          method: 'POST',
-          url: Cypress.env('admin_login_url'),
-          form: true,
-          body: {
-            username,
-            password,
-            next: Cypress.env('admin_url')
-          },
-          headers: {
-            'X-CSRFTOKEN': token,
-          },
-          followRedirect: false
-        }).then(response => {
-          expect(response.status).to.eql(302)
-          expect(response.headers).to.have.property('location')
-          expect(response.headers.location).to.not.contain('login')
-        })
-      })
-    })
-})
-
 context('Admin interface', () => {
   it('should redirect anonymous users to the login page', function() {
     cy.visit('/admin')
--- a/cypress/integration/admin.spec.js
+++ b/cypress/integration/admin.spec.js
@@ -1,6 +1,3 @@
-const ADMIN_LOGIN_URL = '/admin/login/'
-const ADMIN_URL = '/admin/'
-
 Cypress.Commands.add('login', () => {
   cy.visit('/admin/login/')
   cy.get('[name=csrfmiddlewaretoken]')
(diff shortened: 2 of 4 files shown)


cypress/integration/paradigm.spec.js
```diff
@@ -147,12 +147,19 @@ describe('paradigms can be toggled by the show more/less button', () => {
 })
 
 // Can lift skip off this once admin can be tested.
```
andrewdotn, line 149:
```suggestion
```

cypress/integration/paradigm.spec.js
```diff
@@ -147,12 +147,19 @@ describe('paradigms can be toggled by the show more/less button', () => {
...
   let lemma = 'nipâw'
   let englishLabel = 'they'
   let nehiyawewinLabel = 'wiyanaw'
   let linguisticLabel = '3s'
 
+  beforeEach(() => {
+    // As of 2021-06-02: paradigm label switching is only available to
+    // logged-in users. If you are reading this, and paradigm label switching
+    // is enabled for all users, please delete this entire beforeEach() block.
+    cy.login()
```
andrewdotn, line 160: You could have a test that the menu doesn’t appear when not logged in. Then, when the feature is available to all, that test will start failing, and someone will delete it.

cypress/support/commands.js
```diff
@@ -1,28 +1,7 @@
...
-// -- This is a child command --
-// Cypress.Commands.add("drag", { prevSubject: 'element'}, (subject, options) => { ... })
-//
-//
-// -- This is a dual command --
-// Cypress.Commands.add("dismiss", { prevSubject: 'optional'}, (subject, options) => { ... })
-//
-//
-// -- This is will overwrite an existing command --
-// Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
```
andrewdotn, line 25: Thanks for doing this! I am not a fan of checking in big blocks of auto-generated stuff.

cypress/support/commands.js
```diff
@@ -1,28 +1,7 @@
...
+const { join: joinPath } = require('path')
+
+// Why does this path traverse OUTSIDE of the cypress/ directory only to traverse back into it?
+const CYPRESS_USER_JSON = joinPath(__dirname, '..', '..', 'cypress', '.cypress-user.json')
```
andrewdotn, line 4: Because it used to be in `src`.

```suggestion
const CYPRESS_USER_JSON = joinPath(__dirname, '..', '.cypress-user.json')
```

nienna73 (approved):
Nice!

cypress/support/commands.js
```diff
@@ -97,6 +76,59 @@ Cypress.Commands.add('visitLemma', {prevSubject: false}, (lemmaText, queryParams
...
+ *
+ *    cy.readCypressUserJSON()
+ *      .then(({username, password}) => {
+ *        // use username and password
+ *      })
+ *
+ */
+Cypress.Commands.add('readCypressUserJSON', () => {
+  cy.readFile(CYPRESS_USER_JSON)
+})
```
andrewdotn, line 92: Arrow function with `{}` but no `return`: intentional?

cypress/integration/admin.spec.js
```diff
@@ -1,61 +1,22 @@
...
+      expect(pathname).to.contain(Cypress.env('admin_login_url')))
   })
 
   it('should allow login', function() {
     // If this test fails, you are probably using `manage.py runserver` without
     // USE_TEST_DB=True, because the `cypress` user only gets created in the
     // test database.
     cy.visit('/admin')
-    cy.readFile(CYPRESS_USER_JSON)
+    cy.readCypressUserJSON()
```
andrewdotn, line 13: My personal preference, up to you: I avoid uppercase in variable names, so I’d call this `readCypressUserJson`. Makes it more readable when acronyms bump into each other in names, like `getHttpsJson`.

cypress.json
```diff
@@ -5,5 +5,9 @@
   "projectId": "8r2xra",
   "defaultCommandTimeout": 10000,
   "supportFile": "cypress/support/commands.js",
-  "pluginsFile": false
+  "pluginsFile": false,
+  "env": {
+    "admin_url": "/admin/",
+    "admin_login_url": "/admin/login/"
```
andrewdotn, line 11: Maybe someday, not worth doing now, just an idea: do we already have something for JS to read django paths for reversing? Could cypress read that from the django website it’s testing, or from a file that a management command writes to disk?

cypress/support/commands.js
eddieantonio: Wait... I literally do not know why this works. I added the `return` but it seems to not matter...? This can be chained off of _without_ the `return` statement. `cy` must do some weird magic.

PR #833, opened 2021-06-02
fix: bad comment syntax

eddieantonio:
I thought I could use jinja2 comment syntax in a Django template file and the mistake made it to production 😫 (for logged-in users only)


nienna73 (approved):
nobody needs to know...

PR #834, opened 2021-06-02
change: follow up to #823

eddieantonio:
Address some feedback that I didn't notice until after the  #823 was merged >.<

### DIFF
--- a/cypress/integration/paradigm.spec.js
+++ b/cypress/integration/paradigm.spec.js
@@ -154,9 +154,24 @@ describe('Paradigm labels', () => {
 
   beforeEach(() => {
     // As of 2021-06-02: paradigm label switching is only available to
-    // logged-in users. If you are reading this, and paradigm label switching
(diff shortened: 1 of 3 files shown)


cypress/integration/paradigm.spec.js
```diff
@@ -155,11 +154,26 @@ describe('Paradigm labels', () => {
...
-    // logged-in users. If you are reading this, and paradigm label switching
-    // is enabled for all users, please delete this entire beforeEach() block.
+    // logged-in users.
     cy.login()
   })
 
+  it('should only be available for logged-in users', () => {
+    // Undo the beforeEach() that logs in:
+    // (I wanted to make deleting this code as easy as possible)
+    cy.visit(Cypress.env('admin_url')).get('[href*="logout"]').click()
```
eddieantonio, line 164: Yes, this test logs in, then immediately logs out. Why? Makes it SUPER easy to delete both this and the `beforeEach()` above it.

andrewdotn (approved):

PR #836, opened 2021-06-04
refactor: move language-independent paradigm code to morphodict.paradigm package

eddieantonio:
# What is in this PR?

Mostly moving files around:

 - moved `panes.py` and `manager.py` to `morphodict.paradigm` package
 - moved their unit tests to `test_$*.py` within the same package: **NOTE** these unit tests test with Plains Cree data, however the code they test _should_ be language-independent
 - moved `default_paradigm_manager()` (crkeng-specific) to `CreeDictionary.paradigm.generation`

**NOTE**: although [the docs state](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/main/docs/directory-structure.md#proposed-layout) that test data should be in a directory called `testdata` (makes sense), the [pytest plugin](https://github.com/gabrielcnr/pytest-datadir) we're currently using **has no way of customizing the name of the `data` directory**. Any ideas on how to handle this?


nienna73 (approved):

andrewdotn (changes requested):
This requires a bit more coordination, as [my PR from this morning](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/pull/835
) also moves the FST files around.

PR #841, opened 2021-06-07
change: test hfstol is now in LFS

eddieantonio:
Pushes the stray `.hfstol` to LFS storage.

The diff here is really wild!


andrewdotn (approved):
It would be nice to get rid of this file, but this is definitely an improvement.

PR #845, opened 2021-06-08
Add procfile

andrewdotn:


scripts/dev-bootstrap
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -eu
+
+function run() {
+    echo "${@}"
+    "${@}"
```
eddieantonio, line 7: hey dawg i heard you like `make` 😂

scripts/dev-bootstrap
```diff
@@ -0,0 +1,25 @@
...
+}
+
+for LANG_PAIR in \
+    arpeng \
+    crkeng \
+    cwdeng \
+    srseng \
```
eddieantonio, line 14: If you _really_ feel like it, I'm sure we could infer these lists from what files are called `*-manage.py` then use some [expansion magic](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) to get rid of the `-manage.py` stuff.

But I think this script will be so seldom used that it's not worth the confusing bashisms!

eddieantonio (approved):
LGTM! Thanks for making this and documenting it! I think @nienna73 should have a look at this before it's merged, because it incurs a workflow change.

nienna73 (approved):
The docs do a great job of explaining this change!

scripts/dev-bootstrap
andrewdotn: Ha, did not think of that. I’ve added a comment mentioning it could be turned into a small Makefile.

scripts/dev-bootstrap
andrewdotn: Oh, since all the script does is call the manage scripts it wouldn’t even need expansions. That said, as more stuff gets supported for these languages, I expect I will need to replace this script with something more intelligent before we get around to adding another language.

PR #855, opened 2021-06-11
feat: reimplements highlighting for corpus frequency in pane-based paradigms

eddieantonio:
# What's in this PR?

 - **refactor**: move that code to parse `corpus_frequency.txt` to its own file
 - **added**: `{% observed_or_unobserved %}` template tag, because Django's template system makes this kind of thing really ugly 🙃 
 - **added**: `observed_wordform` template context variable to views that render a paradigm

### DIFF
--- a/src/CreeDictionary/CreeDictionary/paradigm/crkeng_corpus_frequency.py
+++ b/src/CreeDictionary/CreeDictionary/paradigm/crkeng_corpus_frequency.py
@@ -1,4 +1,5 @@
 import logging
+from functools import cache
 
 from CreeDictionary.utils import shared_res_dir
 from CreeDictionary.utils.types import ConcatAnalysis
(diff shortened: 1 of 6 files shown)


andrewdotn (commented):
I really like that this is a tag, it keeps the template clean. And having tests for this is wonderful too.

That said, while I know there are some nice things about having the observed wordforms be passed in the template context, I think it *might* be preferable to just have the template tag call the `@cache`d `observed_wordforms()` function directly, because:

  - Fewer moving pieces; it feels a bit off having `observed_wordforms` in `views.py`
  - I worry that having the big set of all observed wordforms directly in the context might *really* slow down the django debug toolbar, which likes to serialize contexts for inspection.

If DJDT is confirmed to perform about the same with this change, I’d be ok with it.

eddieantonio:
> I really like that this is a tag, it keeps the template clean. And having tests for this is wonderful too.
> 
> That said, while I know there are some nice things about having the observed wordforms be passed in the template context, I think it _might_ be preferable to just have the template tag call the `@cache`d `observed_wordforms()` function directly, because:
> 
> * Fewer moving pieces; it feels a bit off having `observed_wordforms` in `views.py`
> * I worry that having the big set of all observed wordforms directly in the context might _really_ slow down the django debug toolbar, which likes to serialize contexts for inspection.
> 
> If DJDT is confirmed to perform about the same with this change, I’d be ok with it.

This is a great suggestion! Yes, putting `observed_wordforms()` in views is a bit strange. I never even thought of what it would do to Django Debug Toobar!

andrewdotn (approved):

aarppe:
One comment, which you might have already taken into account - namely in the case of some cell (i.e. FST-analysis) having multiple possible realizations as word-forms, we should indicated that a word-form is observed only for the form that's been seen in a corpus, and not attached the observedness status on the entire analysis / cell.

PR #858, opened 2021-06-14
feat: compound rows in paradigms

### LINKED ISSUE #507: Split inflections with multiple possible forms over several lines
e.g., instead of this

| Person | Form |
| - | - |
| we |	ninipânân | 
| you and we	| kinipânaw / kinipânânaw |

Let's have this:

| Person | Form |
| - | - |
| we |	ninipânân | 
| you and we	| kinipânaw |
| | kinipânânaw |

---

I know this was there before, but this is unsatisfying... I might need to make a new issue for this. Basically, an entirely new row should be made for any alternatives.

_Originally posted by @eddieantonio in https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/pull/506_

eddieantonio:
# What's in this PR?

Adds compound rows! Do you have inflections that have **multiple valid** realizations? What's the plural of cactus? This PR will place each wordform on its own row instead of concactenating all the results together on the same line!

Before:
kinipânânaw on the same line

After:
kinipânânaw on its own line

- **added**: `CompoundRow` class — contains multiple `ContentRow`
- **added**: `RowLabel` has a `row_span` attribute, which allows it to label multiple rows!
- **added**: `Pane.tr_rows` — iterates rows that should create `<tr>` in the HTML output
- **added**: `SuppressOutputCell` — indicates that a `<td>` should NOT be created in the HTML output
- **added**:  a bevy of unit tests, integration tests, and end-to-end tests

Fixes #507

### DIFF
--- a/src/CreeDictionary/CreeDictionary/paradigm/panes.py
+++ b/src/CreeDictionary/CreeDictionary/paradigm/panes.py
@@ -322,7 +322,7 @@ class ContentRow(Row):
         of ContentRows.
         """
 
-        # fill each COLUMN, then figure out if we need to make a compond row
+        # Fill each **column**, then figure out if we need to make a compound row
         columns = n_empty_lists(self.num_cells)
         for index, cell in enumerate(self.cells):
             columns[index].extend(cell.fill(forms))
@@ -424,6 +424,10 @@ class CompoundRow(Row):
     @property
     def num_cells(self):
         return max(row.num_cells for row in self._rows)
+
+    @property
+    def num_subrows(self) -> int:
+        return len(self._rows)
 
     @property
     def subrows(self) -> Iterable[ContentRow]:
--- a/src/CreeDictionary/CreeDictionary/paradigm/panes.py
+++ b/src/CreeDictionary/CreeDictionary/paradigm/panes.py
@@ -348,7 +348,7 @@ class ContentRow(Row):
         # Create compound rows
         rows = []
         for row_num in range(num_rows_needed):
-            cells = [form_or_empty_cell(col, row_num) for col in columns]
+            cells = [wordform_if_exists_or_no_output(col, row_num) for col in columns]
             rows.append(ContentRow(cells))
         return CompoundRow(rows)
 
@@ -462,6 +462,7 @@ class Cell:
     is_inflection: bool = False
     is_missing: bool = False
     is_empty: bool = False
+    should_suppress_output: bool = False
 
     @staticmethod
     def parse(text: str) -> Cell:
@@ -658,6 +659,16 @@ class EmptyCell(Cell, SingletonMixin):
         return ""
 
 
+class SuppressOutputCell(Cell, SingletonMixin):
+    """
+    Indicates that this should not be rendered at all in the paradigm.
+    Note: even an EmptyCell is "output", in that a <tr> should be created for it in HTML.
+    A SuppressOutputCell should not get even get a <tr>.
+    """
+
+    should_suppress_output = True
+
+
 class BaseLabelCell(Cell):
     is_label = True
     label_for: str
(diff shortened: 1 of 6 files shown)


src/CreeDictionary/CreeDictionary/paradigm/panes.py
```diff
@@ -415,12 +486,12 @@ def contains_wordform(self, wordform: str) -> bool:
...
 
-    def fill_one(self, forms: dict[str, Collection[str]]) -> Cell:
+    def fill(self, forms: Mapping[str, Collection[str]]) -> tuple[Cell, ...]:
         """
-        Returns exactly ONE cell by filling the paradigm.
+        Returns one or more cells by filling the paradigm with the provided forms.
         """
         if not self.is_inflection:
-            return self
+            return (self,)
```
nienna73, line 494: Why does this need to be a tuple?

nienna73 (approved):
What *is* the plural of cactus? 🤔

andrewdotn:
The code looks ok, but honestly I like the version in the first screenshot better, the part with the slash instead of multiple rows.

I don’t see that on the production itwêwina site, so is that something you were experimenting with?

With the two images side by side, seeing `kinipânaw / kinipânânaw` looks like two options for the same thing, where the stacked cells make it look to me like there’s a missing label.

That’s just my opinion though, I’m happy with whatever others decide to do here.

eddieantonio:
> The code looks ok, but honestly I like the version in the first screenshot better, the part with the slash instead of multiple rows.

Yeah, I think styling needs to make it obvious that these forms are related.

> I don’t see that on the production itwêwina site, so is that something you were experimenting with?

Yeah, production itwêwina doesn't even _render_ multiple forms 💀 

> With the two images side by side, seeing `kinipânaw / kinipânânaw` looks like two options for the same thing, where the stacked cells make it look to me like there’s a missing label.
> 
> That’s just my opinion though, I’m happy with whatever others decide to do here.

I think the paradigm presentation needs to be worked on soon (styling, bolding, grouping, etc.). @aarppe, @dwhieb, what do other sources use for when there are multiple realizations of the same inflection?

eddieantonio:
> What _is_ the plural of cactus? 🤔

```
cactus+N+Pl	cactuses
cacus+N+Pl	cacti
```

It can be either _cactuses_ or _cacti_. Both are acceptable and understandable inflections. Now how do you cram that into a single paradigm?

src/CreeDictionary/CreeDictionary/paradigm/panes.py
eddieantonio: The API for `Cell.fill()` changed such that filling _any_ cell can return one or more results, hence a tuple. Maybe this isn't the best API change though :/

aarppe:
Looking at various on-line resources on paradigms, some by more established entities and others by more random matches when searching for 'paradigm(s)', Wiktionary appears to put alternative forms on their own lines (separated by a semicolon, that I'd ignore), e.g. for the Finnish _[hevonen](https://fi.wiktionary.org/wiki/hevonen)_ 'horse'. Whether the alternative words end up on their own lines due to cell size or not is not clear.

image

Another approach is taken by this older resource for Latin, separating alternative forms with commas ([link](http://dcc.dickinson.edu/grammar/latin/1st-declension-case-forms)):

image

This same approach is used in the (Finnish) Language Bureau's on-line dictionary for [Finnish](image):

image

Another alternative approach for Latin uses slashes to separate alternative forms ([link](https://classics.osu.edu/Undergraduate-Studies/Latin-Program/Grammar/Paradigms/personal-endings)):

image

I'd think that we can use any approach that is clear and makes sense to users. A slash is quite broadly understood to separate alternatives.

eddieantonio:
I merged this, but @andrewdotn, I think you're right. Slashes would probably work better, given the examples @aarppe posted. I'll punt that change until #862. In the meantime, @aarppe, this PR fixes the issue that highlighting should apply for each observed _wordform_ rather than _analysis_.

Issue #585, opened 2020-12-17
Implement presentation of reduplication and initial change


aarppe:
While we now show separately information about preverbs, as retrieved from the linguistic analysis, reduplication and initial change are now excluded. We would want to show information about at least reduplication similar to preverbs. The treatment of initial change requires separate consideration.

Reduplication presents a tricky task, as the light/weak and heavy/strong reduplication are realized in multiple ways, based on the following phonology, the meaning of light/weak as indicating _ongoing, continuing_ and of heavy/strong as expressing _intermittent, repeatedly; again and again; here and there_ stays the same across the different realizations. Consider e.g. the following analysis:

```
ê-ma-miyo-mâh-mâmawi-na-nikamot
PV/e+RdplW+PV/miyo+RdplS+PV/mamawi+RdplW+nikamow+V+AI+Cnj+3Sg
```

This should result in the following presentation:

Preverb: ê- <img width="30" alt="image" src="https://user-images.githubusercontent.com/15003407/102447086-7e603c80-3fec-11eb-9cf8-9c928d8f981e.png">
Reduplication: ma- <img width="30" alt="image" src="https://user-images.githubusercontent.com/15003407/102447086-7e603c80-3fec-11eb-9cf8-9c928d8f981e.png">
Preverb: miyo- <img width="30" alt="image" src="https://user-images.githubusercontent.com/15003407/102447086-7e603c80-3fec-11eb-9cf8-9c928d8f981e.png">
Reduplication: mâh- <img width="30" alt="image" src="https://user-images.githubusercontent.com/15003407/102447086-7e603c80-3fec-11eb-9cf8-9c928d8f981e.png">
Preverb: mâmawi- <img width="30" alt="image" src="https://user-images.githubusercontent.com/15003407/102447086-7e603c80-3fec-11eb-9cf8-9c928d8f981e.png">
Reduplication: na- <img width="30" alt="image" src="https://user-images.githubusercontent.com/15003407/102447086-7e603c80-3fec-11eb-9cf8-9c928d8f981e.png">

Then, the <img width="30" alt="image" src="https://user-images.githubusercontent.com/15003407/102447086-7e603c80-3fec-11eb-9cf8-9c928d8f981e.png"> button for any case of light/weak reduplication should show _ongoing, continuing_  and for any case of heavy/strong reduplication: _intermittent, repeatedly; again and again; here and there_.

[...]

PR #859, opened 2021-06-14
Feat/reduplication

nienna73:
## What's in this PR:
* Show reduplication with preverbs
* Move preverbs into a "lexical_information" attribute
* Show "strong reduplication" and "weak reduplication" as applicable
* Test for showing reduplication

Looks like this:
Screen Shot 2021-06-14 at 4 10 29 PM

**Notes:**
This PR addresses part of Issue #585 
This likely **is not** language independent


src/CreeDictionary/CreeDictionary/templates/CreeDictionary/components/search-result.html
```diff
@@ -65,7 +65,7 @@
       {% endfor %}
     </ol>
 
-    {% include "CreeDictionary/components/preverb-breakdown.html" %}
```
andrewdotn, line 68: If `lexical-info.html` is an updated and renamed version of `preverb-breakdown.html`, can you `git rm` the original file so that the diff for `lexical-info.html` shows changes instead of appearing as a brand-new file?

nienna73: It started as an entirely new file for reduplication only, then I renamed it when I decided to do preverbs and reduplication together. I don't think `git rm` will realize that?

aarppe: If we want to make it future proof, if only for crk, we should include prenouns as well. There in Arok's CW; I just haven't incorporated them in the FST yet.

aarppe:
We should add a hyphen to the end of the reduplicative elements, i.e. _ma-_, _mâh-_, etc., to indicate that they are non-independent morphemes.

Also, does the current implementation provide _ay-_ and _âh-_ for the reduplication of vowel-initial stems/pre-whatevers?

src/CreeDictionary/CreeDictionary/templates/CreeDictionary/components/search-result.html
andrewdotn: @nienna73 If the files have similar enough content, git can still infer that the removal + addition is a rename with changes.

src/CreeDictionary/API/search/runner.py
```diff
@@ -10,6 +10,8 @@
 from CreeDictionary.API.search.util import first_non_none_value
 from CreeDictionary.utils.types import cast_away_optional
 
+from morphodict.analysis import relaxed_analyzer
+
```
andrewdotn, line 14: Unused import?

```suggestion
```

src/CreeDictionary/API/search/presentation.py
```diff
@@ -19,6 +20,7 @@ class SerializedPresentationResult(TypedDict):
     wordform_text: str
     is_lemma: bool
     definitions: Iterable[SerializedDefinition]
+    lexical_information: List[Dict[str, Any]]
```
eddieantonio, line 23: Please avoid using the `Any` type. It defeats the purpose of using static type-checking :/

I believe the type you want represents this data, correct?

```python
  {
                    "entry": serialize_wordform(preverb_result),
                    "type": "Preverb",
                    "index": i,
                    "original_tag": tag,
                }
```

You could create a `TypedDict` for this or a dataclass:

```python
@dataclass
class SerializedMorpheme:
    entry: SerializedWordform
    type: Literal["Preverb", "Reduplication", "InitialChange"]
    index: int
    original_tag: FSTTag
```

I feel like this still could be iterated upon :/

src/CreeDictionary/tests/API_tests/model_test.py
```diff
@@ -210,7 +210,21 @@ def test_search_words_with_preverbs():
...
+    """
+    reduplication should be extracted and present in SearchResult instances
+    """
+    results = search(query="nanipâw").presentation_results()
+    assert len(results) == 1
+    search_result = results.pop()
+
+    assert len(search_result.lexical_info) == 1
+    print(search_result.lexical_info)
+    assert search_result.lexical_info[0]["entry"].text == "na"
```
eddieantonio, line 227: The reduplication text should _probably_ include the hyphen, as per @aarppe's suggestion

```suggestion
    assert search_result.lexical_info[0]["entry"].text == "na-"
```

eddieantonio (commented):
I'm curious: why not handle initial change in this PR?

src/CreeDictionary/API/search/presentation.py
aarppe: We should also be prepared for Prenouns, which are quite similar to Preverbs but attach instead to nouns (and are often the same as preverbs, but with an attributive rather than adverbial meaning).

src/CreeDictionary/tests/API_tests/model_test.py
aarppe: The hyphen generally indicates the non-independent status of a linguistic element - which is the consensus case with morphemes such as preverbs.

src/CreeDictionary/API/search/presentation.py
nienna73: Noted. I'll play around with the types and dataclass idea and see what I can do.

nienna73:
> Also, does the current implementation provide ay- and âh- for the reduplication of vowel-initial stems/pre-whatevers?

No, it does not, I can add support for this feature.

src/CreeDictionary/CreeDictionary/templates/CreeDictionary/components/lexical-info.html
```diff
@@ -0,0 +1,60 @@
...
+  {% load static %}
+
+  {# show preverb and reduplication breakdown #}
+  <ol class="preverb-breakdown">
+    {% for item in result.lexical_information %}
+      <li>
+          <span>{{ item.type }}
+            {% if item.type == "Initial Change" %}
+            {% else %}:
+            {% endif %}
```
eddieantonio, line 31:
```suggestion
            {% if item.type != "Initial Change" %}: {% endif %}
```

eddieantonio (commented):
I'm confused about a few parts, but it's getting close to ready!

frontend/css/styles.css
```diff
@@ -710,6 +710,10 @@ a.multiple-recordings__action-button {
   margin: 0;
 }
 
+.preverb-breakdown__preverb-definition-text {
+  white-space: pre-line;
```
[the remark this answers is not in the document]
nienna73: Yes! The initial change tooltip shows all possible initial changes, as outlined in #585 . In order for them all the show up on a newline, we have to include this `pre-line` style.

src/CreeDictionary/API/search/presentation.py
```diff
@@ -43,14 +79,24 @@ def __init__(self, result: types.Result, *, search_run: core.SearchRun):
...
 
-        analysis = result.wordform.analysis or [[], None, []]
         (
             self.linguistic_breakdown_head,
             _,
             self.linguistic_breakdown_tail,
-        ) = analysis
+        ) = result.wordform.analysis or [[], None, []]
+
+        self.lexical_info = get_lexical_information(result.wordform.analysis)
```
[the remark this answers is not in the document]
nienna73: Thanks for pointing this out! I'm changing all of them to `lexical_info` since they all reference the same thing and that's fewer characters to type.

src/CreeDictionary/API/search/presentation.py
```diff
@@ -181,21 +237,42 @@ def to_fst_output_style(value):
...
+        if tag in ["RdplW+", "RdplS+"]:
+            reduplication_string = generate_reduplication_string(
+                tag, first_letters[i + 1]
+            )
+
+        elif tag == "IC+":
+            change_types = get_initial_change_types()
+            _type = "Initial Change"
+            entry = _InitialChangeResult(
+                text=" ",
```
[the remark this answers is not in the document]
nienna73: `text` is the preverb tag or reduplication string in the other instances, so something like `pê-` or `na-`. For IC, there is no string to put in this place, but the template will still look for something, so I chose to use a single empty string

src/CreeDictionary/tests/API_tests/model_test.py
```diff
@@ -213,7 +213,34 @@ def test_search_words_with_preverbs():
...
+def test_search_words_with_reduplication():
+    """
+    reduplication should be extracted and present in SearchResult instances
+    """
+    results = search(query="nanipâw").presentation_results()
+    assert len(results) == 1
+    search_result = results.pop()
+
+    assert len(search_result.lexical_info) == 1
+    assert search_result.lexical_info[0]["entry"]["text"] == "na-"
```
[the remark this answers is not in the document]
nienna73: Yes, there probably should be.

src/morphodict/lexicon/models.py
```diff
@@ -148,6 +148,7 @@ def get_absolute_url(self, ambiguity: Literal["allow", "avoid"] = "avoid") -> st
          it's the least strict url that guarantees unique match in the database
         """
         assert self.is_lemma, "There is no page for non-lemmas"
+        # FIXME: will return '/word/None' if no slug
```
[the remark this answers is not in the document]
nienna73: I think a combination of introducing the `slug` into the DB and the new preverb presentation caused this issue. It is definitely out of scope for this PR, but should be reproducible once these changes are live.

eddieantonio (approved):
LGTM!

nienna73:
Resolves #585

Issue #862, opened 2021-06-16
Review paradigm styling


eddieantonio:
I think it's time to reconsider how the paradigms are presented. This includes the presentation of:

 - [ ] pane headers
 - [ ] row labels
 - [ ] column labels
 - [ ] observed wordforms
 - [ ] unobserved wordforms
 - [ ] inflections with mulitple possible realizations (see #858)

I gather most of this will be CSS tweaks, however, I think there might be some HTML tweaks as well. If we decide to use slashes to separate multiple possible realizations, I think a whole lot of Python code would need to be deleted too 😈

aarppe:
Concerning how to distinguish linguistic content (the cells that the FST generates) from the row and column labels, #568 has a relevant observation/suggestion, based on old NDS, as illustrated in the image below:

<img width="384" alt="image" src="https://user-images.githubusercontent.com/15003407/122323849-fdbfc180-cee4-11eb-9285-7fc76607eabe.png">

I suppose the blue color was motivated by the labels in question having pop-ups with alternative relabelings.

aarppe:
Beyond the above, there are some elements in #505, namely:

- [ ] showing optionally morpheme boundaries (option b.)
- [ ] showing the English phrase translations as pop-ups to the Cree wordforms in the individual cells.

We might consider the rest, i.e. presenting in a pop-up the full range of additional information for each cell with Cree word-forms, separate of this issue.

eddieantonio:
See: https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/pull/864#issuecomment-867274210

Issue #405, opened 2020-05-14
Add a page/legend or help text explaining abbreviations


eddieantonio:
Especially Arok's abbreviations, e.g., `s.t.`/`s.o`/`s/he`/`it`, etc.

aarppe:
And/or a pop-up? I.e. when you hover over `s/he`, you got `she or he` (in the English definitions)?

aarppe:
@eddieantonio This is something I can do. Should we have this as a page available under itwêwina, or an external page (under ALTLab)? Eventually we would also have links to the grammar pages that Arok is in the process of creating, but that would be more comprehensive explanations of Cree grammar, while we'd just be explaining tersely the various code we use - so we might have link elsewhere for "More information".

eddieantonio:
Please create a separate page at <https://www.altlab.dev/> (`altlab-gw:/data/www.altlab.dev/`) and add a link to the page in the footer!

eddieantonio:
We need to explain the emoji too — my suggestion: use the exact words that Dorothy uses when teaching the difference between VTA and VTI (e.g., [pêtâ](https://sapir.artsrn.ualberta.ca/cree-dictionary/word/p%C3%AAt%C3%A2w/) masinahikan/[pêsiw](https://sapir.artsrn.ualberta.ca/cree-dictionary/word/p%C3%AAsiw%C3%AAw/) maskwa).

PR #864, opened 2021-06-17
Small changes to paradigm styles

nienna73:
## What's in this PR:
* changed which elements are bold vs. italic

It looks like this now:
Screen Shot 2021-06-17 at 10 03 26 AM

### DIFF
--- a/frontend/css/styles.css
+++ b/frontend/css/styles.css
@@ -794,8 +794,7 @@ a.multiple-recordings__action-button {
 .paradigm-title,
 .paradigm-header {
   text-align: center;
-  font-style: italic;
-  font-weight: var(--body-font-weight);
+  font-weight: var(--strong-font-weight);
 }
 
 .paradigm-label {
@@ -804,7 +803,7 @@ a.multiple-recordings__action-button {
 }
 
 .paradigm-label--row {
-  font-style: italic;
+  font-weight: var(--strong-font-weight);
 }
 
 .paradigm-label--col {
@@ -820,14 +819,16 @@ a.multiple-recordings__action-button {
 }
 
 .paradigm-cell--observed{
-  font-weight: var(--strong-font-weight);
+  font-weight: var(--body-font-weight);
+}
+
+.paradigm-cell--unobserved{
+  font-style: italic;
 }
 
 .paradigm-cell--no-analysis{
   background-color: lightgray;
 }
-
-
 
 .paradigm-label--row, .paradigm-cell {
   vertical-align: top; /* When a label applies to multiple wordforms */


eddieantonio:
I will abstain from reviewing this, since I was there, but @aarppe, @andrewdotn, and @dwhieb, I'd appreciate your thoughts on these styling changes!

eddieantonio:
Addresses #862.

aarppe:
For separating the linguistic content from the meta-linguistic labels, bolding the latter does work well. But for distinguishing observed forms from merely generated ones, with italicizing the former, I'm not convinced that that formatting will achieve the goal. People might consider italics as lesser that regular, and thin that the forms that are italicized are not observed, and not the other way around.

Honestly, the old scheme of boldfacing observed forms in contrast to unobserved forms was self-evident. Of course, that conflicted with column/row labels, if we'd want to bold-face those to distinguish them clearly.

I could imagine some options:

1. bold-face labels, bold-face and italicize observed forms, and keep unobserved forms with regular font.
2. bold-face labels, keep observed forms as regular, and present the unobserved forms with lesser grey (as long as that is legible).
3. Use a different color for the labels, though still boldfaced (I've suggested blue, as in the old itwêwina), present observed forms in bold-face and unobserved forms with regular font. See: https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/862#issuecomment-862876972

eddieantonio:
> For separating the linguistic content from the meta-linguistic labels, bolding the latter does work well. But for distinguishing observed forms from merely generated ones, with italicizing the former, I'm not convinced that that formatting will achieve the goal. People might consider italics as lesser that regular, and thin that the forms that are italicized are not observed, and not the other way around.

This is actually what the styling indicates. _mîcisow_ is an unobserved form of “mîcisow”!

> Honestly, the old scheme of boldfacing observed forms in contrast to unobserved forms was self-evident.

So I've had the discussion with many people: “what does the bolding mean?” and I've had to explicitly explain it to them. It is not self-evident. Plus, #405 still needs to be done that explicitly explains certain things, like the corpus frequencies.

> Of course, that conflicted with column/row labels, if we'd want to bold-face those to distinguish them clearly.
> 
> I could imagine some options:
> 
> 1. bold-face labels, bold-face and italicize observed forms, and keep unobserved forms with regular font.
> 2. bold-face labels, keep observed forms as regular, and present the unobserved forms with lesser grey (as long as that is legible).
> 3. Use a different color for the labels, though still boldfaced (I've suggested blue, as in the old itwêwina), present observed forms in bold-face and unobserved forms with regular font. See: [#862 (comment)](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/issues/862#issuecomment-862876972) as well as the image below:
> 
> image

We'll have to iterate on the design a bit more. I will continue to resist colouring text blue if it is not a link. Of course, with #405 finished, we could make it a link!

nienna73:
I largely agree with @eddieantonio  's comments. I think blue should be reserved for links, unless we have another very clear colour we're using for links throughout morphodict/those items actually do link to something. Also, having black and grey text side by side does weird things to my eyes and my perception of colours--I often end up thinking they're the same colour and it just looks different based on whether I'm looking directly at the text.

I think we have a lot of information to convey implicitly and getting this to a state where everyone automatically knows what's being presented is going to take multiple iterations

aarppe:
I looked at the revision again. It does a good job in distinguishing the word-forms from the labels. But I have a really hard time distinguishing attested from unattested forms – the italicization really needs focus and is hard to discern. Weirdly, in some ways, I would that the current rendering is almost better, in that the row labels are consistently italicized, cf.

image

If we must still with a single color, i.e. black, I'm not really sure what the best options are. Perhaps they would be the following:

1. boldface row and column labels, as well as attested wordforms, and leave non-attested wordforms as plain regular.
2. boldface row and column labels, mark non-attested word-forms with some 50% grey, and leave the attested wordforms as plain regular.
3. Other combinations, even including borderlines.

Here are some alternative visualizations:

image

image

image

image

image

image

PR #865, opened 2021-06-17
remove: old paradigms (code+layouts)

eddieantonio:
# What's in this PR?

Mostly getting rid of the old NDS-style paradigm filling code, but here are things that were changed or added:

 - **added**: `ParadigmManager.all_analyses(pardigm_name, lemma)` — returns a set of all `analysis+N+Sg` strings for a given lemma/paradigm.
 - **changed**: `cree_inflection_generator.py` uses `ParadigmManager.all_analyses()` instead of old paradigm filler
 - **changed**: `paradigm-with-panes.html` was renamed to `paradigm.html` -- no need to make that distinction anymore
 - **changed**: mîcisow is not a VTI... it's a VAI... 😅 
 - **added**: integration tests for `crkeng` to find some basic inflections of all our regular, inflectional wordclasses

Otherwise, it's just ✂️ ✂️ ✂️  deleting the legacy paradigm filler code!


eddieantonio:
Converting to draft because this broke **auto-translations** :/

andrewdotn:
The test word for auto-translations is acâhkosa. It no longer appears under any of the paradigm tables for <https://itwewina.altlab.app/word/atâhk/>, so it doesn’t get created as a Wordform during `ensuretestdb`. And the current phrase translation only looks in the DB.

andrewdotn:
Specifically, “little star over there” as `ahcahk+N+A+Der/Dim+N+A+Obv` `atâhk+N+A+Der/Dim+N+A+Obv` is no longer in the paradigms.

I’m a little out of the loop on this:
  - Is size ‘full’ meant to completely replace the old ‘linguistic’ paradigm tables?
  - Is it intentional that the diminutive locative is now gone?

eddieantonio:
> Specifically, “little star over there” as `ahcahk+N+A+Der/Dim+N+A+Obv` `atâhk+N+A+Der/Dim+N+A+Obv` is no longer in the paradigms.

This is very good to know!

> I’m a little out of the loop on this:
> 
> * Is size ‘full’ meant to completely replace the old ‘linguistic’ paradigm tables?

It's supposed to have the same content as the linguistic paradigm layouts; however,

> * Is it intentional that the diminutive obviative is now gone?

I believe @aarppe intentionally removed `+Der/Dim` from the layouts since, as the tag `+Der/Dim` suggests, this is _derivational_ morphology, rather than _inflectional_ morphology. Anytime you have derrivation, you essentially have a **new lexeme**, and (usually) a new morphological stem. We ordinarily make an entirely new entry for this, which is the case for _acâhkos_: https://itwewina.altlab.app/word/ac%C3%A2hkos/

@aarppe: I believe this is an issue in the phrase translator. Do we want it working with derivational morphology, _as well as_ inflectional morphology? I think diminuitives could be thought out a little more.

eddieantonio:
@andrewdotn: I changed the auto-translate test example to _niminosâk_ “my cats”. Here are my reasons:

 - niminôsak currently has one analysis: `minôs+N+A+Px1Sg+Pl`; `acâhkosa` has two different analyses with _two different lemmas_: `acâhkos+N+A+Obv` and `atâhk+N+A+Der/Dim+N+A+Obv`
 - `+Der/Dim` is something I believe needs more consideration, as mentioned in a comment above
 - `+Obv` has no straightforward translation into English, as there is no one-to-one correspondance in English; as such, I believe the phrase translation FST could change in this respect, and therefore, any tests that rely on its output
 - In terms of tag mappings, it's the same complexity, `Pl+Px1Sg +cat` vs. `Dim+Obv+ star`
 - niminôsak is something I'd say everyday :)

aarppe:
I agree with all of Eddie's analyses and comments, concerning the status of the diminutives as derivation -- what he is noting is something that Arok has expressed as well. A few additional notes

* the English phrase analysis and generation FSTs are more powerful than what we cover in the full paradigm layouts. In part this is due to me trying to include some non-Cree phenomena, e.g. progressive and repetitive constructions that are needed for Tsuut'ina.
* how to translate obviatives is a good question. In the layouts I used **_further_**, whereas in the phrase translation I chose _**over there**_ instead. Those labels should rather be considered just labels that somehow capture something of what is conveyed with the Cree obviative (i.e. topically less focused), but crucially those labels should not be considered the most accurate translations (i.e. not really a medial or distal demonstrative, even though that captures something of the nature of the lesser, i.e. obviated, focus).

andrewdotn (approved):

PR #866, opened 2021-06-17
ci: run actions on Ubuntu 20.04

eddieantonio:
GitHub is deprecating Ubuntu 16.04 from its Actions runners. I believe it was hard-coded to use 16.04 because Sapir was running 16.04.

Note: `${ACTIONS_OS_VERSION}` variable substitution was not being used, and I think it never worked, so I just deleted that variable!

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -10,7 +10,7 @@ env:
 jobs:
   # Skips deploy if [skip deploy] is present in the commit message
   should-deploy:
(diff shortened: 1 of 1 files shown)


andrewdotn:
Would you be ok setting it to the `ubuntu-latest` literal, resulting in GH auto-upgrading it in the future?

eddieantonio:
> Would you be ok setting it to the `ubuntu-latest` literal, resulting in GH auto-upgrading it in the future?

Done!

andrewdotn (approved):

PR #868, opened 2021-06-18
change: ParadigmManager API no longer makes a distinction between static and dynamic layouts

eddieantonio:
# What's in this PR?

We decided that the `ParadigmManager` doesn't really need to care whether a layout is "static" or "dynamic" anymore.

**NOTE**: I did NOT change the directory structure, so the layouts are still being loaded from `layouts/{dynamic,static}`.

 - **removed**: `ParadigmManager.static_paradigm_for()`
 - **removed**: `ParadigmManager.dynamic_paradigm_for()`
 - **changed**: `ParadigmManager.paradigm_for()` just... does the right thing!
 - **added**: `ParadigmDoesNotExistError` for when you to try to generate a paradigm that does not exist!

+ other misc. refactorings!

### DIFF
--- a/src/CreeDictionary/CreeDictionary/paradigm/manager.py
+++ b/src/CreeDictionary/CreeDictionary/paradigm/manager.py
@@ -13,6 +13,12 @@ ONLY_SIZE = "<only-size>"
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


andrewdotn (approved):
Nice, I like how it cleans things up.

PR #872, opened 2021-06-24
Clean up some old paradigm code

### LINKED ISSUE #806: Make frontend dumber: Communicate certain information completely in HTTP requests to DRY the code
Some knowledge are duplicated in the backend and frontend such as:

 - paradigm "sizes"
 - list of available orthographies
 - current orthography
 - how to set the page title
 - what interactive elements are required on the current page; e.g.,
     * do the paradigm buttons need to be initialized?
     * does audio need to be loaded for each search result?

This can all be provided in the **Django HTML template** and parsed by the JavaScript on page load and on

eddieantonio:
Before I get ready to work on #806, I wanted to clean up some code leftover from the old implementation.


andrewdotn (approved):
Looks ok

BTW I wouldn’t spend much time doing cleanup on this at the moment, I will have a PR in a day or two where I already have [some things like default size working](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/compare/main...andrewdotn:lexicon-wordform?expand=1#diff-198e89c996ab73c49db73d64c0202851caf9772621e7bf2a1dd71290e97d0c30), including the JS bit.

src/CreeDictionary/CreeDictionary/paradigm/manager.py
```diff
@@ -61,6 +63,29 @@ def sizes_of(self, paradigm_name: str) -> Collection[str]:
...
+        """
+        return len(self.sizes_of(paradigm_name)) > 1
+
+    def default_size_of(self, paradigm_name: str) -> str:
+        """
+        Returns the default size of the paradigm. For paradigms with multiple sizes,
+        this is the first (See ParadigmManagerWithExplicitSizes if you want more
+        control over this).
+
+        :raises ParadigmDoesNotExistError: when the paradigm name cannot be found.
```
nienna73, line 80: How do these raise errors without explicitly raising errors? I would like to learn

nienna73 (approved):
Looks good, just one question about raising errors.

src/CreeDictionary/CreeDictionary/paradigm/manager.py
eddieantonio: `sizes_of()` may raise this error, so it's possible that this _also_ raises the exact same error.

PR #874, opened 2021-06-24
Morphodict lexicon wordform

andrewdotn:
This replaces the existing API_wordform database table with a new lexicon_wordform database table that is simpler and much less Cree-specific. It uses the test database of https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/pull/873

All the tests pass, but I have not yet tested this branch’s performance on the full dictionary, so it may need some tweaking before it is ready for deployment.

To avoid changing everything all at once, we are using exported relabellings from the currently-deployed version of itwêwina, rather than rewriting all of that code too.

I will probably update this PR to remove all of the DatabaseManager and DatabaseManager_tests code, as that is now all unused.


src/morphodict/lexicon/test_util.py
```diff
@@ -0,0 +1,21 @@
...
+    [
+        ("hello", "hello"),
+        ("hellø", "hello"),
+        ("hełło", "hello"),
+        ("hêllò", "hello"),
+        ("hěllo", "hello"),
+        (
+            "h̷̨̨̢̝̺̥͕̦̘̺̥̖̺̐ͅe̶̢̛͙̱͔̗͓͓̊̇̓̍̈́͐͋̍̐͑͌̌̿͘͜͠͝͝ļ̴̨̧͔̭̲̲͎̘̱̬̼͉̜̱̤͎̹̞̝̥̲̙̙̮̪̖̽̓̌͊̊̍̆́̎̑͌̇̓̈́́͊͌̾̆̿͊̎͌̈̕̕͜͠ͅl̸̨̧͓̭̻̝͖͕̪͖̩̯̞͙͖͙̦͈̘̥̱̩̫̬̎̈́͗̓̑̎̾͋̍̏͒̽̃̕͠ǒ̵͔̯̜̣̘̹̑͜͝",
+            "hello",
+        ),
```
eddieantonio, line 17: I would honestly want to see genuine examples from the languages we want to use _in addition_ to Unicode gore. Possible Unicode gore:

 - Greek casing: e.g., what should `to_source_language_keyword("σκύλος")` return?  Note that both `σ` and `ς` are lowercase Greek sigma (one is the final form). `.lower()` is context-sensitive, and not safe to apply on a per-character basis — you need to apply it on a full string.

There's more things you can bug me about, like Turkish, katakana fullwidth forms, etc. NFKD might be more appropriate for an "internal" representation.

src/morphodict/lexicon/util.py
```diff
@@ -0,0 +1,19 @@
...
+
+    Currently removes accents, and leading and trailing hyphens.
+
+    There will be collisions but we could use edit distance to rank them.
+    """
+    return (
+        "".join(c for c in normalize("NFD", s) if unicodedata.combining(c) == 0)
+        .translate(EXTRA_REPLACEMENTS)
+        .strip("-")
+        .lower()
```
eddieantonio, line 18: `.lower()` and `.upper()` are not safe to apply on a per-character basis. See: https://unicode.org/Public/UNIDATA/SpecialCasing.txt

```suggestion
        "".join(c for c in normalize("NFD", s).lower() if unicodedata.combining(c) == 0)
        .translate(EXTRA_REPLACEMENTS)
        .strip("-")
```

eddieantonio (commented):
Some preliminary comments. I'm assuming you'll update this PR after #873 is merged?

src/conftest.py
```diff
@@ -6,4 +6,5 @@
     "arpeng/site/settings.py",
     "cwdeng/site/settings.py",
     "srseng/site/settings.py",
+    "CreeDictionary/tests/DatabaseManager_tests",
```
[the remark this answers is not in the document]
andrewdotn: Right now I’m planning to put that into a followup cleanup PR, due to the the sheer bulk of the current one.

src/morphodict/lexicon/test_util.py
andrewdotn: I did not realize that `lower()` isn’t safe on a character-by-character basis! The above code does actually run `lower()` on the whole string at the end, but I have moved it earlier in the function for clarity.

src/morphodict/lexicon/management/commands/importjsondict.py
```diff
@@ -0,0 +1,174 @@
...
+            d = Definition.objects.create(
+                wordform=wordform,
+                text=sense["definition"],
+            )
+            for source in sense["sources"]:
+                try:
+                    d.citations.add(source)
+                except:
+                    breakpoint()
+                    raise
```
eddieantonio, line 157: Leftover debugging code? What happened here?

```suggestion
                d.citations.add(source)
```

src/morphodict/lexicon/management/commands/importjsondict.py
```diff
@@ -0,0 +1,174 @@
...
+        if atomic:
+            with transaction.atomic():
+                self.run_import(json_file=json_file, purge=purge)
+        else:
+            self.run_import(json_file=json_file, purge=purge)
+
+    def run_import(self, json_file, purge):
+        for abbrv in ["CW", "MD", "AE", "ALD", "OS"]:
+            if not DictionarySource.objects.filter(abbrv=abbrv):
+                DictionarySource.objects.create(abbrv=abbrv)
```
eddieantonio, line 61: So every site will have dictionary source entries for all other sites? This seems... odd.

I think DictionarySources will need to be addressed in a future PR — there's definitely some work to be done here, but it's definitely okay to punt that work until later!

src/morphodict/lexicon/migrations/0003_remove_unused_dictionarysource_fields.py
```diff
@@ -0,0 +1,35 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
eddieantonio, line 4: Since this is all coming in one PR, could you squash these linear migrations? It seems redundant to merge in code that adds columns to the database, and then immediately removes them.

src/morphodict/lexicon/models.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+class DiacriticPreservingJsonEncoder(DjangoJSONEncoder):
+    """Stores Unicode strings, e.g., "pê", in the database
+
+    Instead of ASCII-fied "p\u00ea".
+    """
+
+    def __init__(self, *args, **kwargs):
+        kwargs = {**kwargs, "ensure_ascii": False}
```
eddieantonio, line 47: 👍🏼

src/morphodict/lexicon/models.py
```diff
@@ -0,0 +1,261 @@
...
+
+    paradigm = models.CharField(
+        max_length=MAX_WORDFORM_LENGTH,
+        null=True,
+        blank=False,
+        default=None,
+        help_text="If provided, this is the name of a static paradigm that this "
+        "wordform belongs to. This name should match the filename in "
+        "res/layouts/static/ WITHOUT the file extension.",
+    )
```
eddieantonio, line 67:
```suggestion
        help_text="If provided, this is the name of a paradigm that this "
        "wordform belongs to. This name must match the filename or directory in "
        "res/layouts/ (without the file extension).",
    )
```

src/morphodict/lexicon/models.py
```diff
@@ -0,0 +1,261 @@
...
+
+    is_lemma = models.BooleanField(
+        default=False,
+        help_text="The wordform is chosen as lemma. This field defaults to true if according to fst the wordform is not"
+        " analyzable or it's ambiguous",
+    )
```
eddieantonio, line 73: I'm extremely confused. Based on _usage_, what does `is_lemma_ actually mean? Is this help text up to date?

src/CreeDictionary/CreeDictionary/views.py
```diff
@@ -54,20 +55,38 @@ def entry_details(request, lemma_text: str):
         - paradigm-size (default is BASIC) to specify the size of the paradigm
 
     :param request: accepts query params `pos` `inflectional_category` `analysis` `id` to further specify query_string
-    :param lemma_text: the exact form of the lemma (no spell relaxation)
+    :param slug: the stable unique ID of the lemma
     """
 
-    lemma = Wordform.objects.filter(text=lemma_text, is_lemma=True)
-    if additional_filters := disambiguating_filter_from_query_params(request.GET):
```
eddieantonio, line 61: Ah, this view is much clearner in general!

Please delete the documentation in the docstring related to paradigm sizes 👆🏼 👆🏼 👆🏼

src/CreeDictionary/DatabaseManager/cree_inflection_generator.py
```diff
@@ -32,10 +29,7 @@ def expand_inflections(
         lemma, word_class = lemma_and_wordclass_from_analysis(analysis)
 
         if word_class.has_inflections():
-            paradigm_name = convert_crkeng_word_class_to_paradigm_name(word_class)
-            generated_analyses = list(
-                paradigm_manager.all_analyses(paradigm_name, lemma)
-            )
+            raise Exception("convert_crkeng_word_class_to_paradigm_name is gone")
```
eddieantonio, line 32: Why even have this line? Is the `WordClass` enum even used or instantiated anymore?

src/CreeDictionary/phrase_translate/management/commands/translatewordforms.py
```diff
@@ -33,31 +32,34 @@ class InsertBuffer:
...
+    def __init__(self, manager, count=500, assign_id=False):
+        """
+        If assign_id is True, this class will assign IDs to objects added to it,
+        so that they can immediately be used in foreign key references.
+
+        A future Django release should remove the need for this method. Django
+        bulk_create can currently return database-generated IDs on Oracle and
+        Postgres; INSERT … RETURNING support was only added to the SQLite
+        database itself in 3.35.0 released 2021-03-12, so Django does not
+        support it yet.
```
eddieantonio, line 44: 👍🏼

src/CreeDictionary/phrase_translate/translate.py
```diff
@@ -92,26 +93,10 @@ def foma_lookup(fst, thing_to_lookup):
...
-        preverb_tags, lemma_, tags = partition_analysis(
-            cree_wordform_tag_list_or_analysis
-        )
-        preverb_tags = [x + "+" for x in preverb_tags]
-        tags = ["+" + x for x in tags]
-        cree_wordform_tag_list = preverb_tags + tags
+def inflect_english_phrase(analysis, lemma_definition):
+    if isinstance(analysis, tuple):
+        analysis = RichAnalysis(analysis)
+    cree_wordform_tag_list = analysis.prefix_tags + analysis.suffix_tags
```
eddieantonio, line 99: 🤩

src/morphodict/analysis/__init__.py
```diff
@@ -1,21 +1,100 @@
...
+
+
+def rich_analyze_strict(text):
+    return list(
+        RichAnalysis(r) for r in strict_analyzer().lookup_lemma_with_affixes(text)
+    )
+
+
+class RichAnalysis:
+    """The one true FST analysis class.
```
eddieantonio, line 37: 🤣

eddieantonio (commented):
This looks okay. I have only the following reservations about merging things:

 - migrations should probably be squashed considering the add fields only to delete them immidately
 - some documentation must be updates (field help text, some docstrings)

Other than that, we can hopefully merge it soon and try it. There's still some Cree-specific things sprinkled in the `morphodict` package like the assumption of `+` appearing in prefix/suffix tags, the morpheme rankings file, and a few other oddities, but these can be handled at a later time.

src/CreeDictionary/CreeDictionary/paradigm/manager.py
```diff
@@ -103,6 +114,37 @@ def _load_layouts_from(self, path: Path):
...
+                prefix, suffix = self._LITERAL_LEMMA.split(template)
+
+                prefix_tags = prefix.split("+")
+                assert prefix_tags[-1] == "", f"Prefix ${prefix!r} did not end with +"
+                suffix_tags = suffix.split("+")
+                assert suffix_tags[0] == "", f"Suffix ${suffix!r} did not end with +"
+                ret[template] = (
+                    tuple(t + "+" for t in prefix_tags[:-1]),
+                    tuple("+" + t for t in suffix_tags[1:]),
+                )
```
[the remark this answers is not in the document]
andrewdotn: It will work fine for Tsuut’ina. For Arapaho, this code will need to be adjusted since that FST uses `[TAG]` instead of `+Tag`.

src/morphodict/lexicon/management/commands/importjsondict.py
andrewdotn: Yup, exactly.

src/morphodict/lexicon/models.py
andrewdotn: Oops, thank you for catching this; I had meant to update these but then forgot. You can tell something is amiss from the “blah blah blah” comment nearby.

src/CreeDictionary/DatabaseManager/cree_inflection_generator.py
andrewdotn: This is some dead code that I am leaving for cleanup in a future PR.

andrewdotn: This line came about because merging the main branch into this one pulled in a call to `convert_crkeng_word_class_to_paradigm_name()` which was already gone from this branch.

eddieantonio (approved):

src/CreeDictionary/CreeDictionary/paradigm/manager.py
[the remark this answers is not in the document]
aarppe: Is this reference to the + signs because (somewhere) within itwêwina tags were stilled represented as a single squashed string, rather than as distinct multichar symbols when that is the case?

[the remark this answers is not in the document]
andrewdotn: There are two things in the code that jump to mind right now and still use smushed strings:

  - The paradigm files themselves have smushed strings like `${lemma}+V+AI+Ind+12Pl`. The code here parses that into distinct multichar symbols. The generator FST doesn’t need them separated, but the python code does to show the right tags. A more general but slower method to get this info would be to (1) pass smushed string to strict generator (2) pass generated wordform to strict analyzer, getting multichar symbols back (3) when multiple possibilities are returned by step 2, choose the one whose smushing matches the original from step #1.

  - `crk.altlabel.tsv` has tags with the `+` stripped, e.g., `1Sg` and `2Pl`, as well as smushed things like `0Sg+Fut+Def`. The code that deals with relabelling operates on these tags with the `+` signs removed, and there is code in this PR that has to strip off the `+` signs for compatibility with that.

    Switching the altlabel file to use full tags, maybe with a comma separator like `+0Sg,+Fut,+Def` to make it unambiguous which tag tag the `+` belongs to , is something that would be useful down the road.

src/morphodict/lexicon/test_util.py
dwhieb: [Lushootseed](https://en.wikipedia.org/wiki/Lushootseed) and/or [Nuuchahnulth](https://github.com/dwhieb/Nuuchahnulth#sounds-of-nuuchahnulth) would be good test languages, because they use COMBINING COMMA ABOVE (U+313), which inevitably causes problems.

eddieantonio: 💯  I'VE HAD MANY PROBLEMS WITH U+0313. FONT SUPPORT IS VERY BAD :(

andrewdotn: I was very surprised that [the CBC used it successfully](https://www.cbc.ca/news/canada/british-columbia/heiltsuk-nation-indigenous-name-bc-government-identification-1.6093186) in a news story.

dwhieb: How awful would it have been if they reported on that story and didn't manage to display her name right... 😬

Issue #875, opened 2021-06-24
ESPT Does not recognize "I" as an actor


nienna73:
When doing an espt search, phrases such as "I see you" or "I have" are not recognized as phrases and therefore not inflected. Known examples include (please add more in the comments):

"I have"
https://itwewina.altlab.app/search?q=I+have+espt%3A1

"I see you"
https://itwewina.altlab.app/search?q=I+see+you+espt%3A1

"I speak Cree"
https://itwewina.altlab.app/search?q=I%20speak%20Cree%20espt%3A1

andrewdotn:
I believe the English FST is looking for an uppercase “I” but the search machinery lowercases everything at the start so that case is not significant in searches.

aarppe:
Here are more dysfunctional cases:
* we will speak to you and us
https://itwewina.altlab.app/search?q=we%20will%20speak%20to%20you%20and%20us%20Cree%20espt%3A1
* let you and us sleep now
https://itwewina.altlab.app/search?q=let%20you%20and%20us%20sleep%20now%20espt%3A1
* let us eat later
https://itwewina.altlab.app/search?q=let%20us%20eat%20later%20espt%3A1

aarppe:
I've changed the English phrase analyzer so that a lower-case `i` surrounded by space is regularized as `I`. But that revised FST should be incorporated in itwêwina next.

aarppe:
Current status:
- [x] `I` as actor

<img width="370" alt="image" src="https://user-images.githubusercontent.com/15003407/127415900-5f8af80e-724d-4887-8176-c38404073ddf.png">

- [ ] inclusive `you and we` as actor
- [ ] inclusive `you and us` as goal
- [ ] imperatives

PR #876, opened 2021-06-24
Fix indent that broke non-phrase espt searches

andrewdotn:


nienna73 (approved):
oh my

eddieantonio:
I don't believe this directly fixes #875, since it _still_ does not recognize "I" as an actor; however, it fixes the crash!

PR #884, opened 2021-06-29
Source tooltips for itwêwina

### LINKED ISSUE #393: We need tooltips for source acronyms!
Somehow, we forgot to put tooltips when hovering/tapping on `cw` and `md` source acronyms. I think this worked in the old itwêwina but is missing in the current version. 😱

eddieantonio:
# What's in this PR?

 - **added**: tooltips for source abbreviations! #393
 - **added**: hardcoded dictionary citation information in the only place it's ever used
 - **changed**: `prepareTooltips()` works on the entire page; not just the given `NodeList`
 - **refactor**: `unique_id` template filter moved to `creedictionary_extras`, making `inflection_extras` empty, and ready to be deleted
 - **removed**: `AE` (Alberta Elders' Cree Dictionary) dictionary source, as it was artificially added to the database for one wordform, and otherwise does not widely exist.

Fixes #393.


src/CreeDictionary/CreeDictionary/templates/CreeDictionary/components/citations.html
```diff
@@ -0,0 +1,46 @@
...
+      <cite class="cite-dict" data-has-tooltip aria-describedby="tooltip:{{ id }}">{{ source }}</cite>
+
+      <div id="tooltip:{{ id }}" class="tooltip" role="tooltip" data-cy="citation-tooltip">
+        {% if source == "CW" %}
+          Wolvengrey, Arok, editor.
+          <span class="source-title">Cree: Words</span>.
+          Regina, University of Regina Press, 2001.
+        {% elif source == "MD" %}
+          <span class="source-title">Maskwacîs Dictionary</span>.
+          Maskwacîs, Maskwachees Cultural College, 1998.
```
eddieantonio, line 36: These bibliography entries are in MLA format. Annoyingly, they both have a different format. I think the original intent with the `DictionarySource` model and all of its fields was that there would be some sort of bibliography entry formatter that would intelligently format any given source in the citation format required. But this is weird complexity for exactly two dictionary sources. I'm not even sure the other language pairs have this problem, so for now, all of the formatting and required information is simply hard-coded here.

andrewdotn (approved):

PR #893, opened 2021-07-06
fix: Docker build

eddieantonio:
The Docker build does `npm install --only=production` to avoid installing Cypress... this means **the Docker build depends on what's in the `dependencies` and `devDependencies` keys in `package.json`**.

Is there a better way to solve this?


andrewdotn (approved):
At one point I think I had something like `grep -v cypress package.json | sponge package.json` in the CI build :/

PR #898, opened 2021-07-07
srseng_test_db.importjson and paradigms

andrewdotn:
This PR updates the demo version of the Tsuut’ina dictionary to work with the latest database structure, and includes a small test dictionary.


eddieantonio (commented):
THIS IS A HALF-REVIEW. I'll get to the second half tomorrow morning, but didn't want a bunch of pending review comments just... hanging out here.

munge/srseng/dictionary.ts
```diff
@@ -0,0 +1,245 @@
...
+  assignSlugs() {
+    const used = new Set<string>();
+    for (const e of this._entries) {
+      if (!(e instanceof DictionaryEntry)) {
+        continue;
+      }
+
+      if (e.slug) {
+        assert(used.has(e.slug));
+        used.add(e.slug);
```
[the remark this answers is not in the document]
andrewdotn: Yeah I agree that totally doesn’t make sense.

munge/srseng/util.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+  return ret;
+}
+
+/**
+ * Load a TSV file containing a header row, and return an array consisting of
+ * one {$header: value, …} object for each row.
+ */
+export async function loadTsvFile(path: string) {
+  const contents = (await readFile(path)).toString();
```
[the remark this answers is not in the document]
andrewdotn: I … did not know readFile could take an encoding arg! That is super-useful, thanks!

I generally avoid all `*Sync` calls in case the code gets used in, or copied-and-pasted into, async/server code that would then block.

munge/srseng/util.ts
```diff
@@ -0,0 +1,172 @@
...
+ */
+function toBaseCharacters(s: string) {
+  // remove combining characters
+  s = s.normalize("NFD").replace(combiningRegExp, "$1");
+  s = [...s]
+    .map((c) => {
+      switch (c) {
+        // Characters without combining decompositions. You could partly
+        // automate building a table of these by looking for unicode characters
+        // named “LATIN (SMALL|CAPITAL) LETTER X WITH ___”
```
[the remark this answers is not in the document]
andrewdotn: How do I report errata in that errata document?

> this is encoded as [a] single character

munge/srseng/dictionary.ts
```diff
@@ -0,0 +1,245 @@
...
+    if (existing) {
+      return existing;
+    }
+
+    const entry = new DictionaryEntry();
+
+    if (/^\p{Combining_Mark}/u.test(text.normalize())) {
+      console.log(
+        `Warning: ${JSON.stringify(text)} begins with a combining character`
+      );
```
eddieantonio, line 187: Does this happen often...?

munge/srseng/toimportjson.ts
```diff
@@ -0,0 +1,99 @@
...
+      let [_prefixTags, _lemma, suffixTags] = analyses[0];
+      entry.analysis = analyses[0];
+
+      if (suffixTags.includes("+V") && suffixTags.includes("+I")) {
+        entry.paradigm = "VI";
+      } else if (suffixTags.includes("+V") && suffixTags.includes("+T")) {
+        entry.paradigm = "VT";
+      } else if (suffixTags.includes("+V") && suffixTags.includes("+D")) {
+        entry.paradigm = "VD";
+      }
```
eddieantonio, line 90: Else it doesn't have a paradigm?

eddieantonio (approved):
Just a few minor confusing parts, but other than that, looks okay!

munge/srseng/util.ts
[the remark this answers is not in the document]
dwhieb: I never thought to wrap an `(await func())` call like this so I could chain off it... \*files this tactic away\*

munge/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "prettier": "^2.3.2",
+    "sucrase": "^3.19.0"
+  },
+  "devDependencies": {
+    "@types/node": "^15.14.0",
+    "@types/prettier": "^2.3.1",
+    "chai": "^4.3.4",
+    "mocha": "^9.0.1",
+    "typescript": "^4.3.5"
+  }
```
dwhieb, line 18: What's the basis for distinguishing prod vs. dev dependencies for this package? I would have assumed they'd all be dev dependencies, since none of them get used in production code.

munge/srseng/dictionary.ts
```diff
@@ -0,0 +1,245 @@
...
+    }
+    for (const k of this.senses) {
+      if (k.definition === definition) {
+        return;
+      }
+    }
+    this.senses.push({ definition, sources: ["OS"] });
+  }
+
+  fstLemma() {
```
dwhieb, line 34: Just a thought:

```suggestion
  get fstLemma() {
```

munge/srseng/dictionary.ts
```diff
@@ -0,0 +1,245 @@
...
+};
+
+export class Dictionary {
+  /**
+   * FST tags which distinguish the lexeme, e.g., +N and +V, as opposed to tags
+   * that distinguish the wordform within a lexeme, e.g., +Sg and +Pl.
+   */
+  readonly lexicalTags: Set<string>;
+  _entries: (DictionaryEntry | Wordform)[];
+  _byText: Map<string, DictionaryEntry>;
```
dwhieb, line 58: [Private class fields](https://github.com/tc39/proposal-class-fields) are stable in Node [as of v14](https://node.green/#ES2022-features-instance-class-fields), if that tickles your fancy.

```suggestion
  #entries = (DictionaryEntry | Wordform)[];
  #byText = Map<string, DictionaryEntry>;
```

munge/srseng/dictionary.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  constructor(lexicalTags: string[]) {
+    this.lexicalTags = new Set(lexicalTags);
+    this._entries = [];
+    this._byText = new Map();
```
dwhieb, line 63: These would be:

```suggestion
    this.#entries = [];
    this.#byText = new Map();
```

munge/srseng/dictionary.ts
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  // This is a quick-and-dirty version; the git history has a slug_disambiguator
+  // function with a fairly general-purpose algorithm that could be ported to
+  // js.
```
dwhieb, line 68: 👍🏼

munge/srseng/dictionary.ts
```diff
@@ -0,0 +1,245 @@
...
+      );
+    }
+
+    entry.head = text;
+    this._entries.push(entry);
+    this._byText.set(text, entry);
+    return entry;
+  }
+
+  toJson() {
```
dwhieb, line 196: Are you intentionally avoiding `toJSON()` here? If you call this method `toJSON()` you can then call `JSON.stringify()` on any Dictionary instance and it'll behave as expected:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description

munge/srseng/dictionary.ts
```diff
@@ -0,0 +1,245 @@
...
+    // replace non-lemmas with wordform referring to lemmas
+    for (const [key, entries] of byFstLemmaAndLexicalTags.entries()) {
+      const { fstLemma } = JSON.parse(key);
+      const lemmaEntry = minBy(entries, (e) =>
+        stringDistance(fstLemma, e.head!)
+      );
+      assert(lemmaEntry);
+
+      // FIXME: don’t overwrite the lemma headword; requires dictionary code to
+      // first support having headword ≠ FST lemma
```
dwhieb, line 139: 💯

munge/srseng/toimportjson.ts
```diff
@@ -0,0 +1,99 @@
+import { execIfMain } from "execifmain";
+import { writeFile } from "fs/promises";
+import { Command } from "commander";
+import { join as joinPath, resolve as resolvePath } from "path";
+import { loadTsvFile } from "./util";
+import { Dictionary } from "./dictionary";
+import { Transducer } from "hfstol";
```
dwhieb, line 7: 😍

munge/srseng/toimportjson.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const RESOURCE_DIR = resolvePath(
+  __dirname,
```
dwhieb, line 10: Since this is an ES module wouldn't this be undefined?

munge/srseng/toimportjson.ts
```diff
@@ -0,0 +1,99 @@
...
+);
+
+// TODO: the FST should recognize these forms directly, even if only in relaxed
+// mode.
+function headwordToFstLemma(headword: string) {
+  return headword
+    .replace(/[ā]/g, "a")
+    .replace(/[ī]/g, "i")
+    .replace(/[ū]/g, "u")
+    .replace(/[ō]/g, "o");
```
dwhieb, line 32: Just curious: Can you explain why you bracket these characters in the regex even though they're only 1 character? Is that just a preferred convention?

dwhieb (approved):
Looks great! This'll be a great starting point for me when I start processing Tsuut'ina too.

munge/package.json
andrewdotn: For this sort of command-line stuff, the distinction I try to use is dev-stuff/build-time stuff, vs. stuff needed to run the end product. So `typescript` and `mocha` and all the `@types/` stuff would be dev dependencies, while `commander` is a normal dependency since it’s used at run time.

This can be useful when the transpiled code is deployed to a server to run as a backend. Then you save some time and disk space on the install, and may have a smaller security footprint as well.

I realize I haven’t actually followed that well in this particular `package.json`; I will clean that up.

munge/srseng/dictionary.ts
andrewdotn: Thanks for the suggestion. I’m not sold on private class fields; I can’t use them as a general replacement for `_` variables because private class fields can’t be accessed from subclasses, and sometimes I create subclasses that need to access internal variables of the superclass(es).

TypeScript has `private` and `protected` keywords that I could use if I wanted more protection here. Those keywords aren’t enforced at runtime, and there are workarounds, but TypeScript will ddefinitely complain if you accidentally violate them. Even those keywords normally don’t seem worth the effort for what I’m usually doing.

munge/srseng/dictionary.ts
andrewdotn: Only twice, but it’s hard to spot because the combining dot goes on the `"` at the start of the JSON string, and my terminal doesn’t show that well at all.

munge/srseng/toimportjson.ts
andrewdotn: No, that’s just some leftover thing; I either originally started with more source characters in the square brackets, and didn’t remove the brackets when only a single character was left.

munge/srseng/dictionary.ts
andrewdotn: No, not intentionally avoiding; this method does a lot more than just returning a JSON representation, so I will rename it.

munge/srseng/toimportjson.ts
andrewdotn: I did get this working as an ES module, but the level of complexity and brittleness involved in doing so made me decide it wasn’t worth it, and I went back to having TypeScript emit CommonJS instead.

src/crkeng/site/settings.py
```diff
@@ -35,6 +35,8 @@
 STRICT_ANALYZER_FST_FILENAME = "crk-strict-analyzer-for-dictionary.hfstol"
 RELAXED_ANALYZER_FST_FILENAME = "crk-relaxed-analyzer-for-dictionary.hfstol"
 
+MORPHODICT_SUPPORTS_AUTO_DEFINITIONS = True
```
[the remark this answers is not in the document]
andrewdotn: My current thinking is that all the documentation comments for morphodict settings should go into `morphodict.site.settings`, instead of being duplicated in different language-specific files.  I’ve documented the option and set the default there.

andrewdotn:
I’ve made a fair number of changes to this PR, largely in response by getting a much faster FST. Would anyone like to do additional review before I merge it?

eddieantonio (approved):
Looking good!

PR #901, opened 2021-07-07
fix: nêhiyawêwin relabellings current orthography

eddieantonio:
# What's in this PR?

If you choose to see labels in Cree, the labels are now in the selected orthography as well!

localhost_8000_word_nip%C3%A2w_(iPhone 5_SE) (1)
localhost_8000_word_nip%C3%A2w_(iPhone 5_SE)

---

There are a few refactors for the morphodict Orthography subsystem. I think it could use some more work (like an integration with the Preference subsystem), but these changes will do for now!


nienna73 (approved):
Nice

aarppe:
Note that this takes us some distance towards having nêhiyawêwin as the interface language, in that I've considered the various (re)labelings as part of the interface rather than as the central linguistic nêhiyawêwin content.

Issue #701, opened 2021-03-17
English search word resulting in Cree words being analyzed


aarppe:
When searching for the English word _fox_, one of the results, _mahkêsis_, gets analyzed in the process as a diminutive, which it is, but it's not obvious that is actually the desired behavior, cf.

<img width="452" alt="image" src="https://user-images.githubusercontent.com/15003407/111401552-1d014580-868f-11eb-9abf-2fd0bad43aa1.png">

The extra step of analyzing the Cree results that match the English search term is not necessarily what one would expect. In contrast, when searching with the Cree word _mahkêsis_, that is when one would except an analysis of the form as a diminutive, cf.

<img width="462" alt="image" src="https://user-images.githubusercontent.com/15003407/111401875-c3e5e180-868f-11eb-9434-e69d585c75b6.png">

eddieantonio:
There is no re-analysis. Rather, these are matching _two different entries_. Try https://itwewina.altlab.app/search?q=fox+verbose%3A1

The first result has this match data:

```json
{
  "is_lemma": true,
  "wordform_length": 8,
  "target_language_keyword_match": [
    "fox"
  ],
  "source_language_keyword_match": [],
  "morpheme_ranking": 17.7045,
  "cosine_vector_distance": 0.0,
  "relevance_score": 0.08904526406835
}
```

The second result has this:

```json
{
  "is_lemma": false,
  "wordform_length": 8,
  "target_language_keyword_match": [
    "fox"
  ],
  "source_language_keyword_match": [],
  "morpheme_ranking": 17.7045,
  "cosine_vector_distance": 0.0,
  "relevance_score": 0.08904526406835
}
```

It's matching **two different entries**:

```sql
$ sqliite3 db.sqlite3
SELECT id, text, is_lemma, raw_analysis FROM lexicon_wordform;
```

Yields:

| id | text | is_lemma | raw_analysis |
|-:|-|-|-|
| 312392 | mahkêsîs | 0 | `[[], "mahkêsiw", ["+N", "+A", "+Der/Dim", "+N", "+A", "+Sg"]]` |
| 1083977 | mahkêsîs | 1 | `[[], "mahkêsîs", ["+N", "+A", "+Sg"]]` |

Both the readings exist as entries in the database, I believe **because the FST generates both analyses**. Should we be creating entries for +Der/Dim elements that already have entries in the database?

PR #903, opened 2021-07-08
fix: do not run CVD if query looks like Cree

### LINKED ISSUE #897: Typo in search yields nonsensical results
A search with a mis-spelled preverb shows some strange search results:

screenshot 2021-07-06 at 20 47 34

eddieantonio:
# What's in this PR?

The Google vectors have surprising things that may match strangely with Cree queries. So just don't do a CVD search!

Fixes #897
See also: https://github.com/giellalt/lang-crk/issues/31

### DIFF
--- a/src/CreeDictionary/API/search/core.py
+++ b/src/CreeDictionary/API/search/core.py
@@ -69,10 +69,27 @@ class SearchRun:
         return [r.serialize() for r in results]
 
     def add_verbose_message(self, message=None, **messages):
-        assert message is not None or messages
+        """
+        Add any arbitrary JSON-serializable data to be displayed to the user at the
+        top of the search page, if a search is run with verbose:1.
+
+        Protip! Use keyword arguments as syntactic sugar for adding a dictionary, e.g.,
+
+            search_run.add_verbose_message(foo="bar")
+
+        Will appear as:
+
+        [
+            {"foo": "bar"}
+        ]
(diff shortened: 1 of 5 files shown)


src/CreeDictionary/API/search/core.py
```diff
@@ -69,10 +69,27 @@ def serialized_presentation_results(self):
...
+        Protip! Use keyword arguments as syntactic sugar for adding a dictionary, e.g.,
+
+            search_run.add_verbose_message(foo="bar")
+
+        Will appear as:
+
+        [
+            {"foo": "bar"}
+        ]
+        """
```
nienna73, line 85: Thanks for documenting this!

nienna73 (approved):

aarppe:
I'm wondering whether this is at least partially connected with #701?

eddieantonio:
> I'm wondering whether this is at least partially connected with #701?

Not at all. This problem solves the issue of Cree being erroneously interpretted as English, and then an English cosine vector distance search gets made and matches a spurious, non-sensical vector (i.e., ê and pim are vectors in the Google News corpus).

#701 is not clear to me is actually an issue of Cree words being reanalyzed. It seems like "fox" is matching the auto-definition of "mâhkêsis" which is "a little fox". This is an accurate match! The results are presented strangely though.

PR #908, opened 2021-07-13
Drop old API_* database tables

andrewdotn:
The equivalent data is now in new lexicon_* tables instead.


eddieantonio (approved):
HECK YES!

PR #909, opened 2021-07-13
Allow label edits without restarting server

andrewdotn:
This PR:

  - changes the `LABELS` global to a `@cache`d function, so that label
    edits can be viewed during development without restarting the server

  - removes several cached bits stored in the importjson that can be
    computed from labels

  - restores a little bit of previously-deleted code required for that to
    work

### DIFF
--- a/src/CreeDictionary/API/search/presentation.py
+++ b/src/CreeDictionary/API/search/presentation.py
@@ -171,14 +171,13 @@ def get_emoji_for_cree_wordclass(word_class: Optional[str]) -> Optional[str]:
 
     def to_fst_output_style(value):
         if value[0] == "N":
-            return "+" + "+".join(list(value.upper()))
+            return list(value.upper())
         elif value[0] == "V":
-            return "+" + "+".join(["V", value[1:].upper()])
+            return ["V", value[1:].upper()]
         else:
-            return "+" + value.title()
+            return [value.title()]
 
-    fst_tag_str = to_fst_output_style(word_class).strip("+")
-    tags = [FSTTag(t) for t in fst_tag_str.split("+")]
+    tags = to_fst_output_style(word_class)
     return read_labels().emoji.get_longest(tags)
 
 
--- a/src/CreeDictionary/API/search/presentation.py
+++ b/src/CreeDictionary/API/search/presentation.py
@@ -114,15 +114,17 @@ def serialize_wordform(wordform) -> SerializedWordform:
         if inflectional_category := wordform.linguist_info.get(
             "inflectional_category", None
         ):
-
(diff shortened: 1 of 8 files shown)


src/CreeDictionary/API/search/presentation.py
```diff
@@ -141,7 +158,28 @@ def safe_partition_analysis(analysis: ConcatAnalysis):
...
+
+    def to_fst_output_style(value):
+        if value[0] == "N":
+            return "+" + "+".join(list(value.upper()))
+        elif value[0] == "V":
+            return "+" + "+".join(["V", value[1:].upper()])
+        else:
+            return "+" + value.title()
+
+    fst_tag_str = to_fst_output_style(word_class).strip("+")
```
eddieantonio, line 180: Why prepend `"+"` to all of these and then call `.strip("+")` on the return?

src/CreeDictionary/CreeDictionary/templatetags/relabelling.py
```diff
@@ -19,11 +19,14 @@
...
-    "nehiyawewin": LABELS.cree,
-}
+
+def label_setting_to_relabeller():
+    labels = read_labels()
+    return {
+        "english": labels.english,
+        "linguistic": labels.linguistic_short,
+        "nehiyawewin": labels.cree,
+    }
```
eddieantonio, line 29: Nitpick, but I think it's better to pass in the label setting rather than return a dictionary that will be instantly accessed exactly once, i.e.,

```suggestion
def label_setting_to_relabeller(label_setting: str):
    labels = read_labels()
    return {
        "english": labels.english,
        "linguistic": labels.linguistic_short,
        "nehiyawewin": labels.cree,
    }[label_setting]
```

[...]

src/CreeDictionary/CreeDictionary/templatetags/relabelling.py
```diff
@@ -37,7 +40,7 @@ def relabel(context: Context, tags: Sequence[FSTTag], labels=None):
     else:
         label_setting = labels
 
-    relabeller = label_setting_to_relabeller[label_setting]
+    relabeller = label_setting_to_relabeller()[label_setting]
```
eddieantonio, line 43: If accepting the previous suggestion:

```suggestion
    relabeller = label_setting_to_relabeller(label_setting)
```

eddieantonio (approved):
This is great! Just one confusing part for me and one nitpicky suggestion.

src/CreeDictionary/API/search/presentation.py
andrewdotn: Thanks for pointing this out; I just brought back the code from the git history without looking carefully at what it was doing.

PR #910, opened 2021-07-13
Use speech-db's bulk search API

eddieantonio:
**Requires UAlbertaALTLab/recording-validation-interface#229 to be merged into `production`**

### DIFF
--- a/cypress/integration/recordings.spec.js
+++ b/cypress/integration/recordings.spec.js
@@ -6,11 +6,11 @@ context("Recordings", function () {
     it("should display for words", () => {
       cy.intercept("https://speech-db.altlab.app/api/bulk_search?*", {
         fixture: "recording/bulk_search/wâpamêw.json",
-      }).as("bulk-search");
+      }).as("bulkSearch");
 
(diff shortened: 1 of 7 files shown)


cypress/integration/recordings.spec.js
```diff
@@ -15,23 +20,20 @@ context("Recordings", function () {
...
-      cy.visitSearch("wâpamêw");
-      cy.contains("a", "wâpamêw").click();
-      cy.url().should("contain", "/word/");
-
-      // TODO: we should stub a network request,
-      // but Cypress can't deal with fetch() requests :/
-      // It has to use a polyfill, but for various reasons,
-      // that's annoying and brittle
-      //  (e.g., if the URL changes, the polyfill needs
-      //   to be re-enabled).
```
eddieantonio, line 31: The Cypress version we were using originally did not support stubbing these requests, however, the current version has `cy.intercept()` that _does_ work with `fetch()` requests!

nienna73 (approved):
Bulk API is live: https://speech-db.altlab.app/api/bulk_search?q=min%C3%B4s

PR #916, opened 2021-07-16
Docs index

andrewdotn:


docs/index.rst
```diff
@@ -6,13 +6,45 @@
...
+  descriptions.
+
 .. toctree::
    :maxdepth: 2
    :caption: Contents:
 
-   paradigm-layouts
-   ci
-   developers-guide
+   # TOOD: add conceptual
```
dwhieb, line 42:
```suggestion
   # TODO: add conceptual
```

A comment here doesn't mess up the list of files to include in toctree?

dwhieb (approved):
One tiny question, otherwise looks good! I'll follow up with more revisions in another PR soon.

docs/glossary.md
```diff
@@ -7,6 +7,7 @@ to describing certain languages.
 
 ## Terms
 
+<!-- https://myst-parser.readthedocs.io/en/latest/syntax/syntax.html#syntax-directives -->
```
dwhieb, line 10: 👍🏻

docs/index.rst
andrewdotn: `toctree` seems to ignore it but you’re right it’s not valid syntax, so I’ve removed it

PR #920, opened 2021-07-20
feat: relabel FST labels

### LINKED ISSUE #742: Extend relabeling choice to FST analysis
Expanding from #729, when choosing a relabeling mode, apply that to the presentation of the results of the FST linguistic analysis tags as well.

For instance: _nimaskomak_ -> **maskwa+N+A+Px1Sg+Pl** 
- community/Plain English: -> **naming word + like: asikan, iyiniw, maskwa + my + many**
- linguistic mode: **Noun + Animate + Possessor: 1st person singular + Plural**
- nêhiyawêwin mode: **wîhyowin-itwêwin + tâpiskôc: asikan, iyiniw, maskwa + niya + mihcêt**

eddieantonio:
# What's in this PR?

The "display mode" relabellings now affect the "linguistic breakdown popup" i.e., that thing the relabels the FST label suffix tags.

Linguistic breakdown in linguist mode

Fixes #742


src/CreeDictionary/res/crk.altlabel.tsv
```diff
@@ -207,9 +207,13 @@ Med	Medial	Medial	There	anita
 					
 N		Noun	Naming word	wîhyowin-itwêwin	
 N+A	Animate noun	Noun - animate	Naming word - like: asikan, iyiniw, maskwa	wîhyowin-itwêwin - tâpiskôc: asikan, iyiniw, maskwa 	🧑🏽
+NA	Animate noun	Noun - animate	Naming word - like: asikan, iyiniw, maskwa	wîhyowin-itwêwin - tâpiskôc: asikan, iyiniw, maskwa 	🧑🏽
```
eddieantonio, line 210: Sometimes the tags are "NA". Sometimes the tags are "+N+A" ¯\\\_(ツ)\_/¯

src/CreeDictionary/API/search/types.py
```diff
@@ -4,85 +4,18 @@
...
-from CreeDictionary.utils.types import FSTTag, Label
+
 
 Preverb = Wordform
 Lemma = NewType("Lemma", Wordform)
 MatchedEnglish = NewType("MatchedEnglish", str)
 InternalForm = NewType("InternalForm", str)
 
 
-class LinguisticTag(Protocol):
```
eddieantonio, line 23: This protocol and its implementors were used by `relevant_tags`, which is no longer used. Which is a good thing because these classes are super confusing!

src/CreeDictionary/API/search/presentation.py
```diff
@@ -71,9 +89,19 @@ class PresentationResult:
     presentation things like labels.
     """
 
-    def __init__(self, result: types.Result, *, search_run: core.SearchRun):
+    def __init__(
+        self,
+        result: types.Result,
+        *,
+        search_run: core.SearchRun,
+        display_mode="community",
```
eddieantonio, line 97: Note: I had to punch holes through a lot of systems to get the current display mode from the `HttpRequest` down here where it is used to actually do the relabelling. The relabelling cannot (easily) be done in the template -- I tried 😭 😭 😭

src/CreeDictionary/CreeDictionary/templates/CreeDictionary/components/linguistic-breakdown.html
```diff
@@ -35,8 +35,8 @@ <h3 class="linguistic-breakdown__stem"> {{ result.lemma_wordform.stem }} </h3>
           {% endif %}
 
           <ol class="linguistic-breakdown__list">
-            {% for tag in result.relevant_tags %}
-              <li><data value="{{ tag.value }}">{{ tag.in_plain_english }}</data></li>
+            {% for chunk in result.relabelled_fst_analysis %}
```
eddieantonio, line 38: I tried combinging the display mode here with the relabelled FST analysis (having the serialization contain relabellings for all display modes by default), and it was way more complicated than I thought it was. If you have any ideas how to do that here, rather than pulling `display_mode` down several levels of abstraction, please let me know!

src/CreeDictionary/res/crk.altlabel.tsv
aarppe: The difference is that the crk FST outputs `+N+A`, whereas the lexical database uses `NA-XXX`.

aarppe:
A general note here is that the labels used in the example for linguistic relabeling uses the **long linguistic labels**, whereas elsewhere (in the paradigms) we have been using the **short linguistic labels**. In creating the relabelings I was primarily aiming at making the long linguistic labels as the comprehensive option, though I've tried to fill in the gaps in the short linguistic labels as well (so that the paradigms work). One way to resolve this is that we'd offer both short and long linguistic labels as a general relabeling choice.

src/CreeDictionary/API/search/presentation.py
```diff
@@ -59,7 +61,23 @@ class SerializedPresentationResult(TypedDict):
     preverbs: Iterable[SerializedWordform]
     friendly_linguistic_breakdown_head: Iterable[Label]
     friendly_linguistic_breakdown_tail: Iterable[Label]
-    relevant_tags: Iterable[SerializedLinguisticTag]
+    # Maps a display modes to relabellings
```
nienna73, line 64:
```suggestion
    # Maps a display mode to relabellings
```

nienna73 (approved):
Looks good! I don't know of any ways off the top of my head to do the relabelling without passing display mode down so many levels, we could just make this an issue and call it an "enhancement" so I remember to go back to it later if it makes sense to do so.

PR #921, opened 2021-07-21
fix: more specific language for reduplication

nienna73:
## What's in this PR:
* Use the wording from #585 for the reduplication pop-up


eddieantonio (approved):
LGTM!  🤩 🤩 🤩

PR #924, opened 2021-07-22
change: fetch recordings in batches

### LINKED ISSUE #922: maci- doesn't show recordings on the search result page
For whatever reason, _maci-_ doesn't show recording on the search result page, cf.

image

... even though these are shown on the details page, cf.

image

eddieantonio:
We do requests to the bulk search API in batches as to not upset uwsgi. 4KiB of headers is way too much, so we make smaller requests!

Fixes #922
Resolves: UAlbertaALTLab/recording-validation-interface#239

### DIFF
--- a/frontend/js/recordings.js
+++ b/frontend/js/recordings.js
@@ -109,14 +109,13 @@ function showRecordingsExplainerText() {
  * @return {BulkSearchResponse} see https://github.com/UAlbertaALTLab/recording-validation-interface#bulk-recording-search
  */
 async function fetchRecordingUsingBulkSearch(requestedWordforms) {
-  // TODO: chunk results
-  let chunks = [requestedWordforms];
+  let batches = chunk(requestedWordforms);
 
   let allMatchedRecordings = [];
   let allNotFound = [];
 
-  for (let chunk of chunks) {
-    let response = await _fetchRecordingUsingBulkSearch(requestedWordforms);
+  for (let batch of batches) {
+    let response = await _fetchRecordingUsingBulkSearch(batch);
 
     response["matched_recordings"].forEach((rec) =>
       allMatchedRecordings.push(rec)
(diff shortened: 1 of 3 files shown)


frontend/js/recordings.js
```diff
@@ -146,3 +170,27 @@ function mapWordformsToBestRecordingURL(response) {
...
   return wordform2recordingURL;
 }
+
+/**
+ * Chunks the iterable into subarrays of a maximum size.
+
+ * @param {Iterable<T>}  one or more wordforms to search for.
+ * @return {Array<Array<T>}}
+ */
+function chunk(collection) {
```
nienna73, line 180: This took me a while to mentally trace through, but ultimately it makes sense. This replaces the `lodash` package that we imported, right?

nienna73 (approved):

frontend/js/recordings.js
eddieantonio: Correct! I decided I could do it with a tiny function instead of importing a package. Would adding loop invariants help to understanding the code?

nienna73: Maybe? I found following `currentChunk` was the confusing part. I couldn't understand how it wound up with anything in it if it was being reset to an empty list, but after properly thinking about it, it made sense.

frontend/js/recordings.js
```diff
@@ -146,3 +170,27 @@ function mapWordformsToBestRecordingURL(response) {
...
+  const MAX_BATCH_SIZE = 50;
+
+  let chunks = [[]];
+  for (let item of collection) {
+    let currentChunk = chunks[chunks.length - 1];
```
nienna73, line 185: Why not always use `-1` if you're looking for the last item in the array?
```suggestion
    let currentChunk = chunks[-1];
```

eddieantonio: Unfortunatly JavaScript does not work like that :(

(I really wish it did).

eddieantonio: JavaScript is in the _Cobra, D, Haxe, Java, Scala_ language of familys where the most concise built-in way to get the last item is `array[array.length - 1]`. See: https://en.wikipedia.org/wiki/Comparison_of_programming_languages_(array)#Array_dimensions

nienna73: Ah, in my mind it was already doing that at least once. I thought `[[]]` had length 0, not length 1 for some reason

PR #926, opened 2021-07-23
Add munge script for Arapaho

andrewdotn:
This script uses the FSTs to help assign paradigms!


munge/arpeng/run.js
```diff
@@ -0,0 +1,36 @@
...
+
+const argv = process.argv.slice();
+const nodeExecutable = argv.shift();
+const thisScript = argv.shift();
+
+const targetScript = join(dirname(thisScript), "toimportjson.ts");
+
+let child;
+
+child = spawn(
```
eddieantonio, line 19: I looked into converting this into a _true_ `exec()` — that is, replacing the current process, but then realized that [doing that portably in Node is an unsolved problem](https://stackoverflow.com/q/51185115/6626414). So I guess this okay! 😅

eddieantonio (approved):
LGTM!

munge/arpeng/run.js
andrewdotn: Yeah, windows doesn’t have exec :/ I’ll add a comment.

PR #931, opened 2021-07-27
fix: crash when best_lemma_matches() has zero analyses

### LINKED ISSUE #930: Site crashes when searching for ‘ôhi’

eddieantonio:
Fixes #930

Note: I'm not sure if this is a true fix, as I wonder why the imported items do not have analyses. They absolutely should!


nienna73 (approved):
Let's get this out there for now, but I agree that this should be further investigated to see why this happens.

Issue #935, opened 2021-07-29
Need Cree altlab.tsv relabelling for awîna


andrewdotn:
One of the forms of awîna has analysis awîna+Pron+Interr+A+Sg. As there is no relabelling that includes Interr, that distinction is missing from the linguistic popup.

<img width="320" alt="Screen Shot 2021-07-29 at 4 08 04 PM" src="https://user-images.githubusercontent.com/178162/127572016-2fd16281-df08-4367-b4a7-8686138c5624.png">

aarppe:
Fixed with: https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/commit/11ad7f95452e100890df91a2f07fda64909281dd

PR #936, opened 2021-07-29
Language specific templates

andrewdotn:
This PR:
  - Adds settings for human-readable versions of the dictionary and language name, and parametrizes the default index template to use those names instead of itwêwina / Plains Cree for all languages
  - Shows the unlabelled FST tag in the linguistic popup, with a warning to the server log, if there is no relabelling for the tag. On the integration tests, only #935 was encountered.
  - Adds some experimental linguistInfo for Arapaho, together with a custom template to display it on the frontend

Many templates still have `CreeDictionary` in the name, but that will have to be addressed later.


src/arpeng/app/templates/CreeDictionary/components/definition__elaboration.html
```diff
@@ -0,0 +1,5 @@
+<div class="definition__elaboration" data-cy="elaboration">
+{{ lemma.linguist_info.pos }}
+  • {{ lemma.linguist_info.morphologies|join:" " }}
```
nienna73, line 3: Are these hard-coded mid dots? Why not an unordered list?

src/morphodict/site/checks.py
```diff
@@ -0,0 +1,28 @@
...
+
+@register()
+def check_settings(**kwargs):
+    print("running check_settings")
+
+    errors = []
+
+    # borrowed from what django diffsettings command does
+    if not settings.configured:
+        raise
```
nienna73, line 15: What does it raise?

nienna73 (approved):
Just a couple little questions, otherwise looks good!

src/morphodict/site/checks.py
andrewdotn: Oops, this would be much better off as an assert.

src/arpeng/app/templates/CreeDictionary/components/definition__elaboration.html
andrewdotn: That is a great point.

andrewdotn: I tried that and the CSS was getting very fiddly so I will have to come back to this.

PR #939, opened 2021-07-30
Use correct path to news vectors

andrewdotn:


andrewdotn:
This issue caused a minute or two of downtime when deploying #937; it’s currently manually applied in production.

nienna73 (approved):

PR #943, opened 2021-08-05
Tidy static files

andrewdotn:
- Move everything in `src/CreeDictionary/CreeDictionary/static`
    to `morphodict` or `crkeng`, as appropriate
  - Generate frontend files into `generated/frontend` instead of one of
    the more mysteriously-named `static` directories
  - Font Awesome files are all collected together into one directory,
    and duplicate images have been consolidated


dwhieb (approved):
Looks good! This definitely makes the project structure more understandable. ngl, I was kinda lost about it before...

PR #944, opened 2021-08-05
CMRO

nienna73:
## What's in this PR:
* add CMRO for cwdeng
* only show orthography options when they exist
* use the dialect/language name in the orthography drop down


andrewdotn (approved):
This looks good to me … I notice there is still a TODO for showing which orthography is currently selected, but that could be taken care of later.

PR #958, opened 2021-08-18
Fix readthedocs build

andrewdotn:
- Upgrade docutils to allow specifying parser
  - Fail on warnings
  - Fix existing warnings


andrewdotn:
A different dependency version of a documentation-generation library was causing some of our documentation to appear fine when built locally in a pipenv, but not show up on the readthedocs site. Compare https://morphodict.readthedocs.io/en/latest/linguistic-data.html to https://morphodict--958.org.readthedocs.build/en/958/linguistic-data.html

dwhieb (approved):
Looks good!

PR #975, opened 2021-09-16
iOS app

andrewdotn:
This is the first PR for the iOS app to target the main branch. It should include all changes from the mobile branch.

In addition to the changes allowing the code to be merged into the main branch, this also includes some changes compared to the current UAlbertaALTLab `mobile` branch:
  - There is a new `CURRENT_DB_VERSION` constant to allow triggering a copy of an updated database file from the app into the user’s writable storage when the database schema and/or content changes
  - `do-pip-install` has been ported from shell to python and has several fancy new features, such as extracting python package versions from Pipfile.lock, and removing many megabytes of django translation files that the app would never use
  - Use new hfst-optimized-lookup v0.0.13, and compile C/C++ code from PyPI instead of checking in copy
  - Add empty ‘fake’ modules for packages like foma that are not currently used on mobile due to compilation needs

Let me know if you would prefer I break those up into smaller PRs.

I will email you a copy of the migrated database file if you would like to verify that this builds and runs on your machine.


iOS/README.md
```diff
@@ -0,0 +1,268 @@
...
+    [Neuberg](https://learning.oreilly.com/library/view/ios-14-programming/9781492092087/part02ch04.html#idm45495341795176).
+
+ 5. Wait for the app to finish processing. [`watchbuild`] is a handy tool
+    for this.
+
+ 6. In App Store Connect, find the app, click on the TestFlight tab, click
+    on each group of internal or external testers you want to release the
+    build to, and click the + sign under ‘Builds.’
+
+[`watchbuild`]: https://github.com/fastlane/watchbuild
```
nienna73, line 268: These docs are good and thorough!

iOS/app/AppDelegate.swift
```diff
@@ -0,0 +1,192 @@
...
+
+        let pyq = DispatchQueue(
+            label: "python server", qos: .userInitiated, attributes: [],
+            autoreleaseFrequency: .inherit, target: .none)
+        pyq.async {
+            morphodict_register_callback("hello") { print("Hello from swift") }
+
+            do {
+                try self.setupDatabase()
+            } catch {
```
nienna73, line 67: Does this need to catch the error? As in, `catch error {...`

andrewdotn: Good question! Nope, in swift, while you can adjust the behaviour, it is implied that you are catching something into a var named `error`

https://docs.swift.org/swift-book/LanguageGuide/ErrorHandling.html#ID541
> You write a pattern after `catch` to indicate what errors that clause can
> handle. If a `catch`clause doesn’t have a pattern, the clause matches any
> error and binds the error to a local constant named `error`.

iOS/app/sync-python
```diff
@@ -0,0 +1,68 @@
...
+        --include='*.hfstol' \
+        --exclude='*' \
+        "${SRC_DIR}"/ "${APP_PYTHON_DIR}"
+
+rsync -av "${PROJECT_DIR}"/../fakes/ "${APP_PYTHON_DIR}"/
+
+# byte-compile
+python3.9 -m compileall -f -b "${APP_PYTHON_DIR}"
+# remove original python files
+#find "${APP_PYTHON_DIR}" -regex '.*\.py' -delete
```
nienna73, line 67: Is there a reason this is included but commented out?

src/morphodict/runserver/mobile_run_handler.py
```diff
@@ -0,0 +1,146 @@
...
+                self.condition.wait()
+                print("notified")
+
+    def resume(self):
+        print("RunHandler.resume")
+        with self.condition:
+            self.condition.notify()
+
+    def stop(self):
+        print("RunHandler.stop")
```
nienna73, line 90: Are these print statements routed through the logging function to show up in the device logs?

nienna73 (approved):
Just a couple questions, otherwise this is looking good to me!

src/morphodict/runserver/mobile_run_handler.py
andrewdotn: Yup, python’s `sys.stdout` is rerouted so any `print` statements inside Python go to the system log. A hypothetical raw C `printf` inside some python extension wouldn’t be redirected because it wouldn’t involve the Python output layer at all, but I don’t think that’s a big deal.

iOS/app/sync-python
andrewdotn: Good catch! Thanks.

PR #982, opened 2021-09-23
Support core_definition, semantic_definition

andrewdotn:
This should be safe to merge into production now. It will take a bit longer to deploy than normal while new columns are added to the database. Then, it will be ready for an updated importjson file with the new fields populated where appropriate.


dwhieb (approved):
Everything here looks great (thanks for tackling this so quickly!). Just so I understand: the migration file runs through the current database and adds the new definitions fields?

One suggestion for consideration: use `displayDefinition` instead of just `definition`? What do you think?

dwhieb (approved):

aarppe:
So, according to this implementation, would a definition field in CW below:

`acikâsipakwa    bearberry leaves; kinnikinnick; plant-like evergreen (mixed with tobacco in kinnikinnick) [Lt. Arctostaphylos uva-ursi]`

... be represented as follows:

1. (show) definition (this is the entire definition that is shown to the user):
- `bearberry leaves; kinnikinnick; plant-like evergreen (mixed with tobacco in kinnikinnick) [Lt. Arctostaphylos uva-ursi]`

2. core definiion (this is the definition that is used for English phrase translation):
- `bearberry leaves; kinnikinnick; plant-like evergreen`
- [or even shorter as:] `bearberry leaves`

3. semantic definition (this is the definition that is used as a target in search, and included in the creation of the definition sentence vectors, potentially leaving out some elements like the language references like 'Lt.', metalanguage like 'literally', and perhaps even some of the punctuation such as parentheses):
- `bearberry leaves, kinnikinnick; plant-like evergreen, mixed with tobacco in kinnikinnick; Arctostaphylos uva-ursi`

What I was wondering is what happens with cross-references (in CW with the format: `see: X`)? CW doesn't seem to mix cross-references and additional notes to definitions (as above for _acikâsipakwa_). An example of a definition with a cross-reference would be the following:

`iyîkatê-        aside, off to one side [see yîkatê-]`

Would this be presented as follows?:

1. (show) definition:
* `aside, off to one side [see yîkatê-]`

2. core definition (even if this cannot be inflected):
* `aside, off to one side`

3. semantic definition:
* `aside, off to one side; yîkatê-`

I could imagine that on the longer term we might want to show such cross-references in a special manner so that the cross-referenced entry, e.g. _yîkatê-_ above, is given as a link that the user can follow. But I would assume this would mean that show-definition (1 above) would have special formatting (which isn't currently allowed, if I understand the specs correctly), or that such a cross-reference would have its own special field (that would be shown and formatted appropriately)?

dwhieb:
This is how the Toolbox definition for `acikâsipakwa` is now parsed, using the new definition fields:

```jsonc
"senses": [
  {
    "definition": "bearberry leaves",
    // coreDefinition = definition
    // semanticDefinition = definition
    "sources": ["CW"]
  },
  {
    "definition": "kinnikinnick",
    // coreDefinition = definition
    // semanticDefinition = definition
    "sources": ["CW"]
  },
  {
    "coreDefinition": "evergreen-like plant (mixed with tobacco in kinnikinnick)",
    "definition": "evergreen-like plant (mixed with tobacco in kinnikinnick) [Lt. Arctostaphylos uva-ursi]",
    // semanticDefinition = coreDefinition
    "sources": ["CW"]
  }
]
```

It should display as 3 different definitions in itwêwina (I'll confirm this locally soon).

**acikâsipakwa**
1. bearberry leaves (CW)
2. kinnikinnick (CW)
3. evergreen-like plant (mixed with tobacco in kinnikinnick) [Lt. Arctostaphylos uva-ursi] (CW)

There are 3 senses because we agreed with Arok that senses would be semicolon-delimited.

---

The Toolbox definition for _iyîkatê-_ is parsed as follows:

```jsonc
"senses": [
  {
    "coreDefinition": "aside, off to one side",
    "definition": "aside, off to one side [see yîkatê-]",
    // semanticDefinition = coreDefinition
    "sources": ["CW"]
  }
],
```

In itwêwina, that should display as:

**iyîkatê-**
aside, off to one side [see yîkatê-] (CW)

---

Cross-references are not currently passed onto itwêwina in a structured way (they're just embedded in the display definition), but it would be very easy to add cross-references to the import JSON if morphodict were updated to support them (styling them as links).

aarppe:
For the third definition of _acikâsipakwa_, the `coreDefinition` should not include the parenthesize element, if our intention is that `coreDefinition` specifies what is inflected by the English FSTs. Previously, we had implemented on the coding side an interim measure that excluded any parenthesized elements from being passed to English phrase translation, but the longer-term intention was for this to be dealt with appropriate content structuring on the DB side.

Also, if the semantic definition is supposed to indicate what is available to search (as well as what is included in the creation of the definition vector), then I'd believe that semanticDefinition should include the Latin terms as well, though excluding the abbreviation `Lt.` for Latin. Of course, we might decide otherwise just as well.

So, the third definition for _acikâsipakwa_, should rather look like the following: 

```
  {
    "coreDefinition": "evergreen-like plant",
    "definition": "evergreen-like plant (mixed with tobacco in kinnikinnick) [Lt. Arctostaphylos uva-ursi]",
    // semanticDefinition = "evergreen-like plant (mixed with tobacco in kinnikinnick) [Arctostaphylos uva-ursi]",
    "sources": ["CW"]
  }

```

As for the cross-reference case for _îkatê-_, excluding special formatting of the cross-reference element, the one thing that we would want to avoid is including the metalinguistic _see_ available in the search, so that searching with _see_ rather gives us Cree words with that English word featuring in the actual definition. Thus, we'd want to excluded `see` from the `semanticDefinition`. In this respect, the new specs for _îkate_- should rather be something like the following:

```
"senses": [
  {
    "coreDefinition": "aside, off to one side",
    "definition": "aside, off to one side [see yîkatê-]",
    // semanticDefinition = "aside, off to one side [yîkatê-]",
    "sources": ["CW"]
  }
],
```
What we can fine-tune is whether the Cree cross-reference word yîkatê- should be part of the `semanticDefinition` or not. If we take `semanticDefinition` to include not just what can be used in the semantic vector but also any linguistic content that is made accessible to search, then the Cree words should be included.

Finally, eventually formatting the cross-reference with a link to the appropriate referred-to dictionary entry would be quite desirable. We could consider doing so at this time, rather than punting that until later, though strictly speaking it is not integral to this particular change.

dwhieb:
> For the third definition of _acikâsipakwa_, the `coreDefinition` should not include the parenthesize element [...]

The script that extracts notes from Toolbox definitions only looks for `[square brackets]`, because that's the convention we agreed on with Arok for indicating notes. No special parsing is done for content in `(parentheses)`.

I've opened [this issue](https://github.com/UAlbertaALTLab/crk-db/issues/90) in the crk-db repo to remove content contained in parentheses from the `semanticDefinition` field.

---

> I'd believe that `semanticDefinition` should include the Latin terms as well

@andrewdotn and I made a judgment call not to include it, and planned to consult you to double check. Let us know your final decision, and we'll adjust the `semanticDefinition` field accordingly. (Would the semantic vectors even be useful with Latin terms, since they're based on English?)

---

> What we can fine-tune is whether the Cree cross-reference word _iyîkatê-_ should be part of the `semanticDefinition` or not.

This was the other outstanding question we had for you. Let us know your final decision and we'll update the spec.

---

> formatting the cross-reference with a link to the appropriate referred-to dictionary entry would be quite desirable. We could consider doing so at this time, rather than punting that until later, though strictly speaking it is not integral to this particular change.

I've opened issue #987 for this.

aarppe:
What goes in the `semanticDefinition` is influenced by how that content is used. If `semanticDefinition` specifies only what is used for creating the semantic vector for search purposes, then the Latin and other language words could be excluded. But if `semanticDefinition` specified generally what is available to search, then one could argue that other content in the CW definitions should be included, such as Latin terms (which is what I understood Arok to mean in our last discussion). Nevertheless, my understanding is that metalinguistic content in the definitions, such as words like `see:` or `literally:` should be excluded from what is available to search, and thus excluded from `semanticDefinition`.

dwhieb:
> What goes in the `semanticDefinition` is influenced by how that content is used.
> 
> If `semanticDefinition` specifies only what is used for creating the semantic vector for search purposes, then the Latin and other language words could be excluded, in accordance with your argument.
> 
> But if `semanticDefinition` specifies generally what linguistic content is available to search, then one could argue that other words in the CW definitions should be included, such as Latin terms (which is what I understood Arok to mean in our last discussion) - though I think this can merit further thought about the broader implications.
> 
> Nevertheless, my understanding is that metalinguistic content in the definitions, such as words like `see:` or `literally:` should be excluded from what is available to search, and thus excluded from `semanticDefinition`. This is the case of getting all those other entries than _wâpamêw_ and _wâpahtam_, when searching with `see`, that the exclusion of metalinguistic elements from `semanticDefinition` would solve.

@andrewdotn Could you clarify whether the content of `semanticDefinition` gets used in morphodict for searching content, or to create the semantic vectors, or both?

andrewdotn:
The current state of this PR only uses `semanticDefinition` for making semantic vectors.

However I do see (pun intended) value in having a search field that excludes ‘see’ when it is used in the cf sense, but not when used for defining wâpamêw.

I have to update this PR anyway to reconcile it with the incremental import work, so I am happy to use `semanticDefinition` for both search keywords and making semantic vectors.

I mentioned the possibility of excluding stopwords like ‘the’ earlier. I think we should keep those in `semanticDefinition` for now, until there is time to do a more rigorous quantitative analysis of the impact of that change on search quality.

dwhieb:
+1 for the pun

PR #999, opened 2021-09-29
Use logic to check mobile db status

andrewdotn:
Instead of relying on people remembering to update the `CURRENT_DB_VERSION` flag, which would have needed to become language-specific, use SQL queries and logic to determine if the bundled DB needs to copied into place.


nienna73 (approved):
This is awesome, thank you!

PR #1000, opened 2021-09-30
Updates to crkeng FSTs

nienna73:


nienna73:
@aarppe Do you have a progress update on this issue?

aarppe:
@nienna73 The following scripts (probably the latter is better) should tell how the necessary HFSTs are compiled:

https://github.com/giellalt/lang-crk/blob/main/inc/crk.hfscript
https://github.com/giellalt/lang-crk/blob/main/inc/crk-dict.sh

This doesn't still solve the problem of updating the FSTs to conform with the revised crk DB, nor making use of the regular GiellaLT compilation architecture. But when the LEXC content is updated, the script should work to create the necessary new FSTs.

PR #1001, opened 2021-10-01
Add basic cypress tests for all languages

andrewdotn:
Also switch cwdeng default orthography to SRO with macrons.


andrewdotn:
@dwhieb and I saw a bunch of sporadic test failures with the test in `loading-indicator.spec.js`. It seems to be working now?! I’ve made it stash the cypress videos as an artifact in case it breaks again. If it doesn’t break, you can remove the video stuff to make the CI run faster.

dwhieb (approved):
Cypress tests look good! Everything's working locally for me too.

Issue #1006, opened 2021-10-05
Some itwêwina/morphodict search implementation questions


aarppe:
Here are two questions I can recall now which concern the initial stages of search.

1. What exactly is supposed to happen, and actually happens with very short string searches like `ek`, `eka`, `kg`, `km`, and what might be done about it, if anything? For instance, how are these short strings preprocessed before they are considered for search? (e.g. are some characters stripped off first?)

This is connected with an issue posed by Aaron Fay: https://github.com/UAlbertaALTLab/morphodict/issues/986

1a. I think there's the matter of many short strings having a vector, as an "English" string, that sometimes are meaningful as English, e.g. `km`, that produces a meaningful (initial result, e.g.

https://itwewina.altlab.app/search?q=km

1b. But curiously not all such abbreviations produce as immediately meaningful results, e.g. `kg` and `oz`, in the order of increasing meaninglessness:

https://itwewina.altlab.app/search?q=kg
https://itwewina.altlab.app/search?q=oz

1c. And then there are some strings for which I have no good idea of what's actually happening:

https://itwewina.altlab.app/search?q=ek

1d. Finally, some of these less meaningful short strings receive puzzling ranked results - though this may be due to the search vector that seems like a fragment of a Cree word not really matching well any English definition vectors (or the Cree dictionary entry heads), or some interaction with the other factors influencing the relevance ranking.

2. How English phrases get parsed, i.e. why `you all` gets recognized when searching, but not `you and we` and `you and us`, though the FST tool analyzes the latter two as well. Might there be some preprocessing going on that disallows the proper analysis of this inclusive subject (perhaps concerning the co-ordinating conjunction `and`), or might this be just a `21->12` mismatch (this was Jolene's suggestion).

This is connected with an already existing GH issues: https://github.com/UAlbertaALTLab/morphodict/issues/875

I've asked Jolene to look into the latter second question here as part of an overall review of the ESPT functionality, just to check that nothing hasn't accidentally fallen within some cracks, cf. https://github.com/UAlbertaALTLab/morphodict/issues/1005

Issue #1005, opened 2021-10-04
Review workings/mappings of Cree-to-English and English-to-Cree phrase translation


aarppe:
Currently, when typing in some inflected Cree word forms, or English phrases that should correspond to an English phrase or Cree word, respectively, the process appears to work only partially. For instance:

1. Cree word-forms with certain preverbs, e.g. _kikî-wâpamitin_, _kiwî-wâpamitin_,  _kika-wâpamitin_ do not get an English phrase translation, whereas the non-tense form _kiwâpamitin_ does.

2. English phrases with some inflected forms of verbs, e.g. _I saw you_ or _I helped him_ do not appear to get analyzed resulting in a Cree word form, even though the FST tool does work for them, whereas non-inflected variants such as _I see you_ and _I help him_ get matched with a Cree word form.

3. English phrases with certain multi-part arguments, e.g. _you and we_ or _you and us_, do not appear to be analyzed properly, even though in the FST tool they get the expected analyses, whereas other such as _you all_ do work.

nienna73:
As far as I can tell, the auto-translations only appear for entries that exist in the database. So even if an entry is analyzable, it won't get an auto-translation unless it was in the database at import-time. Auto-translations are generated at time of import. I'm trying to import the latest changes to the db now to see if I can fix this somehow.

aarppe:
Ok. This would correspond to a recollection that I had about how the English translations were originally implemented, i.e. that they were pre-generated rather than generated on the fly.

As noted on several earlier occasions, there was a time when we generated the English phrases for all the word-forms in the paradigms, which included the most common grammatical preverbed forms, i.e. those with kî- and wî- (for both Ind and Cnj), ka- for Independent only (as Future Definite), plus then the ta-/ka- Infinitive forms (with only conjuncts).

The idea with the pregeneration was that it would take some while, but not too long, as it wouldn't be done incessantly.

As we've discussed earlier, generating the preverbed verbal word-forms maybe constitutes thrice the forms that we currently have (i.e. adding 8 subpanes to the current 3), so it should take like 4-5 times as long as currently - the time should be along O(n), not O(n^2).

The new transcriptor generating English verb phrases should be smaller and faster and more scalable.

This is probably also one of those coding implementations that we need to work through comprehensively, so that what remains is ideally primarily updating.

nienna73:
[...]

PR #1007, opened 2021-10-14
Fixed (re-)coding of inclusive first person plural tags from English FST a…

aarppe:
Recode English FST analysis for inclusive first person plural with `21` to crk FST format with `12` (for actor: `+21Pl` -> `+12Pl`, goal: `+21PlO` -> `+12PlO`, and possessor: `+Px21Pl` -> `+Pxl12Pl`).

Based on local version seems to work, but one yet needs to check whether one more change is needed concerning the possessor tags in [crk_tag_map.py](https://github.com/UAlbertaALTLab/morphodict/blob/main/src/CreeDictionary/phrase_translate/crk_tag_map.py).

This addresses issues #1005 (point 3), #1006 (question 2), and this earlier comment https://github.com/UAlbertaALTLab/morphodict/issues/875#issuecomment-888722609.

### DIFF
--- a/src/CreeDictionary/API/search/espt_crk.py
+++ b/src/CreeDictionary/API/search/espt_crk.py
@@ -28,7 +28,7 @@ noun_passthrough_tags = {
         "+Px3Sg",
         "+Px1Pl",
         "+Px2Pl",
-        "+Px12Pl",
+        # "+Px12Pl", # Needs to be recoded: 21 -> 12
         "+Px3Pl",
         "+Px4Sg/Pl",
         "+PxX",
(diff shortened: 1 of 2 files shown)


aarppe:
@nienna73 @andrewdotn This seems to fix this matter, though there is this one detail (noted above that should be checked). I somehow thought that we had already changed this, but it must have been eclipsed by other matters.

I believe Andrew's explanation also clarifies why the tensed verb forms with any of the preverbs do not work (any more), but how to resolve that needs some careful thinking.

I tried to check that this doesn't break any tests, and if this is the case, then I do think this can be deployed.

PR #1018, opened 2021-11-18
Revised FST and DB sources for Gunáhà

aarppe:
First draft at next round of TVPD based content for Gunáhà, fixing glitches in previous version, starting with harmonizing choice of lemma for both LEXC and DB source.


src/srseng/resources/dictionary/srseng_test_db.importjson
```diff
@@ -3852,7 +3852,7 @@
     "paradigm": "VI",
     "senses": [
       {
-        "definition": "he/she is drinking (something)",
+        "definition": "he/she is drinking (something) (something)",
```
dwhieb: Is the duplication here intentional?

aarppe:
Nope, it's accidental, and I've revised the code creating the JSON source so that such duplicates would not happen - the challenge is a non-systematic implementation of the coding of argument classes in the XML source.

aarppe:
The modified files should now be as good as I can create them programatically, even pending some confirmations of my interpretation of the XML source from Chris, so these should be releasable in the morphologically intelligent Gunáhà.

aarppe:
@nienna73 @dwhieb I realize that we'll have to resolve in some fashion that the TVPD based JSON source (nor the word-list based FSTs) do not contain the example words in the test DB file (that were extracted from OS). The primary reason for this is that some of those verbs do not occur in TVPD, but that means that we do not in fact cover all paradigm types (at least the `+O` and `+E` cases).

When testing the DBs and FSTs locally, I've just excluded the old DB content with the `--purge` option - otherwise those currently obsolete entries will turn up in Gunáhà, which makes less sense if they're not really supported by the FST (which would be difficult with the word-list based approach).

aarppe:
@nienna73 This is the Gunáhà PR with the updated DB and FST files.

PR #1021, opened 2021-11-19
change: get language code from url

nienna73:
## What's in this PR:
* update the bulk search request to include the language

**Notes:**
This PR must be merged at or around the same time as: https://github.com/UAlbertaALTLab/recording-validation-interface/pull/294

That PR has the API changes for speech-db that allow these changes to work.

### DIFF
--- a/frontend/js/recordings.js
+++ b/frontend/js/recordings.js
@@ -9,25 +9,12 @@ const BULK_API_URL = `${BASE_URL}/${LANGUAGE_CODE}/api/bulk_search`;
 
 function getLanguageCodeFromLocation() {
     const location = window.location.toString();
-    switch (location) {
-        case location.includes("itwewina"):
-            return "maskwacis";
-            break;
-        case location.includes("itwiwina"):
-            return "woodscree";
-            break;
-        case location.includes("gunaha"):
-            return "tsuutina";
-            break;
-        case location.includes("arpeng"):
-            return "arapaho";
-            break;
-        case location.includes("guusaaw"):
-            return "haida";
-            break;
-        default:
-            return "maskwacis"
-    }
+    if (location.includes(`itwewina`) || location.includes(`crk`)) return `maskwacis`;
+    if (location.includes(`itwiwina`) || location.includes(`cwd`)) return `woodscree`;
+    if (location.includes(`gunaha`) || location.includes(`srs`)) return `tsuutina`;
+    if (location.includes(`arpeng`)) return `arapaho`;
+    if (location.includes(`guusaaw`) || location.includes(`hdn`)) return `haida`;
+    return `maskwacis`;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


frontend/js/recordings.js
```diff
@@ -3,7 +3,32 @@ import SimpleTemplate from "./simple-template.js";
 // the specific URL for a given wordform (refactored from previous commits).
 // TODO: should come from config.
 const BASE_URL = "https://speech-db.altlab.app";
-const BULK_API_URL = `${BASE_URL}/api/bulk_search`;
+const LANGUAGE_CODE = getLanguageCodeFromLocation();
+const BULK_API_URL = `${BASE_URL}/${LANGUAGE_CODE}/api/bulk_search`;
+
+
+function getLanguageCodeFromLocation() {
+    const location = window.location.toString();
```
dwhieb, line 11: Ha neat, I didn't know `location` stringifies that way. I usually go with `window.location.href`.

frontend/js/recordings.js
```diff
@@ -3,7 +3,32 @@ import SimpleTemplate from "./simple-template.js";
...
+            break;
+        case location.includes("arpeng"):
+            return "arapaho";
+            break;
+        case location.includes("guusaaw"):
+            return "haida";
+            break;
+        default:
+            return "maskwacis"
+    }
```
dwhieb, line 30: JS case statements are evaluated for whether they're strictly equal to (`===`) to the switch expression. So the case statements are evaluating whether `location === location.includes('itwewina')`, which will always be false (since `location` is a string and `location.includes()` returns a boolean). So this switch block will always hit the `default` statement and return `maskwacis`.

Also, the return statement will automatically break out of the switch statement, so the `break` statements will never be reached and can be removed.

I suggest this approach instead:

```js
if (location.includes(`itwewina`) return `maskwacis`;
// an else statement isn't necessary since you're doing an early return if there's a match
if (location.includes(`itwiwina`) return `woodscree`;
...
```

dwhieb (changes requested):
Caught an issue with the switch statement which will make it behave unexpectedly and suggested a fix.

aarppe:
Should we not be using the ISO codes here, to the extent possible?

dwhieb:
Seems like Glottocodes would be even better, since ISO codes underdistinguish language varieties sometimes.

frontend/js/recordings.js
nienna73: I had a feeling something was off with that switch statement, thanks for catching that!

nienna73:
The language code returned is actually a url-safe version of a language variant in the speech-db. I was initially using ISO codes, but that doesn't allow me to distinguish Stoney Alexis from Stoney Paul, for example.

aarppe:
What we could use for local variants is the same that overall framework that is used for the scripts, under BCP, namely variants:

https://www.rfc-editor.org/rfc/rfc5646.html#section-2.2.5

So, Plains Cree in Maskwacîs could have the code: `crk-maskwacis`

These subvariants are not standardized as far as I understand, but I only briefly skimmed this document.

nienna73:
I can change the language codes, but it will require a refactor on the speech-db side. Are we okay with these changes for now and I can do the refactor in the coming days?

PR #1037, opened 2021-12-17
Paradigm audio

nienna73:
## What's in this PR:
When an inflected form in a paradigm has a recording, show that recording in the paradigm. Looks something like this:
Screen Shot 2021-12-17 at 4 08 11 PM

## What's needs to be done?
The buttons aren't always very responsive, sometimes they need to be clicked multiple times before the audio plays. This needs to be debugged before it's pushed to production.


aarppe:
The design looks pretty much what I'd expect.

For the sake of explicitness, how do we select the recording for the word-form when there are many? In the case of Gunáhà we may have multiple recordings from Bruce, but in the case of itwêwina we may have multiple recordings from multiple speakers (though the paradigm coverage is much lesser).

On the longer term, one might imagine some setting where a user can choose this, but this will probably require some dedicated thinking as to what type of implementation might be best.

nienna73:
The recordings are filtered to only include good recordings, but then we just take the first one returned. We can specify this later on so always chose a particular speaker or add a "best" filter in speech-db as you suggested.

PR #1048, opened 2022-01-21
change: uncomment line of code to include preverbs in espt

nienna73:
I uncommented a line of code to include preverbs in the translation process again.

@aarppe from what I can tell, this makes the search functionality a lot slower, not the import process like I thought.

The query `you will dance espt:1 auto:1 verbose:1` takes about 13 seconds to return results with this change, whereas it only takes about 1-2 seconds on production right now. The results, as far as I can tell, are the exact same.

Feel free to pull this branch and look at it locally, but I'm not sure if this change is worth it at this time.

### DIFF
--- a/src/CreeDictionary/phrase_translate/translate.py
+++ b/src/CreeDictionary/phrase_translate/translate.py
@@ -38,7 +38,7 @@ if package_dir not in sys.path:
 
 from CreeDictionary.phrase_translate.crk_tag_map import (
     noun_wordform_to_phrase,
-    verb_wordform_to_phrase,
+    verb_wordform_to_phrase, permitted_preverb_tags,
 )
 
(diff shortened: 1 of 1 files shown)


aarppe:
Since the English FST we use to analyze the phrase is the same, regardless of whether we make use of the preverbed cases, and since the English FST we use to generate the translations is the same, regardless of whether we exclude the preverbed cases, and those should be pregenerated, I have a hard time understanding how the implementation would slow up so much. When using the verbose option, the analysis of the phrase and its conversion to Cree features happens quickly, and accessing English phrases if pre-generated should not take that long. So, we'd need to eventually explore in more depth why this takes so long.

Nevertheless, in some way, having either espt or auto work separately, but not at the same time, isn't necessarily the greatest loss - but not being able to get preverbed Cree wordforms as translations is undesirable.

image

PR #1067, opened 2022-03-30
fix: relabelling for info table

nienna73:
Relabelling for the info table seems to be working now. Here are the three different modes:

Linguist:
Screen Shot 2022-03-30 at 10 45 55 AM

English:
Screen Shot 2022-03-30 at 10 46 06 AM

Tsuut'ina:
Screen Shot 2022-03-30 at 10 46 21 AM

@aarppe Does this look correct now?

### DIFF
--- a/src/CreeDictionary/API/search/presentation.py
+++ b/src/CreeDictionary/API/search/presentation.py
@@ -239,17 +239,15 @@ class PresentationResult:
             info = re.findall(pattern, analysis)
             cleaned_info = []
             for i in info:
-                print(i)
                 if "<b>" in i:
                     j = i.replace("<b>", "").replace("</b>", "")
                 else:
                     j = i
                 cleaned_info.append(j)
-            print(cleaned_info)
             relabelled = []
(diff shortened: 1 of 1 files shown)


aarppe:
Yes, it looks as it should. In the revised srseng dictionary DB I currently only have two columns, since the TVPB1 doesn't have the TAMA chunks, nor the LEXC-style lexical tier representation, so I'm just giving an example 3rd person singular verb form per aspect. But my understanding of the table implementation is that it really is up to the linguist how the table is specified and what content it has. One question I would have is how is the header row boldfacing specified? By `<b>...</b>` tags in the table? I'm asking this as the code looked like it would remove such boldfacing fields.

aarppe:
Ok, good - and I've added a revised version of the srs-to-eng DB, that makes use of the tabular format, cf. https://github.com/UAlbertaALTLab/morphodict/blob/main/src/srseng/resources/dictionary/srseng_test_db.importjson - so eventually we could update the DB on the production site.

